PR #12, opened 2017-01-22
Update de.po

mockturtl:
From @kghbln


kghbln:
@mockturtl @NikoKrause Thanks a lot. Great collaboration, indeed!

PR #18, opened 2017-01-24
reactivate  places-with-terminal@mtwebster

ccprog:
places-with-terminal@mtwebster has not been working for some time because it
1. referenced `nautilus` and `gnome-control-center`
2. used a custom Menu that contained a function call `PopupMenu.PopupMenu.prototype._init` that used an invalid signature

This fix brings the applet back to a working state and additionally contains one small improvement: After clicking the terminal icon for a place, the popup menu now closes.


mtwebster:
Thanks! :)

PR #21, opened 2017-01-25
Changed author

jake1164:
Updated author to current maintainer of the spice.

I took over maintaining the feed from Jon Brett last year, confirmation can be provided privately.


NikoKrause:
@jonbrett
Are you okay with that?

jonbrett:
Yes I'm fine with that. Sadly I don't have time to maintain this now, but @jake1164 is doing a sterling job, thanks :-)

PR #27, opened 2017-01-26
[location-detection@heimdall] Replacing ip address from panel to tooltip of flag and update error icon

BrainAxe:
I just changed the look of this applet. Before it showed ip address at the right side of country flag on the panel. Now ip address is showing in the tooltips of the country flag. This way the applet use less space of the panel. I also update the error icon. Make sure to create a new api(from ipinfodb) and use it to test this applet. Thanks


clefebvre:
Thanks, I need @heimdall's approval for this. Hopefully he's active on github.

PR #38, opened 2017-01-27
Fix for scripts@paucapo.com

### LINKED ISSUE #37: scripts@paucapo.com
@Odyseus  @NikoKrause 
Can some fix problem with this applet.
It won't show script list, It only works when I manually replace in applet.js file, in line 32
"/.scripts-applet" with "/.local/share/nemo/scripts/"
Then I try to change in settings Scripts directory, some other directory and back to  "/.local/share/nemo/scripts/" directory and then don't work again.

muzena:
Fixed   #37  thanks to  @Odyseus

### DIFF
--- a/scripts@paucapo.com/files/scripts@paucapo.com/applet.js
+++ b/scripts@paucapo.com/files/scripts@paucapo.com/applet.js
@@ -29,7 +29,7 @@ MyApplet.prototype = {
       this.settings.bindProperty(Settings.BindingDirection.IN, "icon", "icon", this._onSettingsIcon, null)
 
       if (this.directory == "" || typeof this.directory == "undefined") {
-         this.directory = GLib.get_home_dir()+"/.scripts-applet";
+         this.directory = GLib.get_home_dir()+"/.local/share/nemo/scripts";
       }
       this.autoupdate_last = this.autoupdate;
       this.directory_last = this.directory;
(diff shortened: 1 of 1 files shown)


NikoKrause:
@Odyseus 
Did this applet work in Cinnamon < 3.2 and will this change break it for the old Cinnamon versions?
If yes, it would be better to go multiversion.

@paucapo

Odyseus:
@NikoKrause.

Yes, as far as I tested, the applet itself works as it should on Cinnamon 3.2.x. The only thing that doesn't work is the directory selection on its settings window, but this patch by @muzena fixes it.

I don't think that enabling **multiversion** for such a simple applet would be needed. Maybe in the future when and if the developers remove the current retro-compatibility for deprecated options types like **checkbox** (which this applet uses) might be necessary, but not just now.

NikoKrause:
Ah yes. The if statement is not true for Cinnamon < 3.2 right?

Odyseus:
@NikoKrause.

That's right.

**In Cinnamon < 3.2.x:** dir = file:///path/to/dir
**In Cinnamon > 3.2.x:** dir = /path/to/dir

NikoKrause:
Yes, than you are right. Makes no sense for multiversion in this case.

P.S.: A friendly correction: Cinnamon >= 3.2.x   ;)

paucapo:
Hi all, I'm not using Cinnamon anymore, anyone can take care of this simple applet?

Thanks!

PR #39, opened 2017-01-28
Added applet timeout@narutrey

NaruTrey:
A simple applet for Cinnamon to control breaks when working with your computer.


Odyseus:
Hello, @NaruTrey.

Check your .pot file. Its header is malformed.

NaruTrey:
I did not know about the utility `cinnamon-json-makepot`, thanks

NikoKrause:
As Odyseus said.
You don't need `generatePOT.awk` and `generatePOT.sh`.
Open a terminal and go to the directory, where your `metadata.json` file is stored. Then type
`cinnamon-json-makepot --js timeout@narutrey.pot`
The pot file will be stored besides metadata.json and you can move it in the po directory.
Or you can type
`cinnamon-json-makepot --js po/timeout@narutrey.pot`. In this case the pot file in the po directory will be overwritten.

If you have updated your pot file,
you can update your `ru.po` file with poedit.

1. Open your localized (translated) .po file in poedit.
2. Under “Catalog” top menu, select “Update from POT file…” menu item.
3. Select the newly downloaded .pot file and click Open in file dialog box.
4. Your files will merge and you will be shown a dialog with new and obsolete strings.
5. Proceed to translate the newly translatable strings.
6. Save and exit, your updated .po file is ready.

If you want to test your translation, then type in the same terminal (location of terminal must be again besides metadata.json)
`cinnamon-json-makepot -i`
This will install the po files in "~/.local/share/locale/"

Why is this line in your metadata:
"localedir": "~/.local/share/locale"

And the title of this pull request says timeout@libnaru.so. Shouldn't it say timeout@narutrey?

NaruTrey:
> And the title of this pull request says timeout@libnaru.so. Shouldn't it say timeout@narutrey?

You have a keen eye.

I made the necessary changes, now everything should be OK.

NikoKrause:
What's the purpose of `localedir": "~/.local/share/locale` in metadata.json?

Everything else looks good to me. But I'd like, if someone else confirm this as well, since the whole process is still pretty new. I hope you'll understand.

NaruTrey:
It's an old line at the moment. I wrote this applet 10 months ago, and then my localization is not used without the string. Looked at now, and without that line all works. Now I delete this line.

NaruTrey:
By now it should be ok

PR #49, opened 2017-01-31
update func for setting custom css style class

zagortenay333:


NikoKrause:
@collinss

Odyseus:
@zagortenay333.

The **setCustomStyleClass** isn't a global function, so the `if (typeof setCustomStyleClass === "function")` condition will always be **false**. it should be `if (typeof this.menu.setCustomStyleClass === "function")`.

zagortenay333:
@Odyseus oops wasn't paying attention. Fixed

collinss:
@zagortenay333 Thanks! I don't want to merge this just yet because the version here is a little bit behind, and I'm afraid I'll forget to make sure it's in there after I update it. I'll hopefully have it up to date soon, but it will probably be easiest to simply copy over from the version in the original repo, which means that this pr may need a rebase. I'll let you know when the time comes though.

PR #61, opened 2017-02-02
add timer notifications applet

axos88:


NikoKrause:
A couple of notes:
1. The name seems long to me. That's no problem per se, if you want to have it this way. But if somebody adds an issue or pull request, the title will probably contain only your uuid ;) I'd suggest timer-notification@axos88, because it's obvious that this is an applet and that it is for Cinnamon. But it's up to you.
2. The loadOptions function looks strange. Is it on purpose?
3. Last-edited in metadata.json is added automatically as far as I know. I don't know, if this will cause problems, when you add it manually.
4. What's the purpose of "repo" in info.json? Is it used somewhere? Haven't seen it before. You can use "website" in metadata.json instead, to link to your applets website.

And last but not least:
My biggest concern is the uuid you are using in metadata.json. It should match with the directory name of the applet. Else it will fail to load, if I remember correctly.

axos88:
Thanks for the feedback, I will update the PR shortly.

axos88:
I addressed your concerns

PR #68, opened 2017-02-02
Update ITM to 4.3.2, add README.md and ES2015 source files to Icing Menu, Icing Window Saver #26

### LINKED ISSUE #26: Icing Task Manager: Unable to use with LibreOffice
OS: Linux Mint Cinnamon 18 and 18.1
When opening an existing LibreOffice document and minimising it to taskbar or swiching to another task, the  icon dissapears or, if pinned to taskbar, just sits there as if no doc is in use, rendering the user unable to restore window. Only reappears (sort of) after opening a blank document.

jaszhix:
Icing Task Manager v4.3.2

  * Fixed a bug causing new LibreOffice windows to not be shown.
  * German translation update by [SethosII](https://github.com/jaszhix/icingtaskmanager/pull/93).
  * Added compatibility checks for the experimental Classic Gnome project.

Fixes #26


NikoKrause:
I have a general question and a note. (Just out of curiosity)

Why do you have a src directory in Icing Task manager? What is its purpose?

And the not:
You are storing .mo files in the localization. Probably for the localization script, which copies the .mo files to the needed directories?
It's not needed to store these files. If you need a script I would suggest to change it.
The command `cinnamon-json-makepot -i` used in the same directory as metadata.json does create .mo files on the fly out of the .po files.

I think you have to rename the `locale` folder to `po` for this. But I'm not sure about that.

jaszhix:
It is so I can use most ES2015 features, which requires Babel transpilation. This is explained in the readme files.

The localization setup is inherited from the fork. It seems to work, however I just use msgfmt, not the script in locale. I am always open to pull requests.

NikoKrause:
Ah. I didn't heard of msgmft yet.
Till now cinnamon-json-makepot sufficed for me to create/update translation updates and to install and remove translations.

jaszhix:
The direction I've decided to go with ITM is start making changes for dual compatibility with Cinnamon and Gnome Shell, so I'd like to use platform agnostic tools where possible.

NikoKrause:
Just wanted to let you know, in case you didn't.
Never worked with GNOME Shell yet.

jaszhix:
I forgot to compile de.mo, its added to the PR.

mockturtl:
@jaszhix 

Off-topic, but I'd love to hear your thoughts on this: 

https://www.webreflection.co.uk/blog/2015/12/08/writing-native-apps-with-javascript

jaszhix:
I've used jsgtk, its a really neat library. It works for CJS as well after I submitted a PR, and the author implemented it. It does however only fake the Node environment, and any code that needs to be modified to run for CJS, won't work with it either - it adds the module system, and syntactic sugar. 

I also have made a few Electron apps for work and as side projects - its a really convenient way to create standalone apps if you are familiar with browser and node JS. The main thing Node and G/CJS have in common is they're wrappers for native C bindings, and you can do a lot of powerful things natively with them. I think if more distributions installed Node by default like they do Python, you would see more Node integration in the Linux free desktop ecosystems.

jaszhix:
I bumped the version of Icing Menu - forgot the updater addition wasn't tagged on my repo. Let me know if anything else needs changed to get this merged.

PR #73, opened 2017-02-03
[all-in-one-places@jofer] Applet is not loading correctly

belandsoft:
Fix for applet is not loading correctly on recent Cinnamon versions issue


JosephMcc:
Please use the applet name in the commit message. Something like `all-in-one-places@jofer: Fix applet not loading correctly`

PR #85, opened 2017-02-06
axos88@countdown-timer: fix uuid and broken link in "About..."

NikoKrause:
* UUID in applet.js was wrong and it was not possible to add applet to panel
* Link to website didn't launched.

@axos88


axos88:
thx

PR #90, opened 2017-02-06
Popup Translator (0dyseus@PopupTranslator) update

Odyseus:
- Fixed keybinding display on applet tooltip.

### DIFF
--- a/0dyseus@PopupTranslator/files/0dyseus@PopupTranslator/metadata.json
+++ b/0dyseus@PopupTranslator/files/0dyseus@PopupTranslator/metadata.json
@@ -1,17 +1,18 @@
 {
-    "dangerous": false,
-    "uuid": "0dyseus@PopupTranslator",
-    "max-instances": "-1",
-    "description": "Simple translator applet that will allow to display any selected text from any application on a system in a popup.",
-    "multiversion": true,
-    "version": "1.02",
-    "name": "Popup Translator",
-    "comments": "Bug reports and feature requests should be reported on this applet's repository linked below.",
-    "website": "https://github.com/Odyseus/CinnamonTools",
-    "contributors": "NikoKrause: German localization and bug reports.,Radek71: Czech localization and bug reports.",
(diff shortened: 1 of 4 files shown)


0dyseus@PopupTranslator/files/0dyseus@PopupTranslator/metadata.json
```diff
@@ -1,17 +1,18 @@
...
-    "description": "Simple translator applet that will allow to display any selected text from any application on a system in a popup.",
-    "multiversion": true,
-    "version": "1.02",
-    "name": "Popup Translator",
-    "comments": "Bug reports and feature requests should be reported on this applet's repository linked below.",
-    "website": "https://github.com/Odyseus/CinnamonTools",
-    "contributors": "NikoKrause: German localization and bug reports.,Radek71: Czech localization and bug reports.",
+    "website": "https://github.com/Odyseus/CinnamonTools", 
+    "multiversion": true, 
+    "last-edited": 1486145103,
```
Odyseus: Hello, @NikoKrause.

All my applets/extensions have that property set because I keep my repository in sync with the xlets that I have installed in my system. Never was a problem and still isn't because the **last-edited** property seem to be automatically updated when the .zip files are generated.

In the past I used to remove it. But the metadata file keeps been edited and its properties reordered by who knows what. So I do not bother any more.

NikoKrause: Okay. Let's try it. Only hesitated because I already broken the update mechanism once a few days ago^^

The reorganization of the metadata.json file happens to me as well. It always puts the dangerous flag in there. I don't want it for my applet, so I `git checkout metadata.json` before commit.

Odyseus: > I don't want it for my applet, so I git checkout metadata.json before commit.

That's an *overhead* that I really don't want to be subjected to.

PR #94, opened 2017-02-07
search@mohammad-sn: Added new applet

mohammad-sn:
**Search** is an applet that adds indexing and search capability to Cinnamon desktop. It uses a modified version of mlocate/updatedb in order to search for files.


jaszhix:
Why did mlocate need to be modified, and what is modified?

mohammad-sn:
Hi @jaszhix 
In the original mlocate, updatedb uses whitespace character to seperate directories passed to --prunepaths (used for excluding directories from indexing). As it doesn't support regex or wildcards, there is no way to exclude a directory with whitespaces in its path. I modified updatedb to interpret '?' character in prunepath variable as a wildcard which could replace any character. Modified source code is included in the applet.

Odyseus:
Hello, everybody.

@mohammad-sn: I don't think that you should ship your applet with a binary. In fact, no xlet should be allowed to be shipped with a binary.

What I think you should do is to make your applet work with the binaries provided by the system and provide instructions on how to get and install the modified binaries that you suggest could be more useful to use with your applet (at the USER's own risk).

Providing the source of the binaries isn't much of an improvement either.

- You shipped the source code compressed, forcing the reviewers to download and decompress the files.
- Even if the source would not have been compressed, you are forcing the reviewers to review hundreds of files and thousands of lines of code.
- And even if any reviewer would be willing to do all that job, they have to build from source to compare the resulting binary with the one that you provide with your applet.

I simply don't think that a reviewer should be subjected to such a labor. Just my two cents.

jaszhix:
I agree with Odyseus, it seems like a lot could be abused if binaries are allowed in.

With that said, @mohammad-sn, I think you could do all of this in javascript using Gio and GLib.

mohammad-sn:
@Odyseus, @jaszhix, I'll consider making it work with system's mlocate package. I agree with you about shipping binaries with an xlet.
Thank you.

collinss:
If it turns out that you absolutely can't do something without compiling, we might be able to implement a mechanism where you can include the source files and have them compiled on installation. That would be a lot of work, though, so I would only suggest doing it as a last resort.

clefebvre:
Closing for now because of the presence of the binary files.

PR #99, opened 2017-02-08
clipboard-qr@wrouesnel: fix applet from crashing

NikoKrause:
When closing QR scanner without having scanned anything, the applet does not pop up anymore.
Catch this error.

Restore Readme from wayback machine


NikoKrause:
@wrouesnel

PR #101, opened 2017-02-09
Icing Task Manager: Update to 4.3.3

jaszhix:
* Russian translation update by
[ArionWT](https://github.com/jaszhix/icingtaskmanager/pull/96).
* Updated the Lodash dependency to 4.17.4, and modified it to prevent
CJS warnings.


NikoKrause:
Since we talked about binaries in another pull request.
`.mo` files are not really binary files, but you can't really check/view them and they are created from `.po` files, so no need for them to be there actually.
Can you remove them please?
We'll also remove them form other applets as well.

jaszhix:
No, those binaries are not executable. I'm sorry, that doesn't make sense. They're needed for localization to work.

NikoKrause:
they are created automatically from .po files,

jaszhix:
This applet [manually handles them](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/IcingTaskManager%40json/src/applet.js#L471). We've been over this before already, this was written before the language tools were available, and I don't want to use them at the moment.

jaszhix:
I think this should be merged, because I think opinionated handling of localization would represent a shift in policy, and should require communication to developers ahead of time. ```mo``` files are binaries in the same sense that images are binaries. The other applet you reference was using ```GLib.spawn_sync``` on its included binaries. So, not comparable at all.

NikoKrause:
I'm not speaking for Linux Mint, but for myself.
I as a reviewer can't see what's in the .mo file. I can look at an image file though.

Your manual way of handling this is way too complicated. All you have to do is to store the .po files in a directory called po. Done. Nothing else.

If you install your applet through System Settings, Cinnamon checks for the po directory and creates the .mo files in the needed path.

I don't know what's the opinion of others.
Better ask @collinss

jaszhix:
Its a small update, and I think if this is a requirement we should update it when you remove the ```mo``` files from all the other repo's, but until then I think it should be merged.

jaszhix:
By the way, if you cd into ```src/locale```, run ```msgfmt -cv -o ./mo/ru.mo ./ru.po````, the mo binaries will be identical. The de.mo file was supposed to be included in the last release, and I accidentally only committed it into the src directory. I agree your suggestion sounds more ideal, but perhaps for next time.

collinss:
I think it is a good policy to have nothing in this repository that cannot be reviewed on Github - not just for security reasons, but for maintainability. Applet developers are a particularly transient bunch (as a whole, not individually). We never know who will decide to abandon their applets, and we need to be able to maintain it if that happens. We can't afford to have exceptions, either.

Now I realize that .mo files are a little different than an executable binary, but since we have a mechanism to generate the .mo files at install time, I would prefer to use that moving forward.

That being said, I agree that we can't just implement new rules without warning and expect everyone to bend over backwards to comply immediately. I don't think it's unreasonable to give some time before this policy on .mo files needs to be enforced. In the mean time, I will add this to the readme, and we'll be cleaning up the other applets that have binary data. I will also merge this PR, but please look into making the change at your earliest convenience. If you need some help, or guidance, don't hesitate to ask either.

PR #108, opened 2017-02-10
network@brownsr: add l10n/translation support and German translation

NikoKrause:
also renamed applet from "network" to "Network"

@brownsr: Are you ok with the renaming?


brownsr:
No problem at all !

PR #113, opened 2017-02-12
localipswithsettings@edaubert: new applet

edaubert:
This applet is a fork of ```localip@mrieracrespi``` which add a way to configure what is displayed.
I also change the way the IPs are retrieved because the way it is done on ```localip@mrieracrespi``` doesn't allow to filter the IPs.

There  is still one issue which is not handled about the ```destroy``` of the applet.
Indeed as on ```localip@mrieracrespi```, even if the applet is not displayed anymore, the function which retrieved the IPs are still executed and seems to never stop.
However I didn't find a way to know that the applet has been destroyed so it seems that nothing can be done.


NikoKrause:
A few things.

You should add an icon.png besides info.json.

The description in metadata is quite long. It should be a short description of what the applet does. The Rest can be added to README.md, e.g. that It's a fork of an applet, etc.

Regarding your issue. I think there's a function 
`on_applet_removed_from_panel`
Maybe that's something you searched for.
It's used for example here https://github.com/linuxmint/cinnamon-spices-applets/blob/master/CinnVIIStarkMenu%40NikoKrause/files/CinnVIIStarkMenu%40NikoKrause/3.2/applet.js

edaubert:
Hi,

Thanks for you remarks.
I have tried to take them into account.

I have added the missing icon, and create a README.md.

Thanks for ```on_applet_removed_from_panel``` I had missed it before.
I update the code to take it into account.

Regards.

Odyseus:
Hello, everybody.

@edaubert: In the future, a better way to "clean up" the functions is as follows:

First, declare a property on applet **_init** that will be used to "store" the function:

```javascript
this.myFunctionId = 0;
```

Then "store" the function:

```javascript
if (this.myFunctionId > 0) {
    Mainloop.source_remove(this.myFunctionId);
    this.myFunctionId = 0;
}

this.myFunctionId = Mainloop.timeout_add_seconds(this.settings.getValue("refreshInterval"), Lang.bind(this, function refreshTimeout() {
    this.refreshLocation();
}));
```

And then add the code to do the "clean up" on applet removal:

```javascript
on_applet_removed_from_panel: function () {
    if (this.myFunctionId > 0) {
        Mainloop.source_remove(this.myFunctionId);
        this.myFunctionId = 0;
    }
}
```

By the way, you can simply use **this.refreshInterval** instead of **this.settings.getValue("refreshInterval")**.

edaubert:
I understand the use of `source_remove` but I am not sure this is enough.

What about concurrency ?

Are you sure that it is not possible to have concurrency on the execution of `refreshLocation` and `on_applet_removed_from_panel` ?

Because otherwise, it may happen that the `source_remove` is applied on the running function and so a new one will be added.

collinss:
@edaubert concurrency is not an issue within Cinnamon. For better or worse, Cinnamon is single-threaded, and so there is only one piece of code running at any given time. The Mainloop.timeout_add_seconds will wait the specified time, and then try to run as soon after that as possible, but it will wait until the current line of code has finished executing first.

edaubert:
Ok. Thanks for the explanations.

I made the changes proposed by @Odyseus.

collinss:
@edaubert Sorry, I've got a few more things.
* README.md is in uuid/files/uuid/. In order for this to show in your applet page on the spices website, it should be in uuid/ along with your info.json, icon.png, and screenshot.png
* Line 101 in applet.js uses a synchronous call. This is usually not a problem as most such commands finish very quickly, but it does freeze Cinnamon while it is running (that's how synchronous calls work). If the command takes longer than normal for some reason, the freezes will become noticeable to the user and could even hang the ui indefinitely if the script fails to terminate. You should change this to an async call instead. If you need an example of how to do this, I did the same thing with one of my applets [here](https://github.com/linuxmint/cinnamon-spices-applets/commit/9d81d620befd5f757aae30f0a8af35caecef6f5d).
* In metadata.json, you have an extra argument ```"dangerous": true``` which probably shouldn't be there. It is autogenerated by Cinnamon when the applet is added, and is used to identify to the user any applets that have calls that are potentially "dangerous" (have been known to cause freezes/crashes/etc). In your case it is set to true because of the synchronous call I mentioned in my previous point. It would probably be best to just remove that argument from your pr entirely.

edaubert:
Hi,

Thanks for the new remarks, I have taken them into account.

collinss:
Thanks for being patient with our requests!

PR #132, opened 2017-02-18
Add Cinnamenu@json

jaszhix:
Cinnamenu is a full featured menu applet for Cinnamon. It is a fork and port of the GNOME Shell extension [Gnomenu](https://github.com/The-Panacea-Projects/gnomenu) by The Panacea Projects.

screenshot

### Features

 * View apps in a list or grid.
 * Dedicated column for places or favorites.
 * Integrated bookmarks from Chrome, Chromium, Firefox, Opera, and Midori.
 * Search through apps, files, bookmarks, and places simultaneously.
 * Indicators for running apps.

Note - this applet doesn't need to be transpiled with Babel like my other applets, so its easier for others to submit PRs for.

[Repo](https://github.com/jaszhix/Cinnamenu)


clefebvre:
Files should be in files/$UUID, not files/.

A lot of things don't work with many of the themes. Some labels aren't even visible.

I'm also worried because we have "menu: consider adopting this layout (https://raw.githubusercontent.com/The-Panacea-Projects/Gnomenu/master/Screenshot.png), same as ours/mintmenu, only better." in our roadmap, so although the work done here helps a lot... we might end up integrating this or something very similar into Cinnamon itself.

jaszhix:
Yeah, I saw that, it actually gave me the idea to port it over. I wasn't sure if that was tentative or not, so I did it anyway (for fun). I was assuming the work might benefit Cinnamon, and maybe the project would borrow a few things. Whether it does or not, I find the features helpful and think its worth publishing. If Cinnamon switches to this or something similar, I don't see it being any different than the many other forks of the current default menu applet.

Which labels aren't showing? I replaced most of the classes with the ones the default menu applet uses. I tested this on the Linux Mint theme, Minty-Y/Darker, and a handful of themes from spices.

clefebvre:
logout, restart and apps/places in listview.

Odyseus:
Hello, evrybody.

@jaszhix: Don't leave files/folders inside the **files** folder other than the actual xlet folder. It's known to break xlet installation. You should also change the folder structure of your other xlets.

jaszhix:
This is actually kind of confusing. When this repo was created that was the structure my applets were already in, and I haven't had anyone tell me their installation broke in this way. I see they just recently changed the documentation here. I will update those applets' folder structures when I need to update those applets again.

jaszhix:
The logout/restart buttons are icon-only, but their labels appear in the selected app box. In the apps/places - you mean the container for the list of apps? These are all St.ScrollView containers, and they can all be scrolled to reveal more of the list. If you, or someone else, could elaborate with a screenshot, maybe one that doesn't contain any sensitive files (this can read recent files if the privacy option is enabled in Cinnamon).

I have a busier weekend ahead, so I might not get to this as soon as I thought, but we'll see. I'll mark it [WIP] and Mint devs can feel free to amend this if it helps it get merged.

jaszhix:
> The presence of this UUID directory, inside of files/ isn't actually needed by Cinnamon (as Cinnamon creates it if it's missing), but it is needed to guarantee a proper manual installation (i.e. when users download the ZIP from the Cinnamon Spices website).

@NikoKrause This is the line that confused me...

> The UUID/files/ directory has to be "empty", which means that it should contain ONLY the UUID directory. Else the spice won't be installable through System Settings.

I think it should be explained that the structure is actually UUID/files/UUID, but it seems like cinnamon-settings installations don't suffer from this and its only for manual installations - which seems like its normally for developers anyway. I haven't had any trouble explaining this to people on the ITM repo as far as manual installation.

NikoKrause:
If you have all your files in the files folder there's no problem. But if your metadata.json and applet.js files are in files/UUID and you have some files in the files folder except UUID, then you get the install problem.

brownsr:
Good going, a very interesting looking applet.  My immediate thought on opening it was 'what are these additional entries in the column on the left'.  I soon realised that they are similar to the favourites bar in the standard menu, with additional mode switching buttons above.  That made me realise just how well the standard menu hid complexity by using a few icons in contrast to the list format of the main menu categories.  You might like to think about how to reduce visual complexity for the user. What is in a favourites bar in the standard menu appears to be broken up into three, with mode switches above, a list of favourites, and a second list at the bottom for power and lock functions.  It would be better to my eyes if these three elements were combined visually. It wasn't hard to find the odd issue, but for a work in progress that's to be expected.  All in all this looks promising.  If you can make the complex twiddles in the user interface disappear and make it more intuitive then that would be even better.

jaszhix:
Yeah, this is mostly a representation of Gnomenu in its current state, with a few features reliant on GNOME Shell APIs either removed (workspace thumbnails based on Main.overview), or commented out (frequently used app list). 

To me as a developer and power user, this currently suits my needs. I could see the UX for most people being improved in the ways you suggested, which is definitely interesting to think about. I agree the current default applet is very good, and I've used it for a long time. I think condensing some of these elements and making the interface's features more obvious is a good first step in that direction.

As far as Cinnamon integration, if it makes you feel better, I ported it with a less opinionated work flow and use of libraries, in case it is used as a base for other projects. Its not exactly adhering to Mint's style guide, but its basically there if you format it to four space indentation.

jaszhix:
Changes:

- Change default value for shortcuts-label to false. This setting toggles whether or not the far left column shows labels alongside the places or favorites icons. This will default to false for now so there is less text clutter.
- Re-comment places/favorites button, as it was only uncommented from the original version for experimentation.
- Remove toggleCategoryWorkspaceMode  …Reimplementing this feature in Cinnamon would require including several versions of files from GNOME Shell's imports.ui, which would involve hacking global objects - out of this project's scope.
- Remove hide-panel-menu setting as it is reliant on a removed feature over an unsupported API
- Fix list/grid toggle button not re-rendering the app list view
- Revert menu-favorites-box class on PanelMenuButton.bottomPane
- Revert changing of default for shortcuts-display

jaszhix:
[This line](https://github.com/linuxmint/cinnamon-spices-applets/pull/132/commits/034f9eb7de1244179814dd69553be39aa5b88565#diff-3d760011bd6bd4759c1811b03baf0b5fL4219) was removed because it caused an extra gap on the right of the menu button on the panel. 

The menu in PanelMenuButton (subclassed from PanelMenu.Button) is overridden with the menu from its parent callee class, CinnamenuButton (subclassed from Applet.TextIconApplet like the default applet), so PanelMenuButton's actors are being added to the stage that way.

brownsr:
Issues with the latest version:

- The places/bookmarks to the left are coming up with very small icons.  The only place I can see what each corresponds to is at the bottom right of the whole menu.  That's a long way to look across for a description and makes mouse control a bit awkward. 
- Not sure what the criteria to appear in the places/bookmarks list are.  
- There are missing icons for toggle startup apps view, restart, logout user and preferences (bottom right)
- Some of the favourite apps are coming up with '-' on the line underneath them
- I can't see what the 'Bookmarks' button does.  All I can see it doing is blanking out the menu when I press it.
- There is a lot of space to the right in list view. Perhaps the control toggles and search bar can be moved left to allow this to be closed down ?
- The search functionality appears to be searching the applet names fine, but is not bringing up things using the description.  The word 'document' gets very different results in this menu compared to the default menu, for example
- 'startup apps view' is strangely titled. It appears to toggle between all applets and the favourites. I'd be tempted to remove it, as the way that the favourites appear as the first category makes them easy to locate, so I can't see that a way to reduce the applet list to just the favourites adds anything.

The main menu functionality seems to work really well. Performance is nice and snappy.  I'm not convinced about the design choice to put a variable length list of places/bookmarks to the left, the way the default menu adds it as a category seems more intuitive as it means that the location and the full description are visually nearby.  

Progressing well :)

jaszhix:
> The places/bookmarks to the left are coming up with very small icons. The only place I can see what each corresponds to is at the bottom right of the whole menu. That's a long way to look across for a description and makes mouse control a bit awkward.

Yeah, that's why I gave them labels. You're right, we may want to kill that column and move it to the category list.

> Some of the favourite apps are coming up with '-' on the line underneath them

This is actually a feature, but its not obvious what it means. It indicates the app is running. I couldn't find a class for a small indicator dot so its just hard coded at color #FFF for now.

> There is a lot of space to the right in list view. Perhaps the control toggles and search bar can be moved left to allow this to be closed down ?

I like that idea.

> The search functionality appears to be searching the applet names fine, but is not bringing up things using the description. The word 'document' gets very different results in this menu compared to the default menu, for example

I carried over Icing Menu's search method over (uses [this](https://github.com/miguelmota/knuth-morris-pratt/)), because I found sometimes I would get the wrong results when searching the default way. Like searching for "settings" brings up everything but settings. What I think is needed though is prioritizing the name in the order of the search results, and give exact matches a higher score, and concatenate the remaining results matched by description behind it.

> 'startup apps view' is strangely titled. It appears to toggle between all applets and the favourites. I'd be tempted to remove it, as the way that the favourites appear as the first category makes them easy to locate, so I can't see that a way to reduce the applet list to just the favourites adds anything.

I agree, it seems a bit unnecessary and there's a bit of a lag if it opens with "All applications" because it has to render the biggest list on init, which isn't likely what the user will be wanting to see, looking through a huge list (not me anyway). 

> There are missing icons for toggle startup apps view, restart, logout user and preferences (bottom right)

This is not what I'm seeing. So to make sure we're on the same page, I created a quick diagram.

menu_diagram

Thanks for the feedback.

brownsr:
I suspect the missing icons are simply because you have additional icons on your computer that are not in the standard set, perhaps you have installed various additional  gnome packages in the course of the development.  The restart button may not be necessary, as pressing shutdown calls up the dialog that chooses between options.  Logout in your app uses 'user-logout-symbolic', logout in the standard app uses 'system-log-out'.  I imagine I would find the same if I investigated the other missing icons.

jaszhix:
menu_vm

I installed this in a development VM I use for testing ITM using Cinnamon Git, and I still see all the icons, but I had to install a ton of things to compile Cinnamon there, so I'm installing a fresh LM18.1 on a new VM.

jaszhix:
Changes:

Fix search regression
Remove shortcuts-label, shortcuts-display, hide-shortcuts, hide-useroptions
Remove category-selection-method
Remove category-hover-delay setting
Remove hide-categories setting
Fix grid column classes
Comment St.Scrollview constraint
Fix grid items getting cut off near the end
Fix schema for apps-grid-column-count
Refactor _displayApplications method
Merge AppListButton and AppGridButton
Use all app metadata in search, but prioritize the name in the order
Clean up supporting files
Trigger menu width change on view mode toggle
Restructure menu layout
Move/repurpose shortcuts column right of the app list, remove the extension preferences button, fix list toggling in bookmarks, places, and recent views.
Subclass CategoryListButton, GroupButton with PopupMenu.PopupBaseMenuItem
Remove height on cinnamenu-categories-workspaces-scrollbox
Remove startup apps display
Move recent and bookmarks buttons to category list

brownsr:
That's getting to be a pretty slick looking menu.  Context menu still needs addressing, and I can't see what the bookmarks button does [and it stays locked on when pressed].  But the main menu stuff is fast, clean and uncluttered.  Nice.  I've limited time during the working week to test things, but I'll do my best if you put further changes through.

jaszhix:
Thanks for the help. I cleaned up some of the CSS so its more cross-theme compatible on my repo. I'll push more changes here when I have more, so there's not a ton of commits from me here. Or I could close this and reopen it with a squashed commit. I'm still thinking about how I want to handle the context menu - I copied over the base context menu class from the default applet, but seems like it will have to be integrated differently because I'm using different kinds of actors for the menu items.

brownsr:
Seeing this image makes me realise a little more about why I am not seeing the bookmarks.  I had been assuming bookmarks as in the nemo bookmarks, you clearly mean browser bookmarks from what I can see.  I see nothing.  That makes me assume there must be some other indexing package you have installed in order for this to work. The link you posted to gnomenu seems to indicate that gir1.2-gda-5.0 is what is needed, so I will try that.

jaszhix:
@brownsr Ah yeah, that's a good point. I forgot about the Gda requirement - ITM uses it as well for Firefox bookmarks. I still need to test Chromium and Opera, and maybe I'll track down the problem.

jaszhix:
Changes:

Add autoscroll, grid icon size scaling, theme compatibility fixes
Create unique bookmark arrays
Add a spacer to pad the toggle button from the search box
Fix chromium bookmark indexing, recurse bookmark structures
Add category icon size setting, theme introspection
Fix context menu behavior
Add overlay context menus for grid view
Add context menu functionality
Add context menu functionality
Move CinnamenuButton, CinnamenuPanel, buttons classes into separate files
Rename PanelMenuButton to CinnamenuPanel
Fix list view app labels not appearing on LM theme, fix category button hover CSS, narrow container width

I tested it on 20+ themes and I'm not seeing any issues. I think its good to go for a 1.0 release, unless someone finds a crippling bug.

NikoKrause:
A few notes:
* The icons in the grid view are HUGE.
* The items in All Applications and in Office are not alphabetically ordered (I'm using German l10n)
* If I open an application through the menu, e.g. firefox, the menu grabs the focus of the cursor. When I left click anywhere on the panel, firefox opens. Right click shows the following errors: http://pastebin.com/MnhEQPwK
* I think it's a little bit slow. I don't have the best hardware, but compared to other menus it's really slow.

brownsr:
There's clearly a condition where the icons can double their dimension.  Not all categories, but quite a few, perhaps those with fewer entries ?  I could trigger this from the search:

(cinnamon:5251): Cjs-WARNING **: JS ERROR: TypeError: itemActor._delegate._app is undefined
CinnamenuPanel.prototype.selectActiveContainerItem@/home/simon/.local/share/cinnamon/applets/Cinnamenu@json/panel.js:1359
CinnamenuPanel.prototype._doSearch@/home/simon/.local/share/cinnamon/applets/Cinnamenu@json/panel.js:1507

I have no speed problems, if NikoKrause is getting a load of error log messages that might be related as a lot of errors being logged will definitely not help speed .  One time through I was seeing some of the same .xsession-errors as NikoKrause

(cinnamon:5073): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x55e39184a830 StScrollView.menu-context-menu:last-child] which is not in the stage.

But other times through I never saw this.

NikoKrause:
The icons in categories with many items are smaller than the icons in categories with just a few items. I have a category with 2 applications only. The icons are huge.

When I hover over the AllApps category it lags for a second before even showing up.
Sometimes there are blank spots before the icons show up.

jaszhix:
Thanks for the feedback. I'll investigate these errors, and see if I can reproduce them. Yeah the grid icon size scales by how many apps are in a list. This was intended to address the "lots of empty space" issue in the grid view. I think its okay for icons to look huge if they're using the space available. You can turn this off in the applet settings.

jaszhix:
Changes:

* App launching bug fixed
* Apps ordered alphabetically
* Improved category navigation lag
* Improved icon rendering
* Fixed favorites being updated closing the menu, optimized how they're refreshed

PR #133, opened 2017-02-18
Remove IcingMenu@json

jaszhix:
I plan on maintaining [this](https://github.com/linuxmint/cinnamon-spices-applets/pull/132) applet instead.


NikoKrause:
Should this be removed now or when the Cinnamenu is uploaded?

jaszhix:
@NikoKrause I guess we might as well wait.

collinss:
Yeah, I don't see the point in removing it until we have something to replace it with.

jaszhix:
I'll let you guys decide if you want this removed. There's a bug where the categories repeat, and may have to do with the height calculation being changed. I'm just trying to budget my time, but I'm not against it staying on the repo if someone wants to try to fix it.

collinss:
It don't think there is any reason to keep it if you're not going to maintain it. Except for the extra layout settings there doesn't seem to be a lot of difference between this and the stock menu applet.

PR #154, opened 2017-02-23
feeds@jonbrettdev.wordpress.com: fix some variable warnings

collinss:
@jake1164 Thought I'd run this by you before merging in case it interferes with your workflow.

FYI, I also submitted [this pr](https://github.com/linuxmint/Cinnamon/pull/6303) to Cinnamon. When it's merged, it will allow you to move your feeds list to the settings api if you want.

### DIFF
--- a/feeds@jonbrettdev.wordpress.com/files/feeds@jonbrettdev.wordpress.com/applet.js
+++ b/feeds@jonbrettdev.wordpress.com/files/feeds@jonbrettdev.wordpress.com/applet.js
@@ -72,7 +72,7 @@ FeedApplet.prototype = {
         this.feed_queue = [];
 
         try {
-            debug_logging = this.settings.getValue("enable-verbose-logging");
+            let debug_logging = this.settings.getValue("enable-verbose-logging");
 
             // Initialize a debug logger
             this.logger = new Logger.Logger({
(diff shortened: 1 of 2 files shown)


jake1164:
The merge is fine, I can just pull those changes before I create a pull with my current changes.


The updated settings would have been nice to know, I have been redoing the feed management stuff.


Question about the new settings api @collinss, and this is why I am currently limiting to 1 instance of the applet.


Currently, if you have multiple applet instances, when a user adds a new instance of the applet it creates a new "instanceid.json" settings file.


However, when you remove that instance the settings file goes away (including all the feeds). SO if you are on a laptop, and have the applet on a panel on a secondary screen and take the applet off, the applet no longer is visible (another bug, it really should move, temporarily, to a visible panel), and you remove that applet and then add it to the visible panel, all settings (feeds) are lost.

Will this be the same with the new settings api?

Reproduce using feeds as example:

  1.  If you currently have any feeds back them up.
  2.  Remove any instances of feeds
  3.  Change max-instances to 3 (any number above 1) in feeds metadata.json
  4.  Add feed to a panel
  5.  Note that NNN.json is created in .cinnamon/configs/uuid
  6.  Remove that instance
  7.  Note that NNN.json is now gone, along with any feeds in it.

I know I don't want to loose all my feeds settings every time I move an applet around.

________________________________

collinss:
I'm not entirely sure of the answer to your question without some testing, but I can tell you that I didn't touch that stuff when I reworked the settings. What you describe actually sounds like a bug that we saw before with settings being inappropriately removed for multi-instance applets. It was also causing some issues with my sticky notes applet. IIRC it was fixed within the last release or two.

PR #155, opened 2017-02-24
Slovenian translation for the Weather applet

ratek1:
I added Slovenian translation for the Weather applet. I hope everything is OK.

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/po/LINGUAS
+++ b/weather@mockturtl/files/weather@mockturtl/po/LINGUAS
@@ -27,6 +27,7 @@ ro
 ru
 tr
 sk
+sl
 sv
 uk
 vi
--- a/weather@mockturtl/files/weather@mockturtl/po/sl.po
+++ b/weather@mockturtl/files/weather@mockturtl/po/sl.po
@@ -1,0 +1,438 @@
+# Slovenian translation for "Weather" cinnamon applet
+# Copyright (C) 2011
+# This file is distributed under the same license as the weather@mockturtl package.
+# Ratek1 <Ratek1@github>, 2017.
+# 
+#, fuzzy
+msgid ""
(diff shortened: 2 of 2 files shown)


collinss:
@mockturtl You seem to get a lot of these translations prs. Do you want us to wait for your review on them or would you prefer we simply merge them? I'm fine either way.

mockturtl:
@ratek1 Hvala!  :+1:

mockturtl:
@collinss It's fine with me if you just merge them.

Is the LINGUAS file still necessary?

collinss:
@mockturtl OK, sounds good.

I don't think it's necessary. I don't remember seeing it in other applets with translations. You'd have to ask @NikoKrause to be sure though. He's the resident expert on applet translations.

NikoKrause:
I started just recently. What was the LINGUAS file for?

It suffices to put the po files in a directory called po. The rest will be done automatically, when you install the applet from System Settings.
Here's a small tutorial:

https://github.com/NikoKrause/CinnVIIStarkMenu/issues/19

Odyseus:
Hello, everybody.

I don't think that the LINGUAS file was ever needed. The Cinnamon code doesn't even contain any reference to a so named file (I searched only on linuxmint/Cinnamon repo and not on other packages).

For what I can grasp (barely) from [the documentation](https://www.gnu.org/software/gettext/manual/html_node/po_002fLINGUAS.html), the LINGUAS file is used to tell the system all available languages that will be installed. And then there is a [LINGUAS environment](https://www.gnu.org/software/gettext/manual/html_node/Installers.html#Installers) variable used to limit the languages that will be installed.

Anyway, Cinnamon doesn't care if there is a LINGUAS file or not (at least regarding xlets localizations). It will install all .po files found inside a po folder without exceptions.

PR #156, opened 2017-02-25
hamster@projecthamster.wordpress.com: make this applet work on my LMDE Cinnamn 3.2.7 and fix the icon display on the applets'list

edaubert:
I have done some small fixes to on `hamster@projecthamster.wordpress.com` as it failed to start on my system (Linux Mint LMDE, Cinnaomn 3.2.7)

- remove `const _` as I didn't see its usage and it was failing the start of the applet
- remove the comment at the top of prefs.js file as it was failing the settings access
- remove one line inside the metadata.json to use icon.png (which has also been copied from the parent folder)


NikoKrause:
The _ is used for translations. In this case the applet has a const _ and a function _ . Probably that's causing some problems.

Comments shouldn't be a problem. Maybe it's because the Shebang #! Is not in the first line.
Instead of deleting the comment you should move it under the line with the Shebang #!

Please squash your commits to one and use the spices name in the commit message title like you did in the pull request title.

@collinss

edaubert:
I have done the change about te Shebang.

What about const _ and the function ?
Is it better to remove the function and keep the const ?

NikoKrause:
No. It's better to remove the const _ like you did.

Does it work for you now with the Shebang in first line?

Since collinss made those changes, he'll look over it and merge. But it looks good to me now.

edaubert:
Yep it works fine with the Shebang in the first line.

collinss:
Looks good. I don't think you really need the shebang anyway. I've never seen it used in js files before in Cinnamon.

PR #165, opened 2017-03-03
weather@mockturtl applet: [[Feature]] Added in the Functionality to Summon the Applet by Key Command

WammKD:
@mockturtl: The ability to change the key binding was added to the settings-schema.json file and the necessary library was imported into applet.js to grab the keybinding from the settings and assign it to the function on_applet_clicked().

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -339,11 +339,12 @@ MyApplet.prototype = {
 
   // Override Methods: Applet
 , _onKeySettingsUpdated: function _onKeySettingsUpdated() {
-    if (this.keybinding != null)
+    if (this.keybinding != null) {
       Main.keybindingManager.addHotKey(UUID,
                                        this.keybinding,
                                        Lang.bind(this,
                                                  this.on_applet_clicked))
+    }
   }
 
 , on_applet_clicked: function on_applet_clicked(event) {
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -344,8 +344,6 @@ MyApplet.prototype = {
                                        this.keybinding,
                                        Lang.bind(this,
                                                  this.on_applet_clicked))
-
-    this.on_applet_clicked
   }
 
 , on_applet_clicked: function on_applet_clicked(event) {
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -11,6 +11,7 @@
 const Cairo = imports.cairo
 const Lang = imports.lang
 // http://developer.gnome.org/glib/unstable/glib-The-Main-Event-Loop.html
+const Main = imports.ui.main
 const Mainloop = imports.mainloop
 
 /**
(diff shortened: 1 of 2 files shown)


weather@mockturtl/files/weather@mockturtl/applet.js
```diff
@@ -328,6 +338,16 @@ MyApplet.prototype = {
...
 
   // Override Methods: Applet
+, _onKeySettingsUpdated: function _onKeySettingsUpdated() {
+    if (this.keybinding != null)
+      Main.keybindingManager.addHotKey(UUID,
+                                       this.keybinding,
+                                       Lang.bind(this,
+                                                 this.on_applet_clicked))
+
+    this.on_applet_clicked
```
jaszhix: Looks like a typo, this function is left here without being called.

collinss:
Please change the commit message to include the name and/or uuid of the applet. This will help keep our history clean and easy to read. Also, you should include the same information in the title (we have about 200 applets) and reference the author's name in the main comment.

@mockturtl

WammKD:
I'm sorry about that; that's a fair point. The mentioned changes have been done.

weather@mockturtl/files/weather@mockturtl/applet.js
WammKD: It wasn't; I'm not entirely certain how `bindProperty()` works but the applet worked as so when I tried it out on my machine. I just tried it again with parentheses and it, likewise, worked as it was intended so I can definitely add them in, if that's preferred.

jaszhix: You don't need that method there on line 348, it won't do anything. Lang.bind binds ```this``` to the method's context when its invoked from the signal event, so you don't need to call it again.

WammKD: Gotcha; that makes sense. I've removed the call, now.

weather@mockturtl/files/weather@mockturtl/applet.js
```diff
@@ -328,6 +338,14 @@ MyApplet.prototype = {
    }
 
   // Override Methods: Applet
+, _onKeySettingsUpdated: function _onKeySettingsUpdated() {
+    if (this.keybinding != null)
+      Main.keybindingManager.addHotKey(UUID,
```
mockturtl: As a matter of style, I would prefer braces around this statement since it spans multiple lines.

mockturtl:
I'm ambivalent about this feature.  I don't think I would use it myself, but I don't oppose it.

How does it interact with Cinnamon's key bindings? _(Settings -> Keyboard -> Shortcuts)_  I worry a bit that an applet shouldn't silently override a setting there.  Maybe this feature belongs in Cinnamon itself?

WammKD:
The biggest advantage is for mouseless usage, which is what I tend to lean towards.

Understandable; it will favor the applet that the key is bound to and, if you don't quit the applet, launch the application likewise attached to that key binding in Cinnamon's key bindings. It's a really valid point I hadn't thought of and I'm all for Cinnamon implementing a way to call applets via keyboard command. There just isn't any way I'm aware of, currently, for Cinnamon and having to suddenly switch to the mouse only for applets (when the largest boon of them, to me and why I use Cinnamon, is the settings/info. being right at your panel and not in a separate window to open and then close) was aggravating.

weather@mockturtl/files/weather@mockturtl/applet.js
WammKD: I updated with curly braces.

collinss:
@mockturtl Cinnamon does not currently have a mechanism to make sure these hotkeys do not conflict with each other, though it probably should. Still, this is the intended method for adding hotkeys to applets. If you're worried about key binding conflicts, you could always set the default value to nothing(```""```), and then the user can set their own.
Anyway, are you OK with this being merged now?

mockturtl:
> Still, this is the intended method for adding hotkeys to applets.

:+1: LGTM

Thanks @WammKD!

PR #175, opened 2017-03-07
Add Chinese translation for GPaste Reloaded and Sticky Notes

giwhub:
Add Chinese translation for GPaste Reloaded and Sticky Notes.


collinss:
@feuerfuchs are you ok with this new translation?

Feuerfuchs:
I am! Thanks for the translation @giwhub. :)

giwhub:
You're welcome.

collinss:
@giwhub Thanks for using separate commits for each applet. In the future it might be helpful to use a separate pr for each author. It was fine here because the other author was me, but the more authors that have to approve, the longer it takes, and the harder it is to keep track.

giwhub:
Thanks to let me know that,I will use separate commits for each applet next time.

PR #197, opened 2017-03-12
Add Chinese translation for Network usage monitor

giwhub:
Network usage monitor @clefebvre: Create zh_CN.po.


collinss:
@giwhub you don't need to include the author in the title, just in the comments. Also, it can't be next to other characters like you did above or it won't recognize it as a mention. To be sure, you can hit the Preview tab before submitting your PR or posting a comment, and if you did it right, the mention should become a boldface link to the person's profile page like so:
@clefebvre

giwhub:
Okay.

PR #202, opened 2017-03-13
Add Chinese translation for Host Entries

giwhub:
Host Entries @hoffis-eck.de: Create zh_CN.po.


collinss:
Since we can't determine the author's Github name, I'm going to merge this.

PR #217, opened 2017-03-16
Add Chinese translation for Stopwatch for Cinnamon 1.8+

giwhub:
stopwatch @pdcurtis: Create zh_CN.po


pdcurtis:
I have no problems with this commit but I have to admit I am not be able to check it!

PR #247, opened 2017-03-22
Cinnamenu@json: add l10n template, translation updates

NikoKrause:
* update translations against template
* complete German translation

@jaszhix


jaszhix:
This looks great, thanks for helping with the translation. 

Feel free to merge.

Odyseus:
Hello, everybody.

@NikoKrause: A little observation. You don't need to call **Gettext.bindtextdomain** on all files. With just calling it once in the first js file that is loaded is enough. In the case of this applet, I guess that would be the applet.js file.

NikoKrause:
@Odyseus 
Better?

Odyseus:
@NikoKrause 👍

PR #254, opened 2017-03-24
Update ru.po Icing Task Manager

Faust30000:


jaszhix:
I don't mind this being merged, but can I be notified so I know to update it in the src dir? That's how my build script works, it watches the src dir and copies to IcingTaskManager@json

PR #269, opened 2017-04-02
workspace-grid: Fixed BarIndicatorStyle not being draggable

robertdhernandez:


brownsr:
ping @hernejj

brownsr:
@robertdhernandez Hi Robert, I've flagged this to the original author in case he/she wishes to reply.

brownsr:
No reply from author, suspect applet is abandoned.  So will pull this now.

hernejj:
This is not abandoned, the author is just busy with other things at the moment :)

brownsr:
Delighted to hear from you @hernejj I always liked this applet.  Glad to hear that you are still keeping tabs on things.  Hope merging this was OK with you, it seemed to be well behaved when I tested it.

hernejj:
This change makes sense. I finally realized what is going on here. CInnamon has internalized the GIt repos for all applets. I think this makes a lot of sense because now anyone can fix things and revamp abandoned applets. 

Thanks for looking after this change in my absence :)

Issue #115, opened 2017-02-13
classicMenu@dalcde: not working


sepatel:
When adding the applet, even when restarting cinnamon (alt+f2, 'r', <return>), the "Menu" applet sits there doing nothing. Every click on it produces the following output in the .xsession-errors log file.

```
(cinnamon:2316): Cjs-WARNING **: JS ERROR: Exception in callback for signal: open-state-changed: TypeError: this.applicationsBox is undefined
MyApplet.prototype._onOpenStateChanged@/home/sepatel/.local/share/cinnamon/applets/classicMenu@dalcde/applet.js:133
_emit@resource:///org/gnome/gjs/modules/signals.js:140
PopupMenu.prototype.close@/usr/share/cinnamon/js/ui/popupMenu.js:2404
PopupMenuManager.prototype._closeMenu@/usr/share/cinnamon/js/ui/popupMenu.js:3610
PopupMenuManager.prototype._onEventCapture@/usr/share/cinnamon/js/ui/popupMenu.js:3599
```

My system information in case it is triggered by hardware/drivers

```
$ inxi -Fxz
System:    Host: calamity Kernel: 4.8.0-36-generic x86_64 (64 bit gcc: 5.4.0)
           Desktop: Cinnamon 3.2.7 (Gtk 3.18.9-1ubuntu3.1) Distro: Linux Mint 18.1 Serena
Machine:   Mobo: Micro-Star model: MS-16K2 v: REV:1.0 Bios: American Megatrends v: E16K2IMS.308 date: 11/02/2016
CPU:       Quad core Intel Core i7-7700HQ (-HT-MCP-) cache: 6144 KB
           flags: (lm nx sse sse2 sse3 sse4_1 sse4_2 ssse3 vmx) bmips: 22400
           clock speeds: max: 3800 MHz 1: 799 MHz 2: 799 MHz 3: 799 MHz 4: 799 MHz 5: 799 MHz 6: 799 MHz
           7: 799 MHz 8: 799 MHz
Graphics:  Card-1: Intel Device 591b bus-ID: 00:02.0
           Card-2: NVIDIA Device 1c20 bus-ID: 01:00.0
           Display Server: X.Org 1.18.4 driver: nvidia Resolution: 1920x1080@60.00hz, 1920x1080@60.02hz
           GLX Renderer: GeForce GTX 1060/PCIe/SSE2 GLX Version: 4.5.0 NVIDIA 375.26 Direct Rendering: Yes
[...]

PR #274, opened 2017-04-04
classicMenu@dalcde: applet removal

NikoKrause:
remove broken and abandoned menu applet

ping @dalcde

closes https://github.com/linuxmint/cinnamon-spices-applets/issues/115


dalcde:
I would agree with this

PR #283, opened 2017-04-05
feeds@jonbrettdev.wordpress.com

jake1164:
Update to manage feeds and allow multiple instances.


jake1164:
This is sorta a rhetorical question at this point since the work has been done, but what is the stance on python 2 vs python 3 vs supporting both?  I went ahead and continued to support both, however, its a serious PIA with handling encoding (utf-8) in python 2 vs python 3.  I would love to just support python 3, but will that make the applet unavailable for any users or does everybody have python 3 at this point?

collinss:
@jake1164 Python 3 should be fine. Many of our apps have already been ported. Xlet settings uses python 3 for example, though Cinnamon-settings uses python 2. Bottom line - any system that uses Cinnamon should have python 3 installed.

NikoKrause:
30 commits o.O
Also some of those are useless, cause they didn't changed nothing.

muzena:
@jake1164 
Feeds aplet don't work.
I think, the path in line 30 or 31 applet.js file is incorrect.
when i changed  line 30 
const DATA_PATH = GLib.get_home_dir() + "/.cinnamon/" + UUID;
to
const DATA_PATH = GLib.get_home_dir() + "/.cinnamon/configs/" + UUID;
then applet work. 
Also applet won't remember manually add feeds, after restarting PC they dissapear.

PR #287, opened 2017-04-06
netusagemonitor@pdcurtis (NUMA) Update to version 3.1.0

pdcurtis:
* Version numbering harmonised with other Cinnamon applets and added to metadata.json so it shows in 'About...'
 * icon.png copied back into applet folder so it shows in 'About...'
 *  Add translation support to applet.js and identify strings
 * Changes to remove leading and trailing spaces and replace with fixed spaces
 *  Add po folder to applet
 * Create batterymonitor.pot using cinnamon-json-makepot --js po/batterymonitor.pot
 * Version and changes information update in applet.js and changelog.txt
 * Update README.md (2x)


pdcurtis:
@NikoKrause @Odyseus and others who might be able to help.

I am trying to harmonise my Applets with other Cinnamon Applets, (the purpose of this pull request) and I have one outstanding area with harmonising this applet where I would appreciate some help.

The right click (context) menu is recreated at various times as it contains a list of the current network interfaces and which are active.

This means that the three 'standard' entries for Configure.. About... and Remove are no longer added automatically and should be added in my rebuilds  of the context menu.

I have always replaced the Configure entry but I am have difficulty in finding how to make the About.. and Remove..   entries. I have only found a couple of examples and they were very specific and I have not been able to make them work probably because I am not correctly picking up the parameters required or adding the correct `const` entries.

Any suggestions on how to do this seemingly simple task would be welcome.  

In addition how do the automatically added entries get translated and how do I make my replacements consistent.

collinss:
@pdcurtis There are a couple of options. The applet class has the function ```finalizeContextMenu``` that will add them automatically for you if they're not there already. If you are using the same context menu that is created by the api, it should be simple enough to implement this. A more "proper" solution would be to create a ```PopupMenu.PopupMenuSection``` in the context menu. You can than use that to hold your menu items and then just refresh that instead of the whole menu. This one might take a little more work, but still shouldn't be too hard.

Odyseus:
Hello, everybody.

@pdcurtis: I would use a `PopupMenu.PopupMenuSection`, like @collinss suggested. It will allow you to keep your code almost exactly as you have it now with just some tweaks. The following code will add all your menu items to the applet context menu without touching the default items.

```javascript
    // Build right click context menu
    buildContextMenu: function() {
        // Not needed.
        // this._applet_context_menu.removeAll();

        if (this.myMenuSection)
            this.myMenuSection.destroy();

        this.myMenuSection = new PopupMenu.PopupMenuSection();
        this._applet_context_menu.addMenuItem(this.myMenuSection, 0);

        // Rest of buildContextMenu function.
        // Items added to this.myMenuSection insted of this._applet_context_menu.
    },
```

Note that I used the index 0 (zero) to insert the menu section. When I add items to de context menu, the items are normally added at the beginning of the menu. But in the case of your applet, the menu section was added after all the default items. I don't know why that happened, that's why the forced insertion point. Maybe @collinss could shed some light on this?

pdcurtis:
@collinss @Odyseus Thank you very much for the suggestions. I think it is going to take a little time to try these out and make sure I understand them fully and that there are no unforeseen problems so I will probably go ahead with the requesting pull request as it stands and implement the menu fixing in a separate branch.

I had no idea about the the function finalizeContextMenu and had expected to have to add separate 'calls' similar to what is in Odesius@window-list-fork. This looks a more elegant solution and one which may maintain consistence through changes in cinnamon versions better.

As far as the reasoning in that part of the code: I think that originated in a much more basic form in the original Network Speed Monitor (NSM) project by adreadec who discontinued the project in the days of cinnamon 1.5 (https://github.com/andreadec/cinnamon-applet-netspeed). Since then it has developed into a much more powerful and flexible tool and I put it onto github three years ago but some of the original thinking still remains and I have adopted the approach of "If it ain't broke don't fix it"

pdcurtis:
@collinss @Odyseus Thank you again, your suggestion seems to have worked perfectly for the menu and sub menu - everything is retained during a refresh of the network interfaces.

I will run it for a couple of days just to make sure and also wait for #290 to be merged then upload as a new pull request

PR #288, opened 2017-04-07
Add a new applet for accessing cheatsheets from the panel

Centurix:
This is a new Cinnamon applet for accessing cheat/crib sheets from the panel.


brownsr:
Hi @Centurix , I found out after merging your applet that the line in metadata.json that has  dangerous  : false can cause applet scanning issues, so I removed it.  You will want to similarly remove it in your own repository I suggest.

Centurix:
Thanks, didn't realise. I appreciate the help!

Take care

Chris

PR #290, opened 2017-04-08
26 Applets: Remove dangerous flags

NikoKrause:
>The dangerous property in metadata.json should not be there. It interferes with the scanning of the applet when it is installed by the user.

pings:
@hulygun @mickaelalbertus @centurix @nooulaif @pdcurtis @hanspr @jaszhix @feuerfuchs @fmete @entelechy @axos88 @Odyseus @jake1164

### DIFF
--- a/0dyseus@CustomCinnamonMenu/files/0dyseus@CustomCinnamonMenu/metadata.json
+++ b/0dyseus@CustomCinnamonMenu/files/0dyseus@CustomCinnamonMenu/metadata.json
@@ -1,7 +1,6 @@
 {
     "uuid": "0dyseus@CustomCinnamonMenu",
     "max-instances": -1,
-    "dangerous": false,
     "multiversion": true,
     "description": "Custom version of the default Cinnamon Menu applet, but infinitely more customizable.",
     "version": "1.20",
--- a/0dyseus@DesktopHandler/files/0dyseus@DesktopHandler/metadata.json
+++ b/0dyseus@DesktopHandler/files/0dyseus@DesktopHandler/metadata.json
@@ -2,7 +2,6 @@
     "description": "Various shortcuts to handle desktop actions.",
     "max-instances": "-1",
     "multiversion": true,
-    "dangerous": false,
     "uuid": "0dyseus@DesktopHandler",
     "version": "1.06",
     "last-edited": "1473874426",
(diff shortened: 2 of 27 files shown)


pdcurtis:
I always wondered why it had been added to some of my applets and I am glad the "dangerous" is removed.

mtwebster:
@pdcurtis Those are normally added to the metadata of an *installed* applet by cinnamon-settings.  The last-edited was a way to compare time stamps with the spices website and the one in the installed folder to determine if there was an update.  During startup of cinnamon-settings, that dangerous flag would be added by a one-time scan of the applet files.

I think when this repo was made, clem just installed all of the applets via the gui, so we got all of these modifications as well.

pdcurtis:
@mtwebster Thanks for the information. Do you know what the criteria for dangerous was and are the applets likely to be scanned in the future - if so I would like to 'fix' whatever caused the flag to be set as I am sure it put people off loading my applet.

Odyseus:
Hello, everybody.

@pdcurtis: You can't control wether the "dangerous" key will appear or not on an xlet metadata.json file. If you remove it, that key will appear again the next time an xlet manager is opened. And it will be set to **true** or **false** depending on the existence of certain strings (as of now "spawn_sync", "spawn_command_line_sync", "GTop" and "get_file_contents_utf8_sync") inside any .js file on your xlets folder.

As you can see, some of the "unsafe strings" could be replaced by simply using their async counterparts. But in the case of **GTop**, I would think that there is no alternative option.

NikoKrause:
@pdcurtis: If a user installs an applet the first time, it gets scanned and the dangerous flag will be added.
You can check it. Move one of your applets folders away from the .local/share/cinnamon/applets folder (so manually uninstall), restart Cinnamon and move the applet folder back (so install manually). Maybe restart Cinnamon again, not sure. The dangerous flag should be added.

When I work on an applet and it crashes, (i.e. when I get the popup info, that the applet couldn't be loaded) the flag is added as well, after I geht it back working.

Before I create a commit, I check with `git status` if metadata.json was changed. If I didn't changed anything in there I `git checkout metadata.json`.

pdcurtis:
Thanks for the explanations. I have found one example of a `spawn_command_line_sync` which I have been able to change in code originating in vnstat@linuxmint.com which I have changed to `spawn_command_line_async` without any obvious problems. I think it was written by @clefebvre and it might be worth also changing in that applet if it does not cause any problems.

I also use GTop and any suggestions of how to get network statistics without use of GTop would be welcome. My use is  minimal namely:      
 ```
 GTop.glibtop_get_netload(this.gtop, this.monitoredInterfaceName);
        this.upOld = this.gtop.bytes_out;
        this.downOld = this.gtop.bytes_in;
```
Why is GTop a problem? I could understand GLib as you can call script files which could be malicious. Now all applets are on Github wouldn't it be better to search and individually check the small number of applets using GTop and synchronous calls?

PR #294, opened 2017-04-09
netusagemoninitor@pdcurtis (NUMA) Update to v3.1.1

pdcurtis:
* Additional PopupMenu.PopupMenuSection added  so 'standard' context menu items are retained when the menu is refreshed.
 * Update netusagemonitor.pot using cinnamon-json-makepot --js po/netusagemonitor.pot as a string was removed.
 * Version and changes information updated in applet.js and changelog.txt
 * Update README.md (2x) to acknowledge contributions from @collinss and @Odyseus for this  easy and elegant solution to the problem

### DIFF
--- a/netusagemonitor@pdcurtis/README.md
+++ b/netusagemonitor@pdcurtis/README.md
@@ -80,7 +80,9 @@ The applet is based on a well tried core from the netspeed applet and has been t
 
 The internal changes required in the applet to allow translations are being implemented but no translations are available at this time. Translations are usually contributed by people fluent in the language and will be very much appreciated. Users please note I am unable to take responsibility for the accuracy of translations!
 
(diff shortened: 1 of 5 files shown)


JosephMcc:
Just a heads up, you misspelled Odyseus.

pdcurtis:
Thanks, I have edited the PR. If its anywhere elsewhere I will edit at the next iteration as I have a change to modify a 'dangerous' synchronous to asynchronous call which ought to be a separate commit.

collinss:
@pdcurtis FYI you can always add more commits to a PR by pushing to the same branch you used to create it.

pdcurtis:
Yes I realise that one can push extra commits or even change the existing one by a forced change although I assume that is frowned on as one is changing history which has been made public.

In this case I did not want to make the change immediately as I always Iike to test on a few machines for a couple of days to make sure I have not missed anything unless it is a urgent bug fix.

collinss:
No worries, I just wasn't sure if you knew about this capability. No need to worry about the history changing, though. It's certainly not a good idea to do it in master, but we do it all the time in PRs (eg. when we need to rebase, or squash our commits after completing a WIP). I understand that some people are quite adamant about that, but IMO it's better to have a clean history when merging a pr.

PR #296, opened 2017-04-10
sound-with-apps-volume: added russian translation

Faust30000:
@koutch are you okay with this change?


Koutch:
Hi @Faust30000 I'm okay ^^, and I'm glad to seen my applet still in used
as you can see it, it's been a while I don't work on my applets, and I still have no time for it. So don't wait my opinion if you want to work on it, or you'll may wait a long time ;).

PR #298, opened 2017-04-10
Cinnamenu: Update to 1.2.0

### LINKED ISSUE #185: Feature Request: Ability to disable 'Bookmarks' listings in Cinnamenu
I have hundreds of bookmarks. Cinnamenu pauses for nearly 6 seconds every time I roll-over the BOOKMARKS menu item.

Request:
Please give us an option to disable/ignore bookmarks and remove them from Cinnamenu. 

Thank you!

jaszhix:
* Added ability to toggle bookmarks
* Increased the resolution of the applet icon
* Category buttons are now deactivated while searching
* Translation file restructuring thanks to [NikoKrause](https://github.com/linuxmint/cinnamon-spices-applets/pull/247)
* Test bug fix for users encountering the menu not displaying when clicked
* Fixed the menu expanding in height and width when toggled open, and expanding beyond its allocated dimensions

Closes #185


NikoKrause:
I see you replaced every `_("translatable string")` with _('translatable string').

The problem with that is, that if you create/update the translation pot file with `cinnamon-json-makepot` command, those strings won't be recognized.

You can try and compare. I don't know, if it's a bug or a feature.

jaszhix:
That seems like a bug that should be fixed in cinnamon-json-makepot. Single quotes strings look a lot cleaner to me. When CJS uses mozjs38, you may also need to account for string template syntax which looks like

````js
`This string has a ${variable}`
```

NikoKrause:
Not sure about the variable stuff. Normally there's something like that:
`_("Failed to launch '%s'").format(this.name)`

Also with `%d` for numbers.

Just saying, that for now, if you change " to ' you won't be able to update your pot file with the makepot script.

I saw @Odyseus fixed the script to be able to scan subdirectories (for multiversions). It's merged and will be available in LM 18.2.
Not sure if he did a bugfix for that as well, though.

jaszhix:
Looks like a typo maybe? https://github.com/linuxmint/Cinnamon/blob/cd180f074d5d450f3442613b4827f7aa4c987cd5/files/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py#L134

Shouldn't the language be set to JavaScript on xgettext? Unless there's a nuance I'm missing.

JosephMcc:
The convention that was established by gnome for js stuff was to single quote non-translatable strings and double quote translatable ones. It makes them easier to search. We don't always follow it but probably should.

NikoKrause:
Ah, it's a feature!

jaszhix:
Ok, I'll revert it later today. It wasn't intentional, I just didn't diff the spices repo's version with what's on my own. I think it makes the JS look inconsistent IMHO.

NikoKrause:
You can make it look consistent. But then you loose the feature to autocreate the makepot.

And since it's a feature and not a bug in makepot it's unlikely to get changed.

jaszhix:
Couldn't you accomplish searching for translated strings by searching for an underscore? As I understand it xgettext used to not have a JavaScript option when this line was added, so it seems like a bug.

JosephMcc:
> Couldn't you accomplish searching for translated strings by searching for an underscore?

Yes probably but many function calls use underscores. Essentially most introspected code. It's not a huge deal but just a convention, and imho, a sensible one.

jaszhix:
If you use an exact match (click the quote button in Sublime for instance), it will take you to those translated lines. Maybe this is the consensus in GNOME land, but I don't think its a very universal convention.

JosephMcc:
Maybe not. I think the idea though was to help find string not properly wrapped by _(). In any case I don't feel strongly enough about it to fight for it :)

Odyseus:
Hello, everybody.

@jaszhix: When I first started using the `cinnamon-json-makepot` command and saw the `--language=C` argument, I thought about changing that, so I investigated the reason behind its use. When I discovered the why, I simply desisted of the idea of changing it to avoid useless discussions and headaches. So I created my own version of the cinnamon-json-makepot.py script with the correct argument for js files (`--language=JavaScript`), plus I added to it support to scan Python files and to blacklist keys found on the settings-schema.json file.

Using the wrong language for the `--language` argument can bring unforeseen complications. It might throw warnings that shouldn't throw and **it will** insert wrong comments into the .pot files for one. Using the wrong language to scan Python files can throw actual errors, even from strings found inside comments. ¬¬

As for following conventions, we should be following Cinnamon's conventions, not Gnome's. [And Cinnamon uses `--language=JavaScript` argument for `xgettext` to scan JavaScript code for the generation of the cinnamon.pot file](https://github.com/linuxmint/Cinnamon/blob/cd180f074d5d450f3442613b4827f7aa4c987cd5/makepot#L6).

jaszhix:
Ah, I see. I would open a PR now and try to fix it but I'm not sure how to test things in other locales at the moment - I'd rather not push anything I can't test. If I were you I'd PR it and see what happens.

PR #304, opened 2017-04-11
Add Chinese translation for Download and upload speed

giwhub:
@cardsurf


NikoKrause:
@giwhub 
I noticed you added in a few PR the following line in the settings-schema:
`Settings for download-and-upload-speed@cardsurf`

The line "Settings for %applet" is added automatically to the applets settings. I noticed it as well, that it's not translated, therefore I created a fix for Cinnamon itself (https://github.com/linuxmint/Cinnamon/pull/6408/files).

It's merged and will be available in Cinnamon 3.4 (Linux Mint 18.2).
So the line you added won't be necessary when Cinnamon 3.4 comes out.

cardsurf:
I approve the changes.

giwhub:
I see,thats good,thanks for telling me that.

PR #305, opened 2017-04-11
 temperature@fevimu: Degree symbol

Ilis:
Replace unspaced unicode upper O symbol with real degree symbol and spacing between numbers and °C
https://en.wikipedia.org/wiki/Celsius


Ilis:
@fevimu

PR #315, opened 2017-04-16
Update Chinese translation for rancher@centurix

giwhub:
@centurix


NikoKrause:
@giwhub
The header infos appear twice. Can you fix it?

giwhub:
You're welcome,Centurix.
And thanks for NikoKrause,I was too careless.

PR #316, opened 2017-04-16
Update pot file and add Chinese translation for Feeds Reader

giwhub:
@jake1164


jake1164:
Wow, very nice work..

Thanks!

giwhub:
You're welcome.

PR #354, opened 2017-04-19
Translating to Russian language  "Sticky Notes by collinss"

satory-ra:


NikoKrause:
@satory-ra 
How did you translated it?
I see two problems.
1. The "fuzzy" comment should be deleted
2. Russian has 3 Plural forms, your file says `"Plural-Forms: nplurals=2; plural=(n != 1);\n"`

When you use a proper editor like poedit, this should be done automatically, when Russian language selected.

satory-ra:
I translated it through the form in the browser.
I installed your stickers. I saw that it performs a certain function. Translated: I chose the most clear concepts for Russian-speaking users.

satory-ra:
You can fix this line like this. Maybe it will be more correct.
```
#: applet.js:206
msgid "Change theme"
-msgstr "Выберите тему"
+msgstr "Выбор темы"
```

NikoKrause:
Ah. I don't mind the translation itself. Only the Informationstechnik provided in the header.

It would be better to translate with poedit.

satory-ra:
I installed poedit, I will now rewrite the translation.

PR #386, opened 2017-04-22
rancher@centurix: use proper language IDs & remove en_GB.po

NikoKrause:
…since it's not translated at all. 
ping @centurix


NikoKrause:
@Centurix I'll noticed a `.gitkeep` file in your .po directory. It's empty. Has it any purpose?

Centurix:
Must have missed that, there's no purpose to the .gitkeep. Thanks!

PR #401, opened 2017-04-23
Add translation support for Countdown timer with notifications

giwhub:
@axos88


NikoKrause:
@giwhub
One suggestion:
Don't start or end translations with whitespaces.
Not every translator does watch out for such details. Also try to include only the necessary Information in translations, e.g. you habe to translate "Minutes" 3 times.
_(" Minutes and ")
_(" Minutes ")
_("Minutes: 0")

Try something like this instead:
" " + _("Minutes") + " " + _("and") + " "
" " + _("Minutes") + " "
"_("Minutes") + ": 0"

That's a little bit ugly in the code, but the translator has less to translate and he has not to watch out for whitespaces.

Only a suggestion though.

giwhub:
Thanks a lot, it's really helpful.

axos88:
There have been a few merges in the original repository regarding translations. I will have to merge them in here, and figure out how to integrate this afterwards.

PR #412, opened 2017-04-23
Add translation support for Workspace Name

giwhub:
@willurd


NikoKrause:
The .pot file has no header

giwhub:
Added

PR #417, opened 2017-04-23
printers@linux-man: use proper language ID

NikoKrause:
use "pt" ID for Portuguese (Portugal)
deleted "pt_BR", because there was no difference at all to "pt"

@linux-man


NikoKrause:
@linux-man 
What is the purpose of the `POTFILES.in` file?

linux-man:
No purpose for the normal applet functioning. It's used to create the pot file for translations.
pt_BR file was waiting for possible changes regarding the minor differences between Portugal Portuguese and Brazilian Portuguese (believe me, we have bigger differences than color vs. colour), but it's probably not needed. Can be recreated if some Brazilian propose a different translation.

NikoKrause:
You can create/update the potfile with
```
cinnamon-json-makepot --js po/printers@linux-man.pot
```

Yeah, I don't mind pt_BR being in there as well. Just wondered. Then I'll merge as it is?

linux-man:
I used `intltool-update --pot` that expects a `POTFILES.in`.
The deletion of pt_BR is really a non-issue. I'm OK with it.

muzena:
@NikoKrause 
Why in some spices
cinnamon-json-makepot --js po/printers@linux-man.pot
or
cinnamon-json-makepot -i po/printers@linux-man.pot
commands don't work.
I try in this applet and I get error:

$ cinnamon-json-makepot --js po/printers@linux-man.pot
xgettext (GNU gettext-tools) 0.19.8.1
Copyright (C) 1995-1998, 2000-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Written by Ulrich Drepper.
 
Running xgettext on JavaScript files...
Traceback (most recent call last):
  File "/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py", line 275, in <module>
    Main()
  File "/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py", line 130, in __init__
    tmp = tempfile.NamedTemporaryFile(prefix="cinnamon-json-makepot-")
NameError: global name 'tempfile' is not defined

NikoKrause:
Not sure what happens. It works for me.
Since you are running Cinnamon nightlies that's maybe a new bug.
@Odyseus?

Odyseus:
Hello, everybody.

@muzena: You have to provide a file name, not a path. The correct command is as follows:

```shell
cinnamon-json-makepot --js filename
```

The .pot extension also wouldn't be needed, it is automatically added. Although, it isn't an error to add it.

To install localizations you don't need to specify a location to the `cinnamon-json-makepot` command. Running the `cinnamon-json-makepot` with the `-i` argument from the xlet folder is enough.

```shell
cinnamon-json-makepot -i
```

NikoKrause:
@Odyseus 
A path worked for me. So that shouldn't be an issue.

Odyseus:
@muzena: It was an error on the cinnamon-json-makepot script. I just submitted a fix for it. You can temporarily fix it by editing the `/usr/share/cinnamon/cinnamon-json-makepot/cinnamon-json-makepot.py` file and add **import tempfile** to the list of imports at the beginning of the code.

muzena:
@Odyseus 
Thanks, now it works.

Issue #152, opened 2017-02-23
vnstat@linuxmint.com doesn't detect wireless device


NathanSalapat:
Using the latest version of the network usage monitor (vnstat@linuxmint.com) and cinnamon 3.2.7 the wireless interface is never properly selected. Ifconfig reports the wireless as being wlp3s0 on my laptop. I looked at the code and found that a fallback is used as eth0. It would appear to me as if the error is in _updateDevice: function() starting on line seventy. I unfortunately do not know javascript or I could probably be more helpful.

Also I feel like I should mention that this works fine on my desktop which has a wireless interface of wlp5s2.

NikoKrause:
Can you take a screenshot of the applet?
Also what's the output of `vnstat`?

NathanSalapat:
I manually corrected the applet so it works, by simply entering the wlp3s0 string where it calls the variable that should have the interface name, but it would just display a small gray box, without the image created by vnstati.

vnstat output is this:
                      rx      /      tx      /     total    /   estimated
 wlp3s0:
       Jan '17     78.42 GiB  /   97.28 GiB  /  175.70 GiB
       Feb '17     29.95 GiB  /   79.02 GiB  /  108.97 GiB  /  136.28 GiB
     yesterday    814.86 MiB  /    3.34 GiB  /    4.13 GiB
         today    450.38 MiB  /    1.72 GiB  /    2.16 GiB  /    5.56 GiB

 enp2s0: Not enough data available yet.

NikoKrause:
Could you use a non modified version of the applet and create a screenshot?

NathanSalapat:
Ya, I can do that, could try and get the output from looking glass too. Not at my computer now, so will be several hours.

NikoKrause:
The first time I used the applet, I had to set up vnstat. I used this tutorial
https://www.howtoforge.com/tutorial/vnstat-network-monitoring-ubuntu/

I have wlp5s0. After ging through the steps there, it showed up for me.

NathanSalapat:
![vnstat](https://cloud.githubusercontent.com/assets/8930456/23289221/f8b1dcf6-fa0d-11e6-8a58-085cebc120e7.png)

Here is a picture of the network monitor when it's not working. I checked out that link you provided, and everything seems to be just fine. I can run the vnstat command without any args and get the output for my active interface.

[...]

PR #420, opened 2017-04-23
vnstat@linuxmint.com: use proper language ID

NikoKrause:
@clefebvre


pdcurtis:
@NikoKrause @clefebvre 
Whilst working on the applet is it worth changing
 ` GLib.spawn_command_line_sync('vnstati -s -ne -i ' + this._device + ' -o /tmp/`
to the async version if only to avoid the "dangerous" label. I did so a while back on netusagemonitor@pdcurtis which had 'borrowed' the code.

clefebvre:
@pdcurtis yes definitely. Please feel free to PR for this.

pdcurtis:
@clefebvre In testing the change  `GLib.spawn_command_line_sync('vnstati -s -ne -i ' + this._device + ' -o /tmp/)`
to the async version I found that the applet did not work a issue had already been raised #152
I have been unable to solve it although I narrowed it down to a small section of one function.
I have replaced the function with   a cut down version of the one used in    `networkusagemonitor@pdcurtis` and provided PR #489 for your consideration as a temporary fix until you have time to correct your own more elegant function.

PR #429, opened 2017-04-24
Add translation support for ACPI

giwhub:
@bro

### DIFF
--- a/acpi@bro/files/acpi@bro/applet.js
+++ b/acpi@bro/files/acpi@bro/applet.js
@@ -2,10 +2,16 @@ const Lang = imports.lang;
 const Applet = imports.ui.applet;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
-const Gettext = imports.gettext.domain('cinnamon-applets');
-const _ = Gettext.gettext;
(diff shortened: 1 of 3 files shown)


NikoKrause:
Did you tried this translation?
I'm not sure, but looking at the code it seems that you translate icon-names.

Normally the icon-names are stored in English. So if you translate them, the applet wouldn't find them.
Habe to check this on my machine yet.

giwhub:
I thought they are names of battery status,just now I viewed the code make I know they are icon names.I didn't test this one as I am on PC.Thanks.

NikoKrause:
What about the state stuff?

giwhub:
This applet shows blank on my pc so I am not able to test it,I was thinking it is only working on laptop.

pdcurtis:
@giwhub @NikoKrause I think it would be sensible to ask for the authors approval for all translations where the applet has not been tested or the translation strings are being identified. 
Great work but could end up with unintended consequences when you identify strings and create the .pot files rather than just an extra .po. 
Another thing to watch is that some applets get the UUID from metadata.uuid already so not a good idea to add a `const UUID ` even if scoping saves the day.
Keep up the great work

PR #414, opened 2017-04-23
resize and make icons square

MrElectroNick:

PR #438, opened 2017-04-25
PDFManager@cinnamon.org: resize and make icon square

NikoKrause:
needs approval by @mickaelalbertus 

original pull request https://github.com/linuxmint/cinnamon-spices-applets/pull/414/files by @MrElectroNick


mickaelalbertus:
I agree with icon changing :)
I don't know where to say it but there is a bug for foreign accents for item in menu ?

NikoKrause:
@mickaelalbertus 
What do you mean? Can you explain the bug any further? (Example, screenshots?)

mickaelalbertus:
You're right. Here a screenshot:
bug_pdf_manager

The french translation for reduce the size of PDF isn't correct with accent. I don't know if it's related to my computer or not. Maybe if someone have the time to see what's wrong?

NikoKrause:
@mickaelalbertus 
This is a bug in the character encoding of the translation file.
It should use UTF-8.
I created a pull request.

PR #449, opened 2017-04-25
bumblebee@pdcurtis: resize and make icon square

NikoKrause:
needs approval by @pdcurtis

original pull request https://github.com/linuxmint/cinnamon-spices-applets/pull/414 by @MrElectroNick


pdcurtis:
@MrElectroNick @NikoKrause 
The same icon is used in the applet folder so it displays in the Context Menu About... and that should also be changed.
Once that is done I approve

NikoKrause:
@pdcurtis better?

pdcurtis:
@NikoKrause Many thanks

PR #450, opened 2017-04-25
bumblebee@marmistrz: resize and make icon square

NikoKrause:
needs approval by @marmistrz

original pull request https://github.com/linuxmint/cinnamon-spices-applets/pull/414 by @MrElectroNick


pdcurtis:
@marmistrz  As the original author of the applet you forked I am interested to know if you intend to continue to support this version as I am thinking of modifying bumblebee@pdcurtis to have a simpler mode for vertical panels and an option of just an icon to save space in the horizontal mode.

marmistrz:
Sorry for the delay, I missed the PR.
@NikoKrause I'm ok with changing the icon. It will basically be only displayed in the applet mgr
@pdcurtis I'm no longer using Bumblebee since I switched to nouveau, so I probably won't be able to really test it. I may try merging your changes. 
Anyway, I think it may be a better idea to integrate what I did my fork into upstream, so that the user can choose the UI flavor. For me, this UI was more visually appealing, but I think it's best to give the user a choice.

NikoKrause:
@pdcurtis @marmistrz 
So what are the plans for this applet? Deleting or keeping?
If pdcurtis wants to combine those and you don't want to maintain it anymore?

pdcurtis:
Hi,  My views are:
 * The current situation is confusing as we have two aplets with the same icon and similar names, especially as one has no README.md to distinguish it, so something needs to be done.
 * My original applet was written partially to satisfy my requirements which gave priority to the GPU temperature rather than a simple indicator which GPU was in use.
 * My applet needs modification to work with a vertical panel and would then have to use a graphic indicator of the GPU in use (as per the fork) and the Temperature would be via the tooltip and left click menus. It could be a user choice to use the same on a horizontal panel.
 * The fork by @marmistrz would also need changes to work on a vertical panel and needs Translation support etc as well as a README.md 
 * I intend to modify several of my applets to work with a vertical panel and Bumblebee and the Battery Monitor with Alerts and Shutdown (BAMS) are top of my list but on a time-scale of weeks not days. 
In summary:
 * In a month or two I should be able to offer a choice of the two UIs and the fork by @marmistrz would not be essential. I can also credit @marmistrz with the idea of adding the simpler GUI.
 * If however @marmistrz wants to continue his applet I am prepared to check it works if he no longer uses Bumblebee.
* I am also happy to receive any contributions to improve mine!

@NikoKrause  What are the teams view on overlapping applets bearing in mind small changes are a good way to introduce new blood into applet writers?

NikoKrause:
That's not confusing. There are a bunch of similar applets with similar names and even with the same name, e.g. System Monitor or Gmail Notifier. ;)

There are a few applets, which are very similar. Some are forks of applets, which were broken. If there are applets, which have just been created for fixing existing broken applets, but do not have any further changes, the working one is kept and the broken one will be deleted.
E.g. Messaging Menu, Messaging Menu v2, Messaging Menu v3

Those have to be tested and it needs to be decided individually.
If you detect broken applets, which can be deleted because of those reasons or because there are other/better options let us know.

NikoKrause:
I'm merging this now

marmistrz:
@pdcurtis I stopped using Bumblebee (currently using nouveau's DRI_PRIME instead) so feel free to merge the functionality into my applet.

pdcurtis:
I have already merged the useful ideas from @marmistrz (including a option of the simplified UI) into the original applet as well as extending it to work on vertical panels so I see little point in further development of the 'simple' version by me.

NikoKrause:
Then we should remove it, if pdcurtis version has the same functionality.

NikoKrause:
@pdcurtis @marmistrz So what was the conclusion with the bumblebee@marmistrz applet?
Shall I remove it?

pdcurtis:
My views have not changed.

I see no point in its retention as I have already merged in the useful ideas from @marmistrz (including a option of the simplified UI) into the original applet as well as extending it to work on vertical panels so I see little point in further development of the 'simple' version. It is however not for me to authorise its removal.

PR #454, opened 2017-04-25
system-monitor@spacy01: use UTF-8 character encoding

NikoKrause:
@spacy01


NikoKrause:
@spacy01 
Watch out for charset, or else you can get a similiar error like this https://github.com/linuxmint/cinnamon-spices-applets/pull/438#issuecomment-297023831

I'll recommend you to use a specific editor for .po files, e.g. poedit

PR #457, opened 2017-04-25
A suggesion on easier way :D

znav:
Isn't there an easier way to do this?
There should be. Open to suggestions on that.


znav:
https://github.com/znav/zDict

PR #474, opened 2017-04-26
location-detection@heimdall: add ability to set API key in Settings

NikoKrause:
@heimdall
Added settings to applet, so the user can easily register on IPInfoDB and set/use his own API key.
And updated translation files.

### DIFF
--- a/location-detection@heimdall/files/location-detection@heimdall/applet.js
+++ b/location-detection@heimdall/files/location-detection@heimdall/applet.js
@@ -3,13 +3,11 @@ const Soup = imports.gi.Soup;
 const Lang = imports.lang;
 const Json = imports.gi.Json;
 const Mainloop = imports.mainloop;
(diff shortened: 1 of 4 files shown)


NikoKrause:
@BrainAxe
Are you using this applet and do you have a valid API-key? If yes, can you check this version. I'm too lazy to register on IPInfoDB ;)

BrainAxe:
@NikoKrause 
Thanks for asking. I am still using this applet and I tested this piece of code with my valid API-key. I think this is a nice update. I am giving some screen-shot of the testing .
screenshot_00000
screenshot_00002
screenshot_00001

NikoKrause:
Thanks for testing

NikoKrause:
This applet is abandoned, the author does not respond and author was set to `none`.
So I'm going to merge this.

PR #171, opened 2017-03-06
jenkins@backuity.org: Handle other build statuses from Jenkins, such as disabled and aborted

vitaliis:
Previously everything which wasn't a success was displayed as a failure.
This was incomplete as there are other statuses for jenkins jobs possible.
This commit makes it possible to differentiate between failed and disabled jobs

PR #475, opened 2017-04-26
Several applets: Set author with no corresponding github user name to none

NikoKrause:
The author names do not respond to any github user name or the authors do not respond at all, therefore set author to 'none' and give credit to them as original author.

### DIFF
--- a/betterlock/info.json
+++ b/betterlock/info.json
@@ -1,1 +1,4 @@
-{"author": "entelechy"}
+{
+    "author": "none",
+    "original_author": "entelechy"
+}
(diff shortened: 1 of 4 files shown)


pdcurtis:
@NikoKrause This looks a nice way round the problem of missing or disinterested authors where the applet is worth keeping and the original author's name is not built in as part of the uuid. It can easily be undone if the author returns.
Is there (or should there be) a mechanism to assign a new person(s) to be responsible and review any PRs short of actually making them the author? At a minimum they should be a willing continuous user capable of understanding the code.

NikoKrause:
Here is an example of an author who didn't wanted to maintain his applet anymore and which now has a different maintainer now: https://github.com/linuxmint/cinnamon-spices-applets/pull/171#issuecomment-285474956

I think if someone sees an applet without any maintainer, he can ask or created a PR to be the maintainer.

PR #484, opened 2017-04-27
Update translation for Network Data Usage Monitor

giwhub:
@pdcurtis

### DIFF
--- a/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/applet.js
+++ b/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/applet.js
@@ -671,7 +671,7 @@ Note Odysius has used the index 0 (zero) to insert the menu section to position 
         this.subMenu1 = new PopupMenu.PopupSubMenuMenuItem(_("Housekeeping and System Sub Menu"));
         this.myMenuSection.addMenuItem(this.subMenu1);
 
-        this.subMenuItem1 = new PopupMenu.PopupMenuItem("Open System Monitor");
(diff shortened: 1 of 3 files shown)


pdcurtis:
Thanks - Agreed

NikoKrause:
@giwhub If it was me, you don't need 3 separate commits for that. What do you think pdcurtis?

giwhub:
How to combine commits?
I am using Github web page version.

NikoKrause:
Ah. You don't use git? I always thought you did this on purpose. 
Then nevermind. I can combine them, when I merge them

giwhub:
I don't use git often.
Git clone/git push larger sources/lots of files from/to Github is very slowly in China.

pdcurtis:
I have used git rebase with the -I option in the terminal which is very powerful after @NikoKrause suggested I combined commits.  
I have also  been looking at using an Android pad when Mobile with Mgit Droidedit and Forkhub so I am interested in any ways of working without a heavy laptop using WiFi in a pub or coffee shop. I had not thought about github on-line tools for coding but it is excellent in offering this sort of conversation. I can also see the advantages with some long collaborative developments during a PR, the vertical panel needed to merge over a hundred commits - three should be easy.

I think it very important to test even after what seems the most trivial change - I have been burnt too often.

NikoKrause:
I thought you use git and it's more work for you to create a commit each time, I just wanted you to know, that one commit would suffice.
But since it's not the case you can keep your workflow. I'll squash them on merge or not. It's no big deal. Actually it's no big deal to squash on merge. It's just pressing a different button.

PR #489, opened 2017-04-27
vnstat@linuxmint.com Patch 1

pdcurtis:
* Changes made by because `_updateDevice` function was not working - Issue #152
         `_UpdateDevice()` commented out and replaced by a cut down version of the one used in
         networkusagemonitor@pdcurtis 
        with two associated functions getInterfaces() and isInterfaceAvailable() added
        `const NetworkManager = imports.gi.NetworkManager` added
        diagnostic call `global.logError("Test not error - vnstat using device: " + this._device);` added. 
        `GLib.spawn_command_line_sync()` replaced by` GLib.spawn_command_line_async()`
 * Tested with normal connections made by network manager.
 
 * Will probably not find connections not made within Network Manager such as Bluetooth PAN connections or USB mobile internet modems.

@clefebvre I failed to find the error in the original active device identification function which sometimes worked but more often did not and always gave errors in the glass log. You may wish to use this version until you have time to debug the original routine which I am sure is more elegant.

I have commented the changes and left the original code in place but commented out and have left in one diagnostic call to `global.logError(e)` I can tidy up if required.

I also carried out the change to an asynchronous call as suggested in #420 which was all I had intended to do - it will teach me not to volunteer suggestions!

### DIFF
--- a/vnstat@linuxmint.com/README.md
+++ b/vnstat@linuxmint.com/README.md
@@ -2,13 +2,13 @@
 
 This applet keeps track of your Internet usage.
 
-In some countries, ISPs limit the amount of data their customers can use. Passed a certain amount of bits downloaded/uploaded, you either get cut off, you pay more or your connection speed is reduced.
+In some countries, ISPs limit the amount of data their customers can use. Passed a certain amount of bits downloaded/uploaded, you either get cut off, you pay more or your connection speed is reduced. The situation is even worse if you are using a mobile connection, especially when roaming.
 
 ## How it does it:
 
 The vnstat daemon runs in the background and collects info about your Internet usage.
 
-The applet detect which device you're currently using, and simply export a graph using vnstati.
+The applet detects which device you're currently using, and simply exports a graph using vnstati.
 
 ## What you need for it to work:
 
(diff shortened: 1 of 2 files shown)


clefebvre:
Hi @pdcurtis,

Thanks. Please remove the comments about old/new and the dates in the code. That information is there via git blame and git log.

pdcurtis:
@clefebvre I will remove the comments about old/new and the dates in the code and push the new version back up. I realise that the information is redundant in the long term, it was just so you could evaluate quickly and without extra work.
You had a diagnostic call to global.logError(e) in the previously version, I will comment it out unless you suggest otherwise.

pdcurtis:
@clefebvre Changes requested have been made, pushed and ready for approval.

pdcurtis:
@clefebvre Changes requested have been made, pushed and ready for approval.

pdcurtis:
Please hold this. There is still a possible issue with computers with several wifi adapters including a USB one would like to check.

pdcurtis:
* This now fixes #152.
 * I have improved the handling of various conditions such as no active device where the previous output is displayed. At no time should there be an empty display, there is always an error message, comment or image.
 * Image storage location moved to $HOME instead of /tmp to give continuity through a restart.
 * I have added an extra note in README.md about the need for extra devices such as Mobile Internet USB sticks to be added to the vnstat database configuration but not given details as it is difficult (issues with permissions).
@clefebvre Ready for your approval

NikoKrause:
@clefebvre 
It's your applet

PR #492, opened 2017-04-28
Add translation for Show or record desktop

giwhub:
@cranes-bill

### DIFF
--- a/montrer-le-bureau@cannelle.org/files/montrer-le-bureau@cannelle.org/applet.js
+++ b/montrer-le-bureau@cannelle.org/files/montrer-le-bureau@cannelle.org/applet.js
@@ -12,20 +12,21 @@ const ModalDialog = imports.ui.modalDialog;
 const GLib = imports.gi.GLib;
 const Gettext = imports.gettext;
 
+// l10n/translation support
+
+const UUID = "montrer-le-bureau@cannelle.org"
+Gettext.bindtextdomain(UUID, GLib.get_home_dir() + "/.local/share/locale")
+
+function _(str) {
+  return Gettext.dgettext(UUID, str);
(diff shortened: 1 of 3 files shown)


cranes-bill:
Thanks, but i would like if somebody could help me fix the English string first. I'm afraid due to my bad English, they are not completely correct. That is the reason why i commented the _gettext support_.

NikoKrause:
Looks mostly good to me.
@giwhub
Can you check?

giwhub:
It looks good but yet has some careless mistakes,I can help if you wish.

cranes-bill:
> It looks good but yet has some careless mistakes,I can help if you wish.


Yes, please!

Also, did you try the applet with translation support enabled, and does it works for you?
I'm asking, because if i enable the _translation code_, the applet fails to load to me. I found the reason is in _predefined message strings_. I mean:
`const LOCKMESSAGE = _("String for translation") + "\n";`
When i write those messages directly into the function, everything work fine. So if that also does not works for you, it must be changed. 

P.S.: Sorry for the bad English!

giwhub:
screenshot-window-2017-04-28-220606

I retry it,it does work for me.

cranes-bill:
That's a good news. So please correct the strings where you things it's necessary, then can merge it!
Thank you very much!

NikoKrause:
@cranes-bill
You try to translate a const here. Changing the value after it's init is not possible. So I'm not sure what happens, if it gets translated.

@cobinja says:
> My assumption would be that it always uses the initial value returned by _(). If the language changes afterwards, I'd say it either keeps the original value or crashes horribly

cranes-bill:
OK, i will change it.

PR #502, opened 2017-04-28
Add translation support for Launchers Draggable Switch

giwhub:
@r0p01ach


NikoKrause:
You missed a whitespace at the end of the tooltip

PR #505, opened 2017-04-28
Ink level monitor@r0p01ach hr

muzena:


NikoKrause:
@muzena
Are this Strings you added (Photoblack, etc.) visible in the applet itself?

Those are strings comparing to something.
But for me it's not clear what it compares.
If it compares to something what is actually in English and you translate it to another language, it will break the applet.

muzena:
@NikoKrause 
I try to test applet but, my printer is out of ink in toner, and maybe why don't show color status.
With these strings there is no difference, no errors has occurred as I noticed.
But before all that to applet work, I need to add printer device path #506 in line 94 applet.js file
/dev/usb/lp and /dev/usblp path do not exist on my system, I have file in/dev/lp0
Before that I get error "Unable to connect printer at port 0"

PR #516, opened 2017-04-29
weather@mockturtl: Apply 3.4 fixes

jaszhix:
@mockturtl 

```
(cinnamon:6699): Cjs-WARNING **: JS ERROR: TypeError: weather.item is undefined
refreshWeather/<@/home/jason/.local/share/cinnamon/applets/weather@mockturtl/applet.js:415:11
soupQueue@/home/jason/.local/share/cinnamon/applets/weather@mockturtl/applet.js:390:7

```

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -409,9 +409,15 @@ MyApplet.prototype = {
         // Polling for likely API throttling
         Mainloop.timeout_add_seconds(5, Lang.bind(this, function() {
           this.refreshWeather(false)
-        }))  
+        }))
       }
+
       let weather = json.query.results.channel
+
+      if (!weather.item) {
+        return false
+      }
+
       let weather_c = weather.item.condition
       let forecast = weather.item.forecast
 


mockturtl:
Cool!  I didn't know 3.4 was available.

Do I understand correctly that something will change in the parser, such that `weather` might exist, but `weather.item` not?

weather@mockturtl/files/weather@mockturtl/applet.js
```diff
@@ -409,9 +409,15 @@ MyApplet.prototype = {
...
         Mainloop.timeout_add_seconds(5, Lang.bind(this, function() {
           this.refreshWeather(false)
-        }))  
+        }))
       }
+
       let weather = json.query.results.channel
+
+      if (!weather.item) {
+        return false
```
mockturtl: I'm not familiar with JS idioms, but I would just log an error (warning?) and `return` here, or else change all `refreshWeather`'s callsites to consume a boolean result.

jaszhix:
Ah, no I'm running a nightly build with the CJS running on mozjs38, so it has broken a lot of things. I'm not sure I understand what you mean consume a boolean. The problem is the weather ajax data is not always valid because its being polled and the weather API likely has a throttling limit. So this returns when there is no weather data, after the polling is triggered again.

mockturtl:
> I'm not sure I understand what you mean by consume a boolean

You `return false`, but none of the callers use the value.

mockturtl:
> the weather API likely has a throttling limit... the old interpreter (mozjs24) ignored these undefined TypeErrors

Ah!  I see now.  Thanks!

jaszhix:
> You return false, but none of the callers use the value.

Ahh, its just a CJS-specific thing where it will log a warning about functions returning undefined, so I do it in all applet code. In other JS projects I will just use ```return``` to end the function.

collinss:
@mockturtl is this ok to merge now?

mockturtl:
@jaszhix 
>  it will log a warning about functions returning undefined

 (╯°□°）╯︵ ┻━┻ 

@collinss 

> is this ok to merge now?

Yes, thanks!

PR #571, opened 2017-05-02
Sound with apps volume@koutch.hr

muzena:


NikoKrause:
What happens with the pot in your commits?

muzena:
I accidently save in poedit

NikoKrause:
So the pot isn't changed right now.

muzena:
No!

PR #572, opened 2017-05-02
stackoverflow-questions-notifier@higuaro.hr: Add hr.po

muzena:


NikoKrause:
Why does the title says "Fix pot?"

muzena:
@NikoKrause 
Fixed.

PR #574, opened 2017-05-02
CinnVIIStarkMenu@NikoKrause: add sv.po

eson57:
Add Swedish translation. Hope this is okay, because I wasn't allowed to upload to this repo at Launchpad.


NikoKrause:
Why? It's okay. I need to close the launchpad stuff someday. It's super slow with import/export.

eson57:
Guess it's because of the restriction below, and I'm not a member of the Linux Mint translation team.

> CinnVIIStarkMenu is translated by Linux Mint with Structured permissions.

NikoKrause:
That shouldn't be a problem. You should be able to make suggestions, which need to be reviewed by the Linux Mint translation team. Anywho. It's totally fine creating PR here. Less work for me.

eson57:
Yes, suggestions made online is ok, but to upload an offline made file, you have to be a team member or the repo have to be on "open permission".
Some times I do make specific string suggestions, but I prefer offline translation (and testing) when it comes to complete new application translations.

PR #578, opened 2017-05-02
window-list-multi-monitor@mazz: applet removal

NikoKrause:
obsolete


NikoKrause:
Also the author is not the github user mazz: https://github.com/linuxmint/cinnamon-spices-applets/pull/470

PR #588, opened 2017-05-04
hwmonitor@sylfurd: Define applet scale ratio on one place

renepavlik:


NikoKrause:
@sylfurd

MrElectroNick:
@renepavlik - Good catch!

PR #604, opened 2017-05-05
[3.4] Several applets: remove duplicate 'let' declarations

NikoKrause:
mozjs38 crashes if one is ecountered within the same scope

this prevents the applet to crash in Cinnamon 3.4

@collinss 
@NikoKrause
@cardsurf
@pdcurtis 
@bownz
@dewman12
@clockworksimon
@pixunil
@nooulaif

### DIFF
--- a/CinnVIIStarkMenu@NikoKrause/files/CinnVIIStarkMenu@NikoKrause/2.6/applet.js
+++ b/CinnVIIStarkMenu@NikoKrause/files/CinnVIIStarkMenu@NikoKrause/2.6/applet.js
@@ -3834,13 +3834,13 @@ MyApplet.prototype = {
                 actor.hide();
             }
         }
-        let actors = this.categoriesBox.get_children();
+        actors = this.categoriesBox.get_children();
         for (var i=0; i<actors.length; i++){
             let actor = actors[i];
             actor.style_class = "menu-category-button";
             actor.show();
         }
-        let actors = this.favoritesBox.get_children();
+        actors = this.favoritesBox.get_children();
         for (var i=0; i<actors.length; i++){
             let actor = actors[i];
             actor.remove_style_pseudo_class("hover");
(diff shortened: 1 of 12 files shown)


pdcurtis:
Please go ahead. 
A problems with mindlessly copying code from place to place. I am glad I am not the only one. 
@NikoKrause 
 * Have you searched all applets or do I need to check my other applets?
 * Have we got details or a version of 3.4 available for authors to use for checks?

NikoKrause:
> Have you searched all applets or do I need to check my other applets?

mtwebster has written a script, which checks duplicate lets.

> Have we got details or a version of 3.4 available for authors to use for checks?

The release cycle will be as always, which is: There will be no extra version for authors. Cinnamon 3.4 has been tagged recently. If you are brave enough you can try to compile it on your own or you can try the Cinnamon nightly ppa. But be aware that you have to figure out how to repair your system on your own, if it crashes, because of missing dependencies or something else.
Bleeding edge distros as Arch or Fedora are usually delivering very fast the new Cinnamon version to their users. (Cinnamon is usually tagged a month before release of Linux Mint, to find bugs and fix them. Issue reports from bleeding edge distro users are helping.)
Then there will be a beta release and after that the stable ISOs. And last but not least the option to update from 18.1 to 18.2.

Right now I'm not using 3.4 on my own. I need my data.
As far as I know the switch to mozjs38 causes applets to crash, when they have duplicate let declaration.
There are probably other problems as well. But I don't know which ones. If issues are found, they will be addressed.

NikoKrause:
> a version of 3.4 available for authors

A version of 3.4 available for everybody (since this is open source): https://github.com/linuxmint/Cinnamon/releases

collinss:
I'm going to merge this since we already have users complaining about broken applets. If there are any issues with the patches here, they can be reverted individually.

PR #671, opened 2017-05-15
Cinnamenu@json: add sv.po

eson57:
Add Swedish translation

These lines are correct in .pot file, but do not translate:
```
#: panel.js:1800
#: panel.js:1823
#: panel.js:1579
#: panel.js:1582
```


eson57:
@jaszhix

PR #680, opened 2017-05-15
collapsible-systray@feuerfuchs.eu: Add support for vertical panels

Feuerfuchs:


jaszhix:
Not sure if you've had the chance to test it on Cinnamon 3.4, but the applet loads, but doesn't collapse the tray. It also is using the vertical arrow icon on a horizontal panel.

May be related to these warnings:

```
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/collapsible-systray@feuerfuchs.eu/applet.js 171]: reference to undefined property this.verticalCollapseIconName
```
```
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/collapsible-systray@feuerfuchs.eu/applet.js 168]: reference to undefined property this._direction
```

Feuerfuchs:
Cinnamon 3.4 hasn't been pushed to the Manjaro repos yet, unfortunately. But I'd be surprised if that was the reason for the warning.
Do these warnings also occur if you use the applet release from my repo? https://github.com/Feuerfuchs/Collapsible-Systray-Cinnamon-Applet/releases/tag/v1.7

jaszhix:
Is that version any different? I am testing the version in this PR. Alternatively you can add ```"cinnamon-version": ["3.2"]``` to your metadata.json, along with any other versions the applet supports since its probably not likely you can fix it without running 3.4.

Feuerfuchs:
No, it should be identical. I'm just trying to understand the problem and want to be sure this isn't caused by some GitHub/git wonkiness. It's baffling because the warnings you posted here point to a basic logical mistake in my applet, i.e. that this._orientation and this.verticalCollapseIconName aren't set. But, I set those right during the initialization. That means, if this fails for you due to compatibility issues, there should be tons of other variables that shouldn't be known as well and the whole applet wouldn't work in the slightest. So what this looks like to me is that some of my changes weren't applied, especially the ones in the applet's initialization stage.

jaszhix:
Cinnamon 3.4 has a newer JS engine based on mozjs38, so I'm thinking its a 3.4 issue. We have been patching applets so they run again in 3.4.

As for the warnings, maybe there's no access to ```this``` from those getters' scope?

Feuerfuchs:
Ah, I was just about to amend my comment. ^^ I took a closer look at the whole initialization phase and I think I know what the problem is: The collapse button fetches the icon during its initialization, i.e. it calls the getter function "collapseIcon" in applet.js. At this point, the settings and several variables haven't been initialized yet. The old engine wasn't bothered by that, but it looks like the new one is. :)

So I'll change that behavior and if I still can't get it working for you, I'll just update the metadata as you said earlier.

jaszhix:
Cool. I also had Cinnamon crash when removing the default system tray applet after loading yours. Log is [here](https://pastebin.com/ktCu7NeE). You may want to add that to the install notes that they should unload the default applet before adding yours to the panel. Also, if you copy/move the readme file from ```/collapsible-systray@feuerfuchs.eu/files/collapsible-systray@feuerfuchs.eu/``` to ```/collapsible-systray@feuerfuchs.eu/```, it will display on the Spices website.

Feuerfuchs:
I updated the readme, thanks for the heads-up!

jaszhix:
No problem. I think you forgot to move it to the base directory though.

Feuerfuchs:
Sooo does it still not work or why is the PR still pending?

jaszhix:
Oh, it doesn't work for me on 3.4, I was waiting to see if you we're going to push something else or edit metadata.json.

Feuerfuchs:
3cf5fe8 is he commit that should have fixed it. So it didn't? :/

jaszhix:
No. I don't expect you to fix it without running 3.4. I would just add ```"cinnamon-version": ["3.2"]``` to metadata.json so users on 3.4 are notified it doesn't work properly.

Feuerfuchs:
Bummer, I thought I had found the issue. But yeah, there's no point trying to fix it blind, so I added the compatibility info for now.

Feuerfuchs:
So the update was just published for my distro and I installed it. No issues whatsoever.

jaszhix:
This is going to be confusing for some, but Cinnamon uses CJS 1.48. When Michael rebased GJS, he made it so it can depend on both mozjs24 *and* mozjs38. Right now most users on 3.4 are using the mozjs24 JS engine, but when Mint 18.2 rolls around (and presumably users on Ubuntu 17.04+), they will be on 38, which does change how the applets are run because it uses a newer version of EcmaScript that handles a lot of syntax errors differently. This is something you can more reliably test by compiling Cinnamon from source for now, but the change will happen for a lot more users later.

NikoKrause:
When you open terminal type `cjs` and then `typeof Symbol` and it says `undefined` you are on mozjs24. I learned that from jaszhix.
```bash
user@mint ~ $ cjs
cjs> typeof Symbol
"undefined"
```

Feuerfuchs:
It returns "function" for me.

bildschirmfoto vom 2017-05-19 09-54-51

jaszhix:
Okay. I never could get the tray to collapse and saw warnings. You can open another PR and I can test again if you want.

PR #698, opened 2017-05-17
bumbleebee@pdcurtis Update to Version 3.2.0

pdcurtis:
3.2.0 is a major new version to support vertical panels and to use prioritise icons instead of text to harmonise with other cinnamon applets such as nvidia-prime and the views of @marmistrz on the UI
 * Change to TextIcon applet
 * Allow use of vertical as well as horizontal panels after version number check to see if they are supported.
 * Addition of setting to hide temperatures on Horizontal panel.
 * New Translation function
 * Changes to improve translation strings
 * Recreate bumblebee.pot
 * Changes to README.md changelog.txt etc 

@marmistrz and ( @NikoKrause ) This covers my plans put forward in #450 and the discussion of the future of bumblebee applets. Any comments/suggestions before merging?


NikoKrause:
@pdcurtis
I know you have the README.md file in UUID/files/UUID for offline reasons.

So you have the README.md file twice. It's double the work to have those both changed. I suggest, that you create a symbolic link called README.md, which has the README.md in UUID/files/UUID as target.
Open terminal and `cd` to the UUID and then `ln -s files/UUID/README.md` should create the symbolic link. In theory this should work.

With the new changes of the spices website, there is a new feature. The website now shows CHANGELOG.md files as well like the README.md files. See https://cinnamon-spices.linuxmint.com/applets/view/282

I suggest to rename changelog.txt to CHANGELOG.md and add a symbolic link as well like with the README.md file.

After this has been merged, the plan was to remove the (abandoned) Bumblebee applet, right?
If yes could you create a PR, which deletes it?

pdcurtis:
I have just removed a horrible mistake where several metadata.json files had been uploaded with dangerous flags added. This resulted from already having symbolic links on my local machine for testing which, of course, work both ways and my final commit picked several up without my noticing. i have forced up a correction.

@NikoKrause I wondered about using symbolic links for README.md  but I was not sure how they would work on Github and I note you only say "in theory this should work". I think an experiment on something unimportant might be sensible before I change them all. It only takes a few seconds to do an extra save after the changes have been made.

I will now start to change my changelog.txt files to CHANGELOG.md and copy so they are visible on web site. They may need to be simplified on the view on the website as they go back a long way and may not be consistent with the way others display them. Maybe I will wait to see what you are doing for your own applets before making mine visible on the web site! 

> After this has been merged, the plan was to remove the (abandoned) Bumblebee applet, right?

I was giving time for @marmistrz to respond but I believe that is now the right thing to do.

NikoKrause:
I'm not totally sure, if the symbolic link would work right. If you want to take the risk and it doesn't work, I would fix it for you.

There is no standard for changelog. And as far as I know, there won't be any. You can write Ingo your changelog what you want to. You can even drop it if you want to. Just wanted to let you know about it, since I saw you use changelogs.

pdcurtis:
@NikoKrause I have added a CHANGELOG.md as you suggested as 3.2.0.1 and the symbolic link as 3.2.0.2 so they can easily be reverted if required.

It looks as if Git handles the symbolic links as a test of switching branches and returning left it in place but it looks odd if you do a diff eg `git diff HEAD~1` shows the folder address with no indication it is a link.

If you are prepared to try this out and see if the web site picks up the link correctly please take off the [WIP] and merge this PR. If it does not work the main part should still be OK and you should not even need to revert the extra two commits.

I have limited Internet at present so I will not be able to respond quickly to messages in the next couple of days.

NikoKrause:
@pdcurtis 
Merged! It works. The changelog is shown on your spices website correctly.

PR #709, opened 2017-05-19
SW++@mohammad-sn: Prevent overriding of Cinnamon styles if not explicitly set in the applet's configuration

jaszhix:
@mohammad-sn

This fixes the following issues when the applet is added to the panel (default settings):

- Icon is now visible and the applet is clickable. Defaults to the ```user-desktop```.
- Preserved the applet width from the theme unless explicitly set.
- Preserved the hover bg color from the theme unless explicitly set.

Misc:

- Removed duplicate ```on_applet_removed_from_panel``` method and added its contents to the first one.
- Removed white space.


mohammad-sn:
:+1:
Thank you!

jaszhix:
You're welcome.

PR #710, opened 2017-05-19
collapsible-systray@feuerfuchs.eu: Mark as compatible with Cinnamon 3.4

Feuerfuchs:
Continuation of #680 to investigate why there are issues for some people on Cinnamon 3.4.

### DIFF
--- a/collapsible-systray@feuerfuchs.eu/files/collapsible-systray@feuerfuchs.eu/metadata.json
+++ b/collapsible-systray@feuerfuchs.eu/files/collapsible-systray@feuerfuchs.eu/metadata.json
@@ -7,5 +7,5 @@
     "last-edited":      "1494879445",
     "icon":             "indicator-applet",
     "name":             "Collapsible Systray",
-    "cinnamon-version": [ "2.8", "3.0", "3.2" ]
+    "cinnamon-version": [ "2.8", "3.0", "3.2", "3.4" ]
 }


jaszhix:
This is what I see:
screenshot from 2017-05-19 03-15-14
screenshot from 2017-05-19 03-14-52

I think I know what's wrong - I didn't realize you had to select which apps are collapsible. Though unrelated, toggling apps from the menu produces St errors:

```sh
(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x55560570af90 StBoxLayout.applet-box ff-collapsible-systray__status-icon] which is not in the stage.

(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x5555ff0ad3a0 StBin:last-child first-child] which is not in the stage.
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed

(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x55560570af90 StBoxLayout.applet-box ff-collapsible-systray__status-icon:last-child] which is not in the stage.
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed

(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x55560570af90 StBoxLayout.applet-box ff-collapsible-systray__status-icon:last-child] which is not in the stage.

(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x55560570af90 StBoxLayout.applet-box ff-collapsible-systray__status-icon:last-child] which is not in the stage.

(cinnamon:29749): St-CRITICAL **: st_widget_get_theme_node called on the widget [0x5555ff0ad3a0 StBin:last-child first-child] which is not in the stage.
Window manager warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed
```

Feuerfuchs:
Ah, I'm glad there's no problem after all. :) But it shows that I must make it more clear of the tray is actually collapsible. The current behaviour is clearly bad usability.

About those warnings, I think my method to move an icon from the always-visible container to the maybe-visible container triggers Cinnamon to try to re-insert the tray icon. Working with tray icons is really annoying, you need so many hacks to get things working...

jaszhix:
Yeah, me too! Its a nice applet and I think you did a good job. I used to see those errors in one of my applets, and my solution was to swap out its context menu with one found in the default applets. Not a learning experience exactly, but these JS bindings are sometimes hard to deal with. GObject APIs aren't always helpful when the bindings return C errors.

PR #721, opened 2017-05-19
github-projects@morgan-design.com: improve translations

NikoKrause:
* Repositories name doesn't need to be translated

* make title translatable

* use format instead of append. Not all languages do have the users name at the end. format replaces %s with the users name. This way the position of the users name is up to the translator

@jamesmorgan please check


NikoKrause:
@jamesmorgan any objections?

jamesmorgan:
👍

Issue #738, opened 2017-05-21
pa-equalizer@jschug.com: Glitches when toggling equalizer through context menu switch


0xf4:
[Have you read the guidelines for contributing?]: <> (Before opening an issue, please read the guidelines for contributing linked above. Else the issue will be closed. Also don't forget to notify the author.)

[Notify author of applet]: <> (You should mention the author on GitHub by typing @ plus his/her username to trigger a notification and bring his/her attention to an issue or pull request. You find the username of the author on the cinnamon spices website: https://cinnamon-spices.linuxmint.com/)
pa-equalizer@jschug.com

[Issue]: <> (Describe your issue below.)

After correcting this extension as per #208, pressing its context menu's switch to activate / deactivate the equalizer sometimes results in it ending in the wrong state, after a few glitches. Nonetheless, the equalizer is activated or deactivated correctly (well, unless the control was already in an incorrect state).
I tracked it down to a race condition caused by some asynchronous operations. I'll submit a pull request briefly in which I force those operations to be synchronous.

As a side note, after fixing this issue and #208, there's one last problem remaining: if we launch `pulseaudio-equalizer`'s graphical interface and activate / deactivate its functionality from there, the applet shows the switch in the opposite state... but I think it's the GUI's fault, not the applet's, since the former, apparently, writes the wrong state in the `equalizerrc` file and there's nothing we could do. So the issue should be corrected in that other project, in my opinion (but I'd like to have someone else checking my hypothesis).

PR #739, opened 2017-05-21
pa-equalizer@jschug.com: Fix #208, #738 and some other minor problems

### LINKED ISSUE #208: pa-equalizer@jschug.com: Not working b/c wrong access to `pulseaudio-equalizer` config files
[Have you read the guidelines for contributing?]: <> (Before opening an issue, please read the guidelines for contributing linked above. Else the issue will be closed. Also don't forget to notify the author.)
@clefebvre @NikoKrause
(the author is "jschug.com", but it seems a corresponding github account doesn't exist; I'm sorry if I notified the wrong people)

[Issue]: <> (Describe your issue below.)
**TL;DR**: Simply put, the applet doesn't work at all in its current form (Mint 18.1 Mate ed. here). I have a solution (see below) but I'm concerned about its (non-)portability, so no PR provided (but a diff).

--

The problem is already described on the [applet's homepage](https://cinnamon-spices.linuxmint.com/applets/view/176), on a comment by user "copecu" (2016-10-21 16:23) which I copy here. Emphasis is mine:
> Hi,
>
> I am on Linux Mint Cinnamon 18 and it seems that this nice applet doesn't work. I can add them to the panel but **toggle equalizer doesn't work and no presets are showed**.

I'm on LinuxMint 18.1 Serena Mate Edition and can confirm the problem.

To reproduce it, just install `pulseaudio-equalizer` and this applet, and check that while the former works as expected the latter exhibits the aforementioned behavior.

I **found the root of the issue and have a working solution for my system**, but I'm unsure if it'd be portable to other distributions (or even to older Mint versions) as-is, so I won't do a pull request (but provide the diff below).

The cause is two-fold:

1. **New versions of `pulseaudio-equalizer` move `pulseaudio`'s config directory** from `~/.pulse` to `~/.config/pulse`. Pulseaudio itself doesn't seem to care if its config is put in either of them; when both are present, it prefers the latter.

2. The calls to `monitor()` and `replace()` on Gio File objects pass **wrong `flags` parameters**: they both use `null` instead of the corresponding constant.

Here's the diff:
`cinnamon-spices-applets/pa-equalizer@jschug.com/files/pa-equalizer@jschug.com/applet.js`
```diff
@@ -9,7 +9,7 @@
 const Gettext = imports.gettext.domain('cinnamon-applets');
 const _ = Gettext.gettext;
 
-const CONFIG_DIR = GLib.get_home_dir() + "/.pulse";
+const CONFIG_DIR = GLib.get_home_dir() + "/.config/pulse";
 const EQCONFIG = CONFIG_DIR + "/equalizerrc";
 const EQPRESETS = EQCONFIG + ".availablepresets";
 const PRESETDIR1 = CONFIG_DIR + "/presets/";
@@ -26,7 +26,7 @@
     load: function() {
         try {
             GLib.spawn_command_line_sync("pulseaudio-equalizer interface.getsettings");
-            this._monitor = Gio.file_new_for_path(EQCONFIG).monitor(null, null);
+            this._monitor = Gio.file_new_for_path(EQCONFIG).monitor(Gio.FileMonitorFlags.NONE, null);
             this._monitor.connect("changed", Lang.bind(this, function(self, file, otherFile, eventType) {
                 if (eventType == Gio.FileMonitorEvent.CHANGES_DONE_HINT) {
                     this._configChanged();
@@ -42,7 +42,7 @@
     },
     save: function() {
         try {
-            let out = Gio.file_new_for_path(EQCONFIG).replace(null, false, null, null);
+            let out = Gio.file_new_for_path(EQCONFIG).replace(null, false, Gio.FileCreateFlags.NONE, null);
             out.write_all(this._rawdata.join('\n'), null);
             out.close(null);
```
maybe the first of the changes could be replaced by a recommendation in the applet's (non-existent) `README` file explaining how to symlink `~/.pulse` to `~/.config/pulse`

0xf4:


0xf4:
Sorry, I somehow combined what was supposed to be two different, separate pull requests. Trying to solve it...

PR #742, opened 2017-05-21
Enable applet for intel_pstate (no frequency change) only governor ca…

kroutal:
Enable applet for intel_pstate (no frequency change  only governor can be change)

Add a fix to enable applet to be on a bottom panel


brownsr:
ping @mtwebster

jaszhix:
This is donated. It has some other issues though (using PanelMenu) which I can fix in another PR.

PR #782, opened 2017-05-29
Argos for Cinnamon (0dyseus@ArgosForCinnamon) initial release

Odyseus:
## Description

Argos for Cinnamon lets you write Cinnamon applets in a language that every Linux user is already intimately familiar with: Bash scripts. Not only that, if you are familiar or more comfortable with any other language (like Python, Ruby, JavaScript or even Perl or PHP), Argos for Cinnamon can handle them as well.

More precisely, Argos for Cinnamon is an applet that turns executables' standard output into panel dropdown menus. It is inspired by, and fully compatible with, the Gnome Shell extension called [Argos](https://github.com/p-e-w/argos) by [Philipp Emanuel Weidmann](https://github.com/p-e-w), which in turn is inspired by, and fully compatible with, the [BitBar](https://github.com/matryer/bitbar) application for macOS. Argos for Cinnamon supports many [BitBar plugins](https://github.com/matryer/bitbar-plugins) without modifications, giving you access to a large library of well-tested scripts in addition to being able to write your own.


jaszhix:
This is awesome. I was considering porting this over at one point - I'm really glad you did this!

NikoKrause:
@ Odyseus. It has no icon and no screenshot

Odyseus:
Hello, @NikoKrause.

I just saw that, thanks. Working on it. 👍

Odyseus:
Hello, everybody.

@jaszhix: Sorry, I didn't see your comment the last time that I commented here.

The original Argos extension is an amazing and very powerful extension and using it in Cinnamon makes it even more powerful.

I hope that I did as good of a job porting it as any real developer (like you) would have done so. :+1:

PR #886, opened 2017-06-11
cheaty@centurix: Stop using the _menuManager in IconApplet as it doesn't play well with the configuration context menu

Centurix:
IconApplet's _menuManager property doesn't manage menus correctly when the applet has a right click context menu for configuration options. Left click the applet and then immediately right click can place Cinnamon in a weird state sometimes.


collinss:
Could you please remove the dangerous property from metadata.json? it is supposed to be generated by Cinnamon settings and shouldn't be there until then.

PR #910, opened 2017-06-17
[hideable-applets@cardsurf] Add a new applet

cardsurf:
An applet that hides applets for Linux Mint Cinnamon

screenshot

[Original repository](https://github.com/cardsurf/hideable-applets).


jaszhix:
I saw some warnings:

```sh
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/hideable-applets@cardsurf/files.js 130]: Too many arguments to method Gio.File.make_directory: expected 1, got 2
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/hideable-applets@cardsurf/files.js 43]: Too many arguments to function GLib.file_set_contents: expected 2, got 4
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/hideable-applets@cardsurf/applet.js 706]: assignment to undeclared variable indexes
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/hideable-applets@cardsurf/files.js 43]: Too many arguments to function GLib.file_set_contents: expected 2, got 4
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/hideable-applets@cardsurf/files.js 43]: Too many arguments to function GLib.file_set_contents: expected 2, got 4
```

Other than that everything seemed to work well in my tests on Cinnamon 3.4 and 3.2. This is definitely useful for freeing up panel clutter.

NikoKrause:
@cardsurf
This sounds like a cool applet. Though I must say, I thought it was more like the [Windows 7 like arrow](https://www.sevenforums.com/attachments/tutorials/328687d1407639595-notification-area-icons-hide-show-system-icons-notifications-notfication_area_icons.png). But it isn't. After trying it, I must say, I would wish it would be more like the Windows 7 arrow. I.e. concrete things that I'd like to see improved (in my opinion):
1. When I press the applet, a popup menu appears, with all applets listed in it. If I press those icons in the popup menu, I expect, it opens the applet, but all it does is toggling visibility of those applets. (But I thinks that's not the intention and would be hard to implement.) Just irritated for the first few minutes, because I expected something else. So forget about this point ^^
2. If I hide applets and restart Cinnamon, the hidden applets are no more hidden. I.e. your applet doesn't remember, which applets are hidden.
3. It's not visible in the popupmenu, which applets are hidden and which ones are not. That's an issue imo. You should highlight, if an applet is hidden and unhighlight, if it's visible. (Maybe using pseudo style class "hover" or "focus" or something like that)
4. You use other/bigger icons in the popup, than those on the panel.  Some applets have the same icon or just a "? question mark" as icon. Therefore it's irritating at first to see, which icons belongs to which applet. It would be better, if you would use the same (symbolic) icons as in the panel. (I think it's hard to get them?). I would at least change the following:
* Use smaller icon size by default (size 22 or what is the default? Since it's possible in the panel settings to "allow cinnamon to change the icon size depending on the panel size", maybe it's also possible to get the icon size of the icons in the panel. Then I would add an option, "Use same size for icons as the panel does" by default and add an option to set custom icon size.)
* The systray@cinnamon.org uses a question-mark as icon. But it definitely has a icon.
5. The icon of your applet is white. It should use the same color as the other symbolic icons in the panel.

cardsurf:
@NikoKrause 
Thanks for the feedback and ideas. Regarding the mentioned remarks:
1. The applet toggles visitbilities of applets in the popup instead of opening them or simulating a left click.
2. The applet saves information about last used values to the CSV file located in: `~/.local/share/cinnamon/applets/hideable-applets@cardsurf/values/last_values.csv`. If a user hide/unhide an applet and immediately restart Cinnamon the changes will not be retained. A saving interval can be configured in the settings under the `Save every` option. Is there a better way to save many values using Cinnamon built-in settings without high CPU usage ? 
3. Sounds like a nice improvement.
4. A default icon size is 40 pixels. This can be customized under the `Icon size` option in the configuration. 
   There could also be a button that sets the icon size to match a panel size.
   Right now the applet looks for applet icons in the: `/.local/share/cinnamon/applets/` directory or in an applet metadata. If an icon is not found it displays the question mark icon. The applet could also check if an `icon.png` file exists in: `/usr/share/cinnamon/applets` to show more applet icons.
5. Is there any good small arrow up symbolic icon that could be used ?

NikoKrause:
2. Not sure. I think you better ask Cobinja or Odyseus for that.
4. You find icons of (official) spices in `~/.cinnamon/spices.cache/applet`. Then you can check `~/.local/share/cinnamon/applets/` for applets, which are not in spices, and `/usr/share/cinnamon/applets` for applets, which are preinstalled with Cinnamon.
5. You can copy `/usr/share/icons/Adwaita/scalable/actions/pan-up-symbolic.svg`

Odyseus:
Hello, everybody.

@cardsurf

> Is there a better way to save many values using Cinnamon built-in settings without high CPU usage?

You could simply use a **generic** type of setting ([See the settings reference](http://developer.linuxmint.com/reference/git/cinnamon-tutorials/xlet-settings-ref.html) and search for "generic"). You can save any type of data in those. By doing so, you can get rid of all the code in your applet that handles the **last_values.csv** file. Also, by doing so, your applet will remember those values every time it is updated by the user.

**About the applet icon:** I think that you can safely use the **pan-up** icon as suggested by @NikoKrause without actually copy it into your applet. This icon (along with the **pan-down**, **pan-end** and **pan-start**) *seems* to exists only on the **Adwaita** and **Hicolor** icon themes and these two icon themes *seems* to exists in all distributions (not the emphasis on **seems**, I cannot be entirely sure).

**About the icon for applets inside the menu:** You could also look for an applet icon on the applet's metadata. You could just check if the **icon** key exists and also you could check if that icon exists in the current icon theme.

cardsurf:
@Odyseus
I will check these solutions. Thanks.

PR #911, opened 2017-06-17
places center: fix misplaced variables

### LINKED ISSUE #249: placesCenter@scollins: Panel Text Doesn't Show Anymore
@collinss I think some of the late changes affects the way the Places Center applet behaves.

I don't see the panel text anymore. I have it in an horizontal panel which should display the panel text.

I've only updated the applet yesterday so I don't know since when this problem arose but it did since I updated.

mohammad-sn:
fixes #249


mohammad-sn:
@collinss

PR #920, opened 2017-06-21
xampp-panel@backids99: changed xdg-open instead of gedit

volkovmqx:


Piiit:
Why not using xdg-open to open with user's preferred text editor?

volkovmqx:
Cool idea ! Thanks @Piiit

PR #932, opened 2017-06-23
feeds@jonbrettdev.wordpress.com - Update to config file location to resolve built in config issue.

jake1164:
This will resolve the issue with right click and pressing config not working. If you have existing feed config file in ./cinnamon/configs/feeds@jonbrettdev.wordpress.com/feeds.json you will need to manually move it to where data is stored .cinnamon/feeds@jonbrettdev.wordpress.com/feeds.json


NikoKrause:
If you want the Changelog to be shown in the spices Website, you can rename it to CHANGELOG.md and it will appear there.

PR #957, opened 2017-06-29
Cinnamenu@json: Update to 2.0.0

jaszhix:
screenshot

Changes:

  * Reorganized the layout for more efficient use of space.
    * Search box has been moved to the bottom right corner.
    * Selected app box has been moved to bottom left column.
  * Added settings from the default menu that were missing.
    * Translations for these were added from CinnVIIStarkMenu's PO files.
  * Added an option to display the description on each app button.
  * Added a user account button showing the current user's avatar.
  * Added key navigation.
  * Added a "Clear List" button for recent files.
  * Renamed the Recent category to Recent Files.
  * Descriptions now scroll horizontally if they are truncated.
  * Faster fuzzy searching.
    * Searches now highlight the matching characters in item names and descriptions.
  * Added filesystem path entry autocompletion based on the Cinnamon menu's code.
  * Added highlighting for newly installed apps.
  * Formatted the settings using the layout API.
  * Removed a lot of stale/dead code from the Gno-menu project, and refactored a lot of the button handling.
  * Removed automatic icon scaling. 
  *  Fixed Opera bookmarks.
  * Fixed the missing gda library error message.

Remaining tasks before merging:

 *  ~~Add keyboard driven navigation~~
   * ~~Add key navigation for context menus~~
 *  ~~Determine replacement of the grid/list toggle button~~
 *  Thorough regression testing

I rewrote a lot of ugly code, and removed some left over stale code from the forked project. I am seeing better performance all around on my end. Thanks to @NikoKrause, @Faust30000, @giwhub, @muzena, @eson57, and @Alan01 for the translation help!


brownsr:
Nice.  Very fast and snappy in use, and looks better, to me anyway.  Not sure about the bookmarks at the top, but I guess that's personal taste. Can this show Opera bookmarks?  I get the Firefox bookmarks with Gir1.2-gda-5.0.
Still munches up memory though, I am on about +200MB over typical Cinnamon memory usage right now

jaszhix:
Thanks! The bookmarks for me still cause unresponsiveness for a couple seconds when hovering over them (I have 300+), so I moved them to the top so people with tons of them like me don't keep accidentally triggering the loading of them. If I can figure out an optimization for adding hundreds of actors to a parent, I can move that back down. 

I'm currently working on code that checks if an actor is visible using position, width, and height on ```scroll-event```, so it only adds actors to the parent if they're visible (or perhaps hides them if they're not). ```Clutter.Actor.is_visible()``` is unfortunately assuming anything on the stage is visible. This is a trick I use in web app projects for rendering huge lists, but still not sure I can get it working in Clutter.

I haven't tested Opera on this install - I'll see if that's working.

brownsr:
The little toggle switch that allows switching between list and grid has disappeared.  I hope it is coming back !

jaszhix:
Oops, I forgot to add that to the changelog. I was going to remove it because I found myself never using it. I suppose I can add it back in after I'm done with key navigation.

brownsr:
weird regression - it's maintaining hover highlighting when moving the cursor off an application.  Most noticeable in all applications and in preferences, can't make it happen in all categories

jaszhix:
Ah, good catch - I was able to reproduce the hover issue. The recent/bookmarks issue is the same issue I mentioned earlier, where long lists are blocking I/O. I could limit the recent items to 20 like the default menu - I thought the JS API handled it for some reason. I'd like to figure out a real solution for it though.

brownsr:
Hover issue still there

jaszhix:
I'm not reproducing it reliably. What are you clicking?

brownsr:
launch menu.  click on all applications.  mouse over the applets on the grid with a swirling motion.

jaszhix:
Hm, weird. I reproduced it right after you mentioned it, but haven't since I started on my latest commits. I'm testing different themes too, default LM theme, Minty Y, Photocopy, etc.

brownsr:
Interesting, it's only really apparent in the theme I was using, Photocopy.  There's some flickering redisplay artefacts visible in Glass-Lemon (for example) on the scroll bar in particular.  Other themes show no issues.

brownsr:
There's an interesting cyclical visual effect associated with the application description that might be related.  Using the theme photocopy I hover over an application with a long description, in the 'all applications category', the description starts to scroll in the description box to the bottom left.  At the same time the effects on the hovered item cycle in sync with the scrolling description.  Also the intensity of the titles on each application grows and drops in sync with the scrolling description.  There's also a minor icon effect, less noticeable. Any multiple hover effect visible clears at the time of the first cycle of the description.

jaszhix:
Oh yeah, I see it too. The way I'm scrolling text is simple string manipulation until I can figure out the maybe possible way to do it in Clutter. I'm guessing its a side effect with pango ellipsis.

brownsr:
the cyclical visual effect comes from this bit of code. I am not sure what you intend it to do as it looks the same if removed to me:

```javascript
    if (clutterText) {
      this.label.get_clutter_text().set_markup(markup);
      clutterText.ellipsize = opts.reset && opts.reset === 1 ? null : Pango.EllipsizeMode.END;
    }
```

jaszhix:
Wow, yeah, that's for app descriptions on buttons, which is disabled by default.

jaszhix:
@brownsr A lot of changes since your last comment. I think I *might* have fixed the bug causing what happened in that crazy screenshot, but I never reproduced it to that degree with Photocopy.

brownsr:
Nice !

PR #1009, opened 2017-07-07
Spices Repo Itself: add CODEOWNERS file

NikoKrause:
GitHub has a new feature: https://github.com/blog/2392-introducing-code-owners

Not sure if it will work


Odyseus:
Hello, @NikoKrause.

I think that it would be better to use a script that would generate the **CODEOWNERS** file automatically. That way, it would be easier to maintain and keep up-to-date.

I made such a script yesterday when I found out about this new GitHub feature (in Python, I completely suck at BASH).

[Give it a try if you want](https://gist.github.com/Odyseus/5010cfbbc517fb5b5b6297594bd59e47). It has to be executed from inside the **.github** folder and it will overwrite an existent **CODEOWNERS** file. it also ignores xlets whose authors are set to **none** (It's just pointless to add a reference to a nonexistent GitHub user).

NikoKrause:
I commented the @none lines out.

I don't think we need a script for that.
This file should be changed as often as the info.json files, i.e.

* If an applet is deleted
* If an author changes
* If someone uploads a new applet 

this file should be edited. And that doesn't happen that often.

collinss:
@NikoKrause If we edit it manually, there is more likely to be mistakes. I know, for example, that I'm going to forget once in a while. And the author shouldn't be expected to edit the file themselves. In fact, I'm not sure we want them to touch anything outside the folder of their own applet. If we automate it like @Odyseus suggests, we can just run it periodically when we think of it, and it should catch anything we might have forgotten.

NikoKrause:
Well idc.

NikoKrause:
@Odyseus 
Your python script works here. Do you want to open a PR?

Odyseus:
Hello, everybody.

@NikoKrause: I can do that. But that script was just a practical example (it doesn't have *safeguards* and it's slow as heck (LOL)). I created an *improved* version ([can be seen here](https://gist.github.com/Odyseus/5010cfbbc517fb5b5b6297594bd59e47) and [this is how the resultant file would look like](https://gist.github.com/Odyseus/5010cfbbc517fb5b5b6297594bd59e47#gistcomment-2145023)).

Before publishing it, I would like some input from you all.

- At the beginning of the generated file is a comment that specifies that the file shouldn't be edited manually and how to generate it.
- The xlets without author are grouped at the end commented out. So it's easier to see them at a glance. The comment that separates these xlets reads *"The following xlets have no author assigned to them."*. Maybe add some more details to this comment?
- I used the word *xlets* so the script can be used in the 4 Spices repositories without the need to edit it.

NikoKrause:
Looks good to me

PR #1010, opened 2017-07-08
Added new applet: redshift@marvel4u

Marvel4U:
New applet to conveniently activate and configure
redshift one shots. Forked from brightness@markbokil.com.


brownsr:
Looks fine, but could you please remove the 'dangerous' flag, as that is not relevant (it's part of the loading process)

PR #1013, opened 2017-07-09
Argos for Cinnamon (0dyseus@ArgosForCinnamon) update

Odyseus:
- Commented out the unused TryExec function to prevent a warning.
- Initialization of the PopupSubMenuMenuItem prototype with an empty label to prevent an error that will be caused when a change in Cinnamon gets merged. ¬¬
- Removed font size style from the README header. This style was causing rendering problems on the Spices website.
- Updated help file.


jaszhix:
Hm, I'll update it to use stricter equality in that PR.

PR #1016, opened 2017-07-10
Added new applet: redshift@marvel4u

Marvel4U:
New applet to conveniently activate and configure
redshift one shots. Forked from brightness@markbokil.com.
"dangerous" flag removed.


Marvel4U:
As requested I removed the dangerous flag compared to my last pull request. (Don't know if there is an easier way than deleting it and creating a new request.) Before I left it in because it was in the applet I forked from and I couldn't find guidelines to when it should be used.
Cheers,
Marvin.

jaszhix:
You just need to push the commits to your branch from git, and they will be added to your PR.

PR #1063, opened 2017-07-22
IcingTaskManager@json: Updated bulgarian translation

AdmiralAnimE:
Fixed bad translations. Not sure if the file is up to date, default.pot is very different.


jaszhix:
Hi, thanks for helping. I didn't have time to update the other PO files, but the default.pot file is the one you will want to use.

jaszhix:
I copied over your updated translations into the now up-to-date bg.po file. Feel free to submit another PR if I messed something up.

PR #1077, opened 2017-07-25
redshift@marvel4u: fix translation pot and po file

NikoKrause:


NikoKrause:
@Marvel4U for your info

Marvel4U:
Thanks. Actually I just realized that in my German system the translation doesn't work. Only "Brightness" is translated. Don't know what I missed for successful translation...

PR #1078, opened 2017-07-25
CODEOWNERS: try ** instead of * so that it has infinite depth

NikoKrause:
https://git-scm.com/docs/gitignore#_pattern_format


Odyseus:
Hello, @NikoKrause.

It seems that there is no indication that the CODEOWNERS file is actually working in this repository (?). Have you seen any change after you added the *"double asterisks"* to it? I followed your tests on your own repository and also didn't see any signs of it working.

On the other hand, I scanned GitHub for repositories (I looked at hundreds of them) using the CODEOWNERS file and in all of them it seems to be working as expected.

Maybe the *default owners* needs to be specified at the beginning of the file? Maybe the repositories that belong to an organization works differently? Or maybe this new feature hasn't propagated to all of GitHub yet? But I think that before continuing to speculate and doing more tests blindly, it might be time to contact GitHub support to shed some light in the matter.

NikoKrause:
Hey @Odyseus ,

I've already contacted GitHub support a few times. The results of the contact was:
1. The tests on my own repo didn't worked, because "the automatic will only happen if the author of the pull request is not the code owner". "This happens because you can not be a reviewer on a pull request that you authored."
2. Another reason why it didn't worked was, "that CODEOWNERS file [wa]sn't working for directories with the @-symbol in its name". "This bug with the `@` in the directory name [...] has now been fixed."
3. After I got the mail, that the bug has been fixed, I observed, that it still doesn't work. So I contacted GitHub support again. I got an answer from Laura (@lecoursen) to "try using `**` instead of `*` here, so that it has infinite depth". But that didn't worked. Now I'm waiting for an answer.

I wonder, which requirements the CODEOWNER feature has.
A thought experiment:
You create a CODEOWNER file in one of your repos and add some user, who has nothing to do with the repo. He gets everytime a notification to review a PR, and has no chance to turn those off.

So I'm not sure how it works properly. Hoping to get an answer soon.

PR #1101, opened 2017-07-30
redshfit@marvel4u: Applet settings and optionally night value changes

Martin1887:
The following changes have been done

- Add settings to remember last user decission (color temperature, brightness, autostart Redshift if it was started before computer shutdown...).
- Add options for night values (change on night can be disabled or enabled with a switch in a similar manner as enabling or disabling Redshfit).
- Applet icon changes for another with smaller rays when Redshift is disabled.
- Applet icon svg included in the applet for consistency (the previous icon could change with user icon theme).
- General icon changed for Redshift icon.

To Do:
- Translations for new labels have not been added.


Martin1887:
Applet changed is redshfit@marvel4u.

NikoKrause:
@Marvel4U Are you okay with this pull request?

NikoKrause:
Can you remove the `dangerous` and the `last-edited` lines in the metadata.json file?

If you want to update the `redshift.pot` file, you can run the following command in the applets directory:
```
cinnamon-json-makepot --js po/redshift.pot 
```

Martin1887:
Of course, I can. But should I create a new pull request or otherwise how can I push the changes in this pull request?

Thanks and regards.

NikoKrause:
Just add a new commit to your branch, then it will show up here in the pull request.

Martin1887:
Ok. Commit done (sorry for the first wrong commit message without applet name).

NikoKrause:
I don't see any difference in es.po and es_ES.po.
One file es.po should suffice, also `Language` in the es.po file should be changed to `es`

Correct me, if I'm wrong.

Marvel4U:
Heyo, 
thanks for adding the other redshift features! Looks nice. 
But there are some small things I'd propose to change before merging. If you get around to look into them, feel free, else I'll give it a shot one of these days.
- The order in the applet menu is now a bit confusing. Maybe there should be a redshift-one shot section and then a redshift-automatic change section. Right now it's a bit mixed up and I think not clear what the two buttons do.
- The two buttons don't interact well. For example if you have both on and turn off the night change, the one-shot is not activated. I guess one should be set to off if the other is activated
- when changing the one-shot settings on the sliders there is now a flickering that I think wasn't there before
- the brightness can be displayed as 57.9999999999% which should be rounded I guess
- nice to have the changing icons! But since my original intention was to imitate a sun-set maybe we could have a day-sun and a sunset-sun? Or a sun and a moon? Opinions?
That's it for now. Do you agree?
Cheers,
Marvin.

NikoKrause:
The Material Design Icon has a lot of icons to choose from: https://materialdesignicons.com/

You find the license of those icon here in the README: https://github.com/Templarian/MaterialDesign

Martin1887:
I agree with your proposals.

Only a notice: @Marvel4U says that only one checkbox (change on night or Redshfit) should be enabled at time. My idea is different: Redshfit checkbox starts or stops Redshift at all, if Change on night is enabled then Redshift is started with transition between day and night, otherwise Redshift is started in one-shot mode.

It's true that now when you disable Change on night mode, Redshift does not start in oneshot mode. This is a bug and I'm going to work in fixing it.

When changes have been finished I will comment again!

Thanks and regards.

Martin1887:
Hello!

I have finished all changes. Below are the changes:

>     The order in the applet menu is now a bit confusing. Maybe there should be a redshift-one shot section and then a redshift-automatic change section. Right now it's a bit mixed up and I think not clear what the two buttons do.
Configuration menu is divided in three sections now. Also, Redshift switch has been pushed just before day/oneshot settings. Hence, in the applet each switch is the separator of each section: general brightness on top, enable/disable Redshift marks Redshift options and Transition on night switch marks night options. It could be a bit confusing that day brightness and colour values are also used in oneshot mode, but I think that adding two more sliders would be weird and the applet too much big. So, oneshot mode is like saying that it is always daylight.

>     The two buttons don't interact well. For example if you have both on and turn off the night change, the one-shot is not activated. I guess one should be set to off if the other is activated when changing the one-shot settings on the sliders there is now a flickering that I think wasn't there before
Bug when changing applet modes have been fixed. Now, one or two seconds happens between changing modes but this is the time lapsed on killing old Redshift transition mode. Flickering has also been fixed launching redshift -x command only on disabled mode. However, in transition mode a bit of flickering happends because Redshift reset temperature and bright to the maximum and then gradually changes color to the corresponding time.

>     the brightness can be displayed as 57.9999999999% which should be rounded
I have not seen never brightness with decimals, and in the code I see rounds in all functions. Could you please fix it if you detect it?

>    I guess nice to have the changing icons! But since my original intention was to imitate a sun-set maybe we could have a day-sun and a sunset-sun? Or a sun and a moon? Opinions?

I think you will love the new icons! Now a black light bulb is used for disabled Redshift, a red light bulb is used for oneshot mode and a red sunset is used for transition mode. Icons have been modified from Material icons.

>     also I'd update the Readme to include the new features
Readme has been updated


I hope you like all the changes!

Below, the TO-DO tasks:
- I have updated pot file and Spanish po file, but german and chiness translations should be done.
- In spite of regenerate pot file and translating to Spanish, I only can see translated the 'brightness' string (that come from the forked applet for you). I have never worked with pot, so I don't know if something more has to be done for translations appear.

And lastly a notice: as settings schema has been changed, you have to delete your ~/.cinnamon/configs/redshift@marvel4u folder to reload the new schema and see the changes. I suppose that Cinnamon will do it when the applet be installed in users computer using Cinnamon applet settings.


Thanks and regards.

Marvel4U:
Hello,
looks really nice, thanks for fixing the issues!
The changed order makes it much more understandable. That the day-sliders have a kind of double role is indeed a bit strange but I think people will figure it out.
The flickering is gone and at least for now I don't get the decimals any more.
And thanks for the new icons, I think they are very clear and mark the differences in state nicely.

About the Language: the translations also don't work for me but I don't know what the issue is here. Running the makepot didn't change anything.
And the dangerous flag is there again. Could it be that it's added automatically somehow?

Anyway these are somewhat minor issues... and since this pull request is marked as such: Approval by author given ;)
Cheers.

NikoKrause:
About translations: Here's a tutorial I wrote some time ago. It explains how to update existing translations and how to test them, when changed. https://github.com/NikoKrause/CinnVIIStarkMenu/issues/19

I'd have a question and a suggestion for the applet:
(1) The Brightness slider at the very top doesn't seem to do anything for me.
(2) It would be nice to see the current/detected location and the current color temperature.

Martin1887:
Hello @NikoKrause.

The brighness slider does also not work for me, but why the Cinnamon-settings-daemon crashes on startup in my system. If you see an error similar to this in ~ /.cinnamon/glass.log, probably this is the cause: Error getting Cinnamon backlight power settings true,,Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown:.

About your suggestions, I think that the applet should not be more saturated. However, having **in the tooltip** current location and max and min brightness and color temperature (only current in one-shot mode) would be nice. I think that current color temperature doesn't worth, since it requires executing Redshift in verbose mode and saving constantly the standard output (I think that Redshift-gtk does it this way, and Redshift has not another option for doing it).

In any case, I think that one issue should be opened for translation fixes and another one for your suggestions, since they are a bit out of scope of this pull request and it has already be merged.


Thanks and regards.

NikoKrause:
My personal view (please feel free to ignore, I don't want to tell you how to design your applet, the following is my opinion only):
Right now all the sliders and switches, which pop up, when you click in the applet are also available in the settings. Once I set up the color temperature, I don't want to change them anymore very often. Also I don't want to change them by accident, which can happen very fast, since the sliders are in the pop up.
I would just show the needed infos about color temperature and location and the switches. And I would show the sliders only in the settings. It's hard to adjust them anyway. They jump with steps of 50, i.e. it's hard to set a temperature of exactly 6500. You have to type it in the settings.

Martin1887:
I think that it is good being be able to change values in applet, for instance if the current day is specially cloudy you could want to change color and brightness values fastly. Another example is people that don't use transition between day and night and want one-shot mode in their way. In addition, maybe not all people know applet settings.

Initially the applet had not settings, they are specially useful for remember user values instead loading defaults in each system startup.

It's true that setting the exact value in the applet is not easy, but I think that many people would play with values until get a confortable set of values, and this is difficult in changing values in settings.

In conclusion, although a bit replicated, settings in applet and applet settings are complementary. I think that it would be very useful having a fast summary in applet tooltip but, as I commented in my previous comment, this is not so easy (specially current Redshift calculated value, the rest of values are feasible with a bit of work).

What do you think about all these questions, @Marvel4U? If you agree maybe I begin this change soon.

@NikoKrause, I appreciate your advises.

Thanks and regards.

NikoKrause:
OK. Valid points. Then just ignore my comments. I don't change values that often.

Marvel4U:
Just as a final comment on this merger...
I super agree with @Martin1887 . Actually the main reason to make this applet was so I can change the temperature quickly on the panel without opening any settings or a terminal to run redshift. I'm exactly the kind of person who uses the one-shot and changes the temperature every half hour depending on how late it already got.

PR #1118, opened 2017-08-05
Tor button applet

Goodevil95:
Add new simple applet which allow to enable/disable Tor.


tor-button@shatur/files/tor-button@shatur/metadata.json
```diff
@@ -0,0 +1,6 @@
+{
+    "dangerous": false,
```
jaszhix: The dangerous prop is set by Cinnamon automatically, and should be removed.

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -0,0 +1,109 @@
+const Applet = imports.ui.applet; //Main
+const St = imports.gi.St; // Needed for translations
+const Gettext = imports.gettext; // Needed for translations
+const GLib = imports.gi.GLib; // Needed for starting programs and translations
+const Settings = imports.ui.settings; // Settings in applet
+const Mainloop = imports.mainloop; // Needed for timer update loop
+const Lang = imports.lang; // Needed for menus
+const UUID = "tor-button@shatur"; // Applet UUID
+
+torAppletCheck = "OFF"; // Show information to applet from '/tmp/.torAppletCheck'
```
jaszhix: These all needed to be declared in the scope with const or let.

```
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 10]: assignment to undeclared variable torAppletCheck
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 11]: assignment to undeclared variable torEnable
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 12]: assignment to undeclared variable applet_running
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 13]: assignment to undeclared variable press_on
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 14]: assignment to undeclared variable press_off
```

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -0,0 +1,109 @@
...
+    
+    _init: function (metadata, orientation, panelHeight, instance_id) {
+        Applet.TextIconApplet.prototype._init.call(this, orientation, panelHeight, instance_id);
+        
+        //Settings
+        this.settings = new Settings.AppletSettings(this, UUID, instance_id); // Settings initialization
+        this.settings.bindProperty(Settings.BindingDirection.IN, // Setting type
+        "location", // The setting key
+        "location", // The property to manage (this.location)
+        this.on_settings_changed, // Callback when value changes
```
jaszhix:
```
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js 41]: reference to undefined property this.on_settings_changed
```

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -0,0 +1,109 @@
...
+    on_applet_removed_from_panel: function () {
+        applet_running = false;
+        this.settings.finalize();
+    },
+    
+    // Update loop timer and check status of Tor
+    updateLoop: function () {
+        if ( applet_running == true) {
+            GLib.spawn_command_line_async('sh ' + GLib.get_home_dir() + '/.local/share/cinnamon/applets/'+UUID+'/check.sh'); // Run check bash script to write Tor status in '/tmp/.torAppletCheck'
+            torAppletCheck = GLib.file_get_contents("/tmp/.torAppletCheck").toString(); // Read information from file
```
jaszhix: Try using [GFile.query_exists](http://devdocs.baznga.org/gio20~2.50p/gio.file#method-query_exists), and handle it if the directory doesn't exist.

```
(cinnamon:6258): Cjs-WARNING **: JS ERROR: GLib.FileError: Failed to open file '/tmp/.torAppletCheck': No such file or directory
MyApplet.prototype.updateLoop@/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js:75:30
```

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -0,0 +1,109 @@
...
+            Mainloop.timeout_add_seconds(this.refreshInterval, Lang.bind(this, this.updateLoop)); // Update loop
+        }
+    },
+    
+    on_applet_clicked: function() {
+        if (torAppletCheck.substr(5,3) == "OFF") { // Check information from file without 'true,' and '\n '
+                                                   GLib.spawn_command_line_async(this.location); // Run tor from location settings        
+        }
+        else {
+            GLib.spawn_command_line_async('pkill tor');  // Kill Tor
```
jaszhix: This fills .xsessions-errors with permission errors when left clicking the applet. I don't see why the applet needs to handle killing tor.

```
pkill: killing pid 404 failed: Operation not permitted
pkill: killing pid 421 failed: Operation not permitted
pkill: killing pid 404 failed: Operation not permitted
pkill: killing pid 421 failed: Operation not permitted
pkill: killing pid 404 failed: Operation not permitted
pkill: killing pid 421 failed: Operation not permitted
```

tor-button@shatur/files/tor-button@shatur/metadata.json
Goodevil95: Thanks!

tor-button@shatur/files/tor-button@shatur/applet.js
Goodevil95: Okay! I'm sorry, I'm newbie in Javascript.

tor-button@shatur/files/tor-button@shatur/applet.js
Goodevil95: Sorry, I forgot remove this function from settings.

tor-button@shatur/files/tor-button@shatur/applet.js
Goodevil95: Yes, I forgot about this :(

tor-button@shatur/files/tor-button@shatur/applet.js
Goodevil95: I'm launching Tor without su privilegies. But I add ability to run and kill Tor with gksu.
Killing Tor is needed when you want to disable Tor. I not found how to stop service and decided that I can just kill service.

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -94,10 +105,10 @@ MyApplet.prototype = {
     
     on_applet_clicked: function() {
         if (torAppletCheck.substr(5,3) == "OFF") { // Check information from file without 'true,' and '\n '
-                                                   GLib.spawn_command_line_async(this.location); // Run tor from location settings        
+                                                   GLib.spawn_command_line_async(gksu+this.location); // Run tor from location settings
```
jaszhix: When I click this, I see:

```
(cinnamon:6258): Cjs-WARNING **: JS ERROR: GLib.SpawnError: Failed to execute child process "tor" (No such file or directory)
MyApplet.prototype.on_applet_clicked@/home/jason/.local/share/cinnamon/applets/tor-button@shatur/applet.js:108:52
Applet.prototype._onButtonPressEvent@/usr/share/cinnamon/js/ui/applet.js:282:21
```
This is because I'm using the latest build from a PPA. The actual command to start Tor for me is ```/usr/bin/tor-browser-en.sh```.

tor-button@shatur/files/tor-button@shatur/applet.js
```diff
@@ -94,10 +105,10 @@ MyApplet.prototype = {
     
     on_applet_clicked: function() {
         if (torAppletCheck.substr(5,3) == "OFF") { // Check information from file without 'true,' and '\n '
-                                                   GLib.spawn_command_line_async(this.location); // Run tor from location settings        
+                                                   GLib.spawn_command_line_async(gksu+this.location); // Run tor from location settings        
         }
         else {
-            GLib.spawn_command_line_async('pkill tor');  // Kill Tor                 
+            GLib.spawn_command_line_async(gksu+'pkill tor');  // Kill Tor
```
jaszhix: I think there's two issues here. ```pkill``` is not really a default app people have installed, and I think the source of errors like this:

```
pkill: killing pid 404 failed: Operation not permitted
pkill: killing pid 421 failed: Operation not permitted
```

is caused by pkill killing the wrong process, that also happen to be owned by root.

Also, left-clicking the applet icon (which again, I would not expect to immediately _kill_ Tor), just hides the Tor window and displays this when I try to start a new process:

screenshot from 2017-08-06 14-18-28

Right after you kill Tor, this displays:

screenshot from 2017-08-06 14-22-23

I think you either should either not kill Tor, or use the Muffin API to do it.

Goodevil95:
@jaszhix I'm sorry for the bad description of my applet!
This applet don't launch browser. This is launch **only** Tor server (also known as Tor Expert Bundle). And you able to connect to Tor network from any browser or app. Just need set proxy settings in in the application you are using.
This applet detects enable or disable Tor server and add ability to launch or shutdown.

Tor server comes with browser ( located in `<directory with .sh>/Browser/TorBrowser/Tor/tor`).
And your browser showed error because Tor Server which he use was shutdown.

In Arch Linux distro Tor is in /usr/bin/tor. You can install Tor server separately or use Tor server that comes with browser (in applet setting you can set Tor path).

jaszhix:
Okay. I missed the tor path setting in the applet config, sorry. I think you should still find an alternative to ```pkill``` since its not included by default in Linux Mint, and probably other distros. Ideally, you could get the pid when spawning the process. Cinnamon has some abstractions around GLib in imports.misc.util that may be worth [checking out](https://github.com/linuxmint/Cinnamon/blob/4459c7fbc057b2d5145dcfd47748e866a05001dd/js/misc/util.js#L67).

tor-button@shatur/files/tor-button@shatur/applet.js
jaszhix: I don't think its a good idea to encourage running applications as root in the userland.

brownsr:
If you've not seen this, then there look to be some useful details here https://wiki.archlinux.org/index.php/Tor#Usage

Goodevil95:
@brownsr Yes, I have seen this.
@jaszhix Are you sure that I need to use `kill` instead `pkill`? In Cinnamon's Util uses `pkill` in [killall](https://github.com/linuxmint/Cinnamon/blob/4459c7fbc057b2d5145dcfd47748e866a05001dd/js/misc/util.js#L196) function.
When I use `kill` I can't kill Tor if it not running by applet.

jaszhix:
Ah, I guess I'm mistaken on pkill. 

One issue I noticed while testing again was I got random "Tor launched" and "Tor closed" notifications, even though tor wasn't running. You could try polling ```pidof tor```, and check stdout.

The Arch docs show tor being run with the tor user, not as root. I don't think browsers should run as root. If you can adjust that, I would feel more comfortable merging this.

Goodevil95:
> One issue I noticed while testing again was I got random "Tor launched" and "Tor closed" notifications, even though tor wasn't running.

Could you test this issue a bit more to tell me when it happens? I don't observe it on my system...
> You could try polling pidof tor, and check stdout

Or maybe you could help me to improve launch detection?
I don't know how to make it :(
> The Arch docs show tor being run with the tor user, not as root. I don't think browsers should run as root. If you can adjust that, I would feel more comfortable merging this.

In Arch need to run Tor Server (not a browser!) from package `tor` with sudo. You can check it on [Wiki](https://wiki.archlinux.org/index.php/tor#Usage) (systemctl requires root privilegies too). Then you can translate any traffic from any app through Tor.
That's why I added the option to run `Tor` with `gksu`. You can disable it in applet settings.

jaszhix:
True, I didn't mean to refer to it as a browser again. Unrelated note, you do lose some of the security/privacy benefits of tor when you use it in another browser not designed for that purpose. 

One option gives users the ability to customize the command they're launching, and another option makes it launch with gksu. Which I guess is fine if you know what you're doing, but it seems like enabling inexperienced users to unknowingly break something, or make their installation insecure.

You need sudo to run a command as a different user, but the command is not being run as root. Its running as the user specified with the -u flag. 

If you don't know how to fix it, you could try looking up the API references in devhelp, or follow the site I [linked](http://devdocs.baznga.org) earlier in the thread. I'm just a reviewer (also my opinions are my own here).

Goodevil95:
> Unrelated note, you do lose some of the security/privacy benefits of tor when you use it in another browser not designed for that purpose

Of course, but package `tor` is some distros or Tor Expert Bundle from official site designed for use Tor network without Tor Browser.

> You need sudo to run a command as a different user, but the command is not being run as root. Its running as the user specified with the -u flag.

Sorry. I'm mistaken on `sudo`. I changed applet to fix it.

Issue #1128, opened 2017-08-07
redshift@marvel4u: Icon does not scale with panel size


sphh:
Have you read the guidelines for contributing?

Before opening an issue, please read the guidelines for contributing linked above. Issues with invalid formats, or issues not mentioning the maintainer of the applet, may not get a response, and eventually will be closed. You can find the author by visiting the applet's page on the Spices website (https://cinnamon-spices.linuxmint.com/), or going to the applet's directory on this repository, and checking the the author field in info.json. If it is set to none in info.json, or there is no author listed on the site, then this step is not necessary.

[Notify author of applet]: @Marvel4U 

 - Distribution: Linux Mint 18.2 Sonya
 - Cinnamon version: Cinnamon 3.4.4
 - Applet version: 2.0.0

[Issue]: <> (Describe your issue below.)
As the subject already say: The applet's icon does not scale if I increase the size of the panel.

NikoKrause:
issues not mentioning the maintainer of the applet, may not get a response

sphh:
Thanks @NikoKrause for pointing this out.

I simply filled in the template when opening a new issue. This has the line

``[Notify author of applet]: <GithubAuthor>``

where I replaced the ``<GithubAuthor>`` with @Marvel4U but this line does not show up in the final message …

@Marvel4U could you please look into this bug report.

@NikoKrause could you please look into the template.

Thanks to both of you!

Martin1887:
This is weird, since icons are svg. I have modified my panel height in a exagerated way to check the issue and I think that the icons scale fine (red sunset):

![imagen](https://user-images.githubusercontent.com/4211771/29134122-fd4a8a6e-7d35-11e7-852c-ec3f8a4ed671.png)


Could you please provide a screenshot?

Thanks and regards.

sphh:
I set the slider for the panel height in the panel settings as far to the right ("larger") as possible. Please also note that it's a top panel and it's set to transparent with shadow in the "Transparent panels" extension. "Allow Cinnamon to scale panel text and icons according to the panel height" is "on".

![screenshot from 2017-08-09 23-13-07](https://user-images.githubusercontent.com/4856462/29146631-1c35e2e2-7d5a-11e7-8e0a-b70cbf3d5d79.png)

All but three applets scale to this height. I admit, it's hard to see the bottom, but the "line" in the CPU field shows the bottom line.

[...]

PR #1149, opened 2017-08-12
[redshift@marvel4u] Tooltip update with more info about Redshift state

Martin1887:
Tooltip update with more info about Redshift state and pot file regenerated.

Again, I only can update Spanish translation, the other translations should be updated.

@Marvel4U, are you ok with this Pull Request?

I have added a note about my contribution in README.md file. I hope that you don't mind it :-).


Marvel4U:
Heyo,
yes approved. Looks very nice.
Ofc good that you added yourself in the Readme. I'd also be fine if we're something like co-authors and can both approve mergers.

jaszhix:
It could be because I'm running Cinnamon git, but enabling this appears to hide all the applets in the far right section of my panel.

screenshot from 2017-08-14 06-28-01

I tested the version on Spices, and  it doesn't hide the other applets like this does for some reason, but the button is still stretched out. Also noticed this in glass.log:

```
info t=2017-08-14T11:21:47.242Z Error getting Cinnamon backlight power settings true,,Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown:
```

It could be my configuration, or the current state of Cinnamon on master, but I'm curious if anyone else is able to reproduce.

Martin1887:
@jaszhix, the Glasslog error is due to brightness settings. The applet tries to communicate with Cinnamon Settings daemon and it is not running or similar (in my scenario the Cinnamon settings daemon dies when Cinnamon starts, I don't know why). This error means that changing brightness slider has no effect, but all Redshift settings work fine.

For me the version of this pull request works fine, I cannot reproduce the error that you comment.


Regards.

jaszhix:
I spun up a Cinnamon 3.4 VM and couldn't reproduce it either. I rebuilt Cinnamon from master, and it appears to stretch the SVG icon out and overlap other applets in the panel section. I'm not sure why it does that, but manually setting the icon size:

```js
this._applet_icon.set_icon_size(24);
```
and placing that after the [```set_applet_icon_symbolic_path``` call](https://github.com/Martin1887/cinnamon-spices-applets/blob/2dd15ce8380384a256807d754dcbb35e285eb584/redshift%40marvel4u/files/redshift%40marvel4u/applet.js#L58) in ```_init``` fixes it.

Martin1887:
The author of the applet, @Marvel4U, has approved the pull request. So, I think it should be merged since icon issue has not any relevance with this pull request.

Icon problems when the panel has different height is being dealt in the issue #1128. I think that this discussion should be continued there.


Thanks and regards.

jaszhix:
I am suggesting it here because you know the code better than me. If I merge it now, then I will PR the fix regardless if it breaks something because it breaks the UI in Cinnamon dev.

Martin1887:
But is this pull request what breaks the UI? Have you tried if version 2.0 of the applet does break it?

It seems really weird to me, since this pull request only has changed the tooltip text.

jaszhix:
Where is version 2.0? Cinnamon git is always changing, I'm just trying to give a heads up. I don't blame you, you wouldn't have any way of knowing...

Martin1887:
Version 2.0 of the applet is the previous delivered version, the master in git of the applet just before you have merged this pull request.

PR #1150, opened 2017-08-13
Fixes for translation and checking Tor status

Goodevil95:
Add checking Tor status when applet initialize and add spaces in popup menu for correct translation.


Goodevil95:
I need to fix some bugs with this PR.

PR #1157, opened 2017-08-15
Fixes for translations and tor status check 

Goodevil95:
Fix translation support for popup menu and add tor status check when applet initialize


brownsr:
That command line sync call has the potential to freeze the UI.  If you can work out a way to do it async it would be safer.

Goodevil95:
@brownsr yes, I know it. With async command I can't read information from file after write:(
It would be great if I find Javascript command to check running program or not.

PR #1170, opened 2017-08-19
redshift@marvel4u: Ensure the icon size is 24px

jaszhix:
@Marvel4U @Martin1887 

Following up on discussion in #1149, this sets the icon size to 24px due to a change in behavior on SVG symbolic icons in Cinnamon 3.6.


Marvel4U:
Approved. I don't see any changed behaviour but that was the point of the discussion...

PR #1179, opened 2017-08-26
[redshift@marvel4u] Changed fixed 24px height to sending panelHeight to the applet

Martin1887:
Hello.

Fixed 24px size has not seemed to me a perfect solution, since when you do changes in the panel size the size changes and I think that 24px is a too much big size.

For me this works fine. @jaszhix and @Marvel4U, could you test if it breaks nothing and works totally fine?

I have also changed tabs to spaces in applet.js because they was mixed.

Thanks and regards.


jaszhix:
Good catch, works for me as well.

Marvel4U:
works perfectly.

PR #1181, opened 2017-08-28
multicore-sys-monitor@ccadeptic23: Fix glibtop warning, refactor DataProviders, update PO files, update JS formatting

jaszhix:


SerafDosSantos:
@jaszhix Since this update the applet is really running wrong for me. Networking monitor is not running well. I can't provide logs. Someone else might better provide more info on this; I'm just mentioning it.

PR #1200, opened 2017-08-31
stevedore@centurix: Fix imports for Cinnamon 3.6

jaszhix:
@Centurix 

The TerminalReader import was removed from docker.js because it was
unused.


Centurix:
Based on the comments from another branch and testing, I'm fine with this.

PR #1201, opened 2017-08-31
rancher@centurix: Fix imports for Cinnamon 3.6

jaszhix:
@Centurix 

The util import is removed from homestead_yaml_reader.js because it is
unused.


Centurix:
Thanks. I understand that this is a change for 3.6, but this fix breaks the applet for previous versions. Unless there's some mechanism in this repo to only distribute this applet for specific versions? (Not aware of such a mechanism, and not everyone upgrades to the latest Cinnamon immediately)

jaszhix:
No problem. Importing through appletManager is backwards compatible with previous versions of Cinnamon. I'm not sure exactly when, but at least Cinnamon 2.6+.

Centurix:
ok, thanks!

PR #1253, opened 2017-09-15
Favorites@bownz: remove applet

NikoKrause:
This applet is abandoned and broken (empty popup menu).
Favorites can be accessed through the menu applet, so there's no benefit in keeping/fixing this applet.


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/65

screenshot-area-2017-09-17-143319

PR #1256, opened 2017-09-15
brightness@markbokil.com: remove applet

NikoKrause:
this applet does not work, the display brightness does not change at all.
Display brightness can be changed through the default power cinnamon applet, therefore this applet is obsolete


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/64

screenshot-area-2017-09-17-143942

PR #1261, opened 2017-09-16
netctlstatus@prmurthy: remove applet

NikoKrause:
applet is abandoned and shows only a blank space in the panel


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/152

screenshot-area-2017-09-17-144528

PR #1262, opened 2017-09-16
netctlsystraymenu@prmurthy: remove applet

NikoKrause:
applet is abandoned and broken (shows only a blank space in the panel)


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/153

screenshot-area-2017-09-17-144640

PR #1264, opened 2017-09-16
system-monitor@ebbes: remove applet

NikoKrause:
abandoned and broken, also very old (still uses gschema)

and there are a few other system monitors


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/90

screenshot-area-2017-09-17-144917
screenshot-area-2017-09-17-144931

PR #1279, opened 2017-09-19
windowlist@cobinja.de: add makepot

NikoKrause:
Usage: Run `./makepot` in the `.po` directory.

This script creates a .pot file and then deletes Strings, you want to exclude from translation, e.g. to exclude strings, which are already translated in Cinnamon itself.


Cobinja:
Not really needed, so closed (after talking to NikoKrause about it)

PR #1302, opened 2017-09-24
lookingglass-restart@markbokil.com: remove applet

NikoKrause:
redundant applet, score: 0
Those settings are available on panel rightclick


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/66
screenshot-area-2017-09-24-171429

PR #1314, opened 2017-09-27
super-window-list@bgstack15: add new applet

bgstack15:
Add new applet which offers a slight improvement over the stock window
list: the toggle setting for "Show windows from all workspaces."


jaszhix:
> The author has little intention of carrying this forward, so if anyone volunteers to maintain/update/delete this applet, please feel free.

I'm not sure this should be merged if its not going to be maintained. Have you thought about opening a PR on Cinnamon so the original window list has this feature?

bgstack15:
I hadn't thought of it. I will look into opening a PR against the "Cinnamon" repo serving file https://github.com/linuxmint/Cinnamon/tree/683409cd9a76f02ba71822b390e0aa7bcb398da6/files/usr/share/cinnamon/applets/window-list%40cinnamon.org. Thanks!

PR #1338, opened 2017-10-01
softyubikey@yubiserver.include.gr: add da.po

Alan01:


NikoKrause:
I see you do a lot oft translations. That's great.
If you want to you can do one PR with separated commits for each applet.

PR #1347, opened 2017-10-03
added applet ipindicator@matus.benko@gmail.com

PrimaMateria:
Hi,
I just realized that you have new spices github project (which is really cool comparing to the old system).
I would like to add my older applet for VPN users.


jaszhix:
Cool applet, it seems to work well. I noticed one small issue, though.

`Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/ipindicator@matus.benko@gmail.com/applet.js 122]: assignment to undeclared variable IpGateway`

This is not crashing the applet because of the try-catch block in your applet's `_init` method, but I recommend fixing it because it may crash the applet in newer versions of Cinnamon. You just need to declare `Debugger` and `IpGateway` with `let` or optionally `const` if they are not re-assigned later.

PR #1386, opened 2017-10-11
cinnamon-timer@jake1164 - Forking axos88@countdown-timer

jake1164:
This is a fork of the countdown-timer. I resolved a large number of the issues and will maintain the applet but wanted to change the UUID to something more standard. 

Let me know if this is an issue since the applet is tagged without an author.


NikoKrause:
It's fine. We'll delete axos88@countdown-timer.

NikoKrause:
The icon.png symbolic link is broken.

NikoKrause:
Is the .gitignore file necessary?

NikoKrause:
I would shorten the description:
`A timer app with visual and auditory notifications`

The following lines are unused:
```
const Cinnamon = imports.gi.Cinnamon;
const Gtk = imports.gi.Gtk;
const AppletMeta = imports.ui.appletManager.applets[UUID];

function PopupMenuItem(label, icon, callback) {
    this._init(label, icon, callback);
}

```

NikoKrause:
Also you can create a pot file with a script in the cinnamon-spices-applets repo, see here:
https://github.com/linuxmint/cinnamon-spices-applets#translations
```
./cinnamon-spices-makepot cinnamon-timer@jake1164
```

And you can use/copy the `.po` files of axos  applet. It has already some strings translated. So it will only need to be updated by translators instead of having to translate it all over again.

jake1164:
Made the requested changes. 

Followup question on the translations, in the future when a change is made just re-run the script to update the pot file and then does anybody need to be notified or is that all automated now?

This app is also ~90% the same as timer-notifications that does not have an author, should it replace both that and axos88?

Let me know if there are any further issues or questions.

NikoKrause:
You as developer just need to update the `.pot` file.

The translator updates the `.po` file of his language against the `.pot` file. There's no need for you to do that.

We can remove both applets. axos88 forked mbokils version. We don't need all 3 applets, if they do the same.

NikoKrause:
You set the symbolic as an absolut path. It works on your machine, but won't work on others. You need to create a relative path:
```
ln -s files/UUID/icon.png
```

jake1164:
Sorry. I updated the link.

PR #1389, opened 2017-10-15
axos88@countdown-timer: applet removal

NikoKrause:
abandoned applet was forked


NikoKrause:
https://cinnamon-spices.linuxmint.com/applets/view/280

PR #1402, opened 2017-10-21
New applet BgMedia@spacy01

spacy01:
This is applet for streaming bulgarian radios and TVs.

PR #1406, opened 2017-10-22
New Applet BgMedia@spacy01

spacy01:
This applet is fork on the previous BgRadio@spacy01 applet.


jaszhix:
I'm confused, are you still maintaining BgRadio@spacy01? I think this applet is really similar and both of their functionality could probably be in one applet. You also didn't close the original PR.

spacy01:
@jaszhix I don't have a intention to maintaining the old BgRadio applet. But i don't suggest to delete it, because probably have people who used it.

Also i cannot just update BgRadio, because the name will not be accurate any more. This new applet now have tv streams too. That's why i leaved both of them...

NikoKrause:
If you don't habe intention in maintaining the old BgRadio applet, you should delete it, if your new applet is a fork of your old applet.

Other Option would be, that you update the old applet with your new features. You can change the name in metadata to reflect that it has TV stream now. All you have to keep is the old UUID.

NikoKrause:
If you keep the old UUID the people, who use your BgRadio applet, can easily upgrade to BgMedia.

jaszhix:
Okay, there's also [another applet](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/radio%40driglu4it) that was forked from your radio applet that's still being maintained. Both applets have the ability to customize radio stations.

Applets sometimes break when Cinnamon is updated, and it puts more strain on us if there's more applets being fixed than there needs to be. I would only submit this one if you intend to maintain it.

I still see the re-declaration of the gettext variable I mentioned in #1402.

spacy01:
Ok, i will update the old one, keeping the UUID, like @NikoKrause said. And will fix the gettxt.

PR #1407, opened 2017-10-22
Update BgRadio@spacy01

spacy01:
So, the applet is completely rewritten, keep the old UUID. Gettext duplicate is fixed.


NikoKrause:
Plesse check your applet with this script and fix the errors it reports:
https://github.com/linuxmint/cinnamon-spices-applets/blob/master/README.md#validation

spacy01:
Ok, @NikoKrause, dangerous flag is been removed. :)

PR #1414, opened 2017-10-26
Update sticky.pot

spacy01:
Added a missed string. @collinss


collinss:
Thanks, I forgot to update the pot file after adding a new feature.

PR #1415, opened 2017-10-27
feeds applet: fix some warnings in .xsession-errors

collinss:
@jake1164 The last one looks like it was a typo, but it also required me to reverse the logic of the conditional for it to work right (I think it worked before - imperfectly - because it was comparing undefined with whatever this.unread_count is, which will always evaluate to false, causing the unread count to be used no matter what). You know the code better than I do. Could you look this over and make sure I got everything right?


jake1164:
Just checked and everything looks good.. thanks!

PR #1416, opened 2017-10-27
bg.po for CinnVIIStarkMenu@NikoKrause

spacy01:
Bulgarian translation.


NikoKrause:
Thanks for translating

PR #1423, opened 2017-10-30
ioDisk@ctrlesc: Fixed regex parsing to handle the updated iostat (11.…

ctrlesc:
…6.0) disk metrics and better stats alignment on the popup.

@NikoKrause this should fix the applet issue with newer versions of sysstat


NikoKrause:
@ctrlesc Are you planning on maintaining this applet again? If yes, I'd set you back as author. It was set to none, because it was assumed to be abandoned.

ctrlesc:
@NikoKrause, you can set me as the author. I can handle the maintenance from here on.  Thanks again for the help.

PR #1424, opened 2017-10-30
ioDisk@ctrlesc: Return of the author

NikoKrause:
Set author back to @ctrlesc

Remove unused imports and trailing whitespaces

Add German translation


NikoKrause:
@ctrlesc 
Is this ok?

PR #1448, opened 2017-11-08
cinnamon-timer@jake1164 - Changed settings to use individual hour, min and sec times.

jake1164:


NikoKrause:
If I were you I would change order second,minutes,hours to hours,minutes,seconds.
Seems more natural to me.

Also I would add a range (min/max) for seconds and minutes from 0 to 59.

jake1164:
excellent points, I'll make those changes in a few hours.  I am guessing I should also re-run the pot script now that I think about it?

NikoKrause:
Yeah, if it contains new Strings to translate

jake1164:
@NikoKrause done.

PR #1480, opened 2017-11-19
bumblebee@pdcurtis: update hrv lng

muzena:


jaszhix:
It looks like this is PO file is for `bumblebee@pdcurtis`.

Issue #1502, opened 2017-11-27
temperature@fevimu: Autoclose popup menu


germanfr:
Hey, could you apply this? It's such a simple thing that I don't want to create a fork for that.
```diff
diff --git a/temperature@fevimu/files/temperature@fevimu/applet.js b/temperature@fevimu/files/temperature@fevimu/applet.js
index b5d9eda..aa018bb 100644
--- a/temperature@fevimu/files/temperature@fevimu/applet.js
+++ b/temperature@fevimu/files/temperature@fevimu/applet.js
@@ -44,7 +44,9 @@ MyApplet.prototype = {
 		try {
 
 			// Create the popup menu
+			this.menuManager = new PopupMenu.PopupMenuManager(this);
 			this.menu = new Applet.AppletPopupMenu(this, orientation);
+			this.menuManager.addMenu(this.menu);
 
         	this.sensorsPath = this._detectSensors();
 
```

NikoKrause:
Thanks!

PR #1512, opened 2017-11-28
temperature@fevimu: Added setting to reduce precision

leplastic:
Hi, I created a setting to reduce the precision of the temperature. In my case, I found the value always changing a bit distractive. With this setting, only the integer part of the value is show for both Celsius and Fahrenheit (28 instead of 28.5).

I labeled the setting as "Show only integer part of temperature". I feel it is a bit geekish for public in general. I'm accepting better suggestions for this if the change is accepted.

@fevimu @NikoKrause

### DIFF
--- a/temperature@fevimu/files/temperature@fevimu/applet.js
+++ b/temperature@fevimu/files/temperature@fevimu/applet.js
@@ -29,6 +29,7 @@ MyApplet.prototype = {
         this.settings = new Settings.AppletSettings(this, UUID, instance_id);
 
         this.settings.bind("use-fahrenheit", "use_fahrenheit", this._update_temp);
+        this.settings.bind("only-integer-part", "only_integer_part", this._update_temp);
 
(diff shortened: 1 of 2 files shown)


NikoKrause:
It can be merged. Maybe you can say `Round temperature to integer`?
Not sure either. Let me know if you want to change it or leave it.

Also can you do me a favor? Can you add those lines, which are suggested here into your pull request? https://github.com/linuxmint/cinnamon-spices-applets/issues/1502

NikoKrause:
Or maybe
```
Use integer values ​​for temperature
```

leplastic:
I followed your last suggestion, which sounds better. I also added the lines from #1502 PR and the popup now closes when clicking somewhere else.

NikoKrause:
Thank you very much.
Also thanks for @germanfr for the autoclose of the popup menu.

PR #1518, opened 2017-11-29
temperature@fevimu : Added European Portuguese translation

leplastic:
Hi again. I noticed the applet was missing a Portuguese translation, so I created one for European Portuguese (pt_PT). It already includes the new setting, so the pot file was updated as well.

@fevimu @NikoKrause (sorry for not including this on last PR, this is my last change)


NikoKrause:
Can you rename the file (and also in the file) from pt_PT to pt only.

Here's our list of languages and their IDs:
https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.translation-tables/LINGUAS

P.S. You don't have to be sorry.

leplastic:
Ok, I think it's correct now :)

PR #1521, opened 2017-11-30
[[Feature]] Display World Times in a Non-Bold Font Style – calendar@simonwiles.net

WammKD:
It more or less boils down to personal preference but I do think there's reasonable logic for this change:

bolding confuses the displayed times with the same level of importance as the date at the top of the applet and tugs the eyes in multiple directions in terms of where they're supposed to give focus.

Making the world times un-bold makes them stick out since they don't share the same style as the Date at the top and that the rest of the calendar likely doesn't share the same font size as the world times.

@simonwiles


NikoKrause:
@simonwiles

Any complains?

simonwiles:
Yes, it seems like it's more an individual preference than anything else.  I suspect mine leans towards having it in bold, tbh, but I'm fine with merging this and we'll see if there's any feedback.  Doesn't make a world of difference either way, of course, but you make an interesting argument so let's give it a try.  Cheers!

PR #1526, opened 2017-12-03
cryptocoin@guantanamoe: Track value of cryptocurrencies, Initial move to cinnamon-spices

imDeprecated:


jaszhix:
This works for me. I noticed one minor warning in xsession_errors:

```
Cjs-Message: JS WARNING: [/home/jason/.local/share/cinnamon/applets/cryptocoin@guantanamoe/applet.js line 88]: reference to undefined property response.Response
```

Doesn't affect anything though, maybe check the truthiness of `response` before accessing response.Response.

Also, if you copy icon.png to `/cryptocoin@guantanamoe/files/cryptocoin@guantanamoe/`, it will show the icon in the applet manager.

imDeprecated:
Thanks for reviewing, I'll fix warnings in follow up pull request

PR #1549, opened 2017-12-06
weather@mockturtl: fix broken location link

### LINKED ISSUE #1547: weather@mockturtl. The city button is set to direct to us.rd.yahoo.com. Link is dead.
* 1.14.0 (22-10-2017
 * Cinnamon version 4.6
 * Distribution - Mint 18.3
 * Intel 3rd Gen Core processor Graphics Controller, modesetting driver

**Issue**
The city button is set to direct to http://us.rd.yahoo.com/dailynews/rss/weather/Country__Country/*https://weather.yahoo.com/country/state/city-*** which is not working.
Just the https://weather.yahoo.com/country/state/city-*** works fine. It needs to be changed.

mockturtl:
Fixes #1547

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -582,7 +582,8 @@ MyApplet.prototype = {
       this._currentWeatherPressure.text = pressure + ' ' + _(this._pressureUnit)
 
       // location is a button
-      this._currentWeatherLocation.url = weather.link
+      let tmp = weather.link.split('*')
+      this._currentWeatherLocation.url = tmp.length > 1 ? tmp[1] : tmp[0]
       this._currentWeatherLocation.label = _(location)
 
       // gettext can't see these inline


mockturtl:
Requesting a review from someone familiar with js, I tested the patch but I'm not confident it's free of dumb mistakes.

weather@mockturtl/files/weather@mockturtl/applet.js
```diff
@@ -582,7 +582,8 @@ MyApplet.prototype = {
       this._currentWeatherPressure.text = pressure + ' ' + _(this._pressureUnit)
 
       // location is a button
-      this._currentWeatherLocation.url = weather.link
+      let tmp = weather.link.split('*')
```
jaszhix: This might break if the link property in the AJAX response doesn't have an asterisk in the string for some reason in the future. You could guard it with something like:

```js
let tmp = weather.link;
if (weather.link.indexOf('*') > -1) {
  tmp = weather.link.split('*')[1]
}
this._currentWeatherLocation.url = tmp;
```
Though its probably a new API that won't change for a while, is my guess.

PR #1559, opened 2017-12-11
weather@mockturtl: Updated polish language

mariachini:


jaszhix:
Please make sure the UUID is in the PR title so we know which applet this is from the list.

PR #1565, opened 2017-12-20
sshlauncher: make icon appear

tve:
I'm running Cinnamon 3.6.6 and the applet icon does not appear in the panel. This commit fixes that for me but I don't have a way to test it on various versions of cinnamon and the docs about how icons work is rather thin. @sumo


jaszhix:
This is a bug fix making the applet usable, so merging.

PR #1570, opened 2017-12-22
CinnVIIStarkMenu: French description update

ebourg:
Hi,

In the french translation the name "Cinnamon" has been translated to "Cannelle", but since it's a proper name it shouldn't be translated.


NikoKrause:
Thanks

PR #1612, opened 2018-01-31
sound150@claudiux - v1.3.0

claudiux:
Hi,

Can you merge my sound150 branch with your master branch, please?

Changelog v1.3.0:

 * No more stylesheet.css; uses theme's style sheet for better look.
 * Colors become optionnal.
 * Applet is now able to increase the volume up to maximum of capabilities of the sound card. But to let a security margin, this option is not proposed to user.
 * Set scripts/generate_mo.sh executable. (Fixes a bug.)
 * Set default max volume to 150%.

Best regards,

Claudiux


claudiux:
The commit 1c52359 fixes a bug that prevented restarting the applet after installing new .mo files.

PR #1613, opened 2018-01-31
sound150@claudiux: v1.3.1 - Fixes a bug

claudiux:
Hi,

After installing the .mo files, the v1.3.0 sound150@claudiux applet doesn't reload properly. This v1.3.1 fixes this bug (now, the applet is no more reloaded; new translations will be waiting the next start up of Cinnamon).

Please, can you merge this sound150 branch with your master branch?

Best regards.

Claudiux


sound150@claudiux/files/sound150@claudiux/applet.js
```diff
@@ -1211,7 +1216,8 @@ MyApplet.prototype = {
     _onIconThemeChanged : function(actor, event) {
         //log('_onIconThemeChanged');
         //imports.ui.appletManager.applets.reloadExtension(UUID, 'APPLET');
-        Extension.reloadExtension(UUID, Extension.Type['APPLET']);
+        imports.ui.appletManager.applets.reloadExtension(UUID, Extension.Type.APPLET);
```
jaszhix: Don't think this will work, `reloadExtension` is only in extension.js. Cinnamon 3.6 has a regression where applet do not reload correctly. Its fixed in 3.8, but you will then need to use the new module importer for reloading to work again. Relevant PRs:  https://github.com/linuxmint/Cinnamon/pull/6878 and https://github.com/linuxmint/Cinnamon/pull/6790. [Here's an example](https://github.com/linuxmint/cinnamon-spices-applets/blob/582e749298f18f58d6e192c5671777f098e1bb6c/IcingTaskManager%40json/files/IcingTaskManager%40json/3.4/applet.js#L30) of how I made my applets use the new importer, and fall back to the old importer if on Cinnamon < 3.8 (git master).

claudiux:
Many thanks for all, @jaszhix !

PR #1638, opened 2018-02-24
hideable-applets@cardsurf: Fix for Cinnamon 3.8

jaszhix:
- Fixed strict mode errors
- Fixed Glib warnings
- Switched applet over to new importer so it reloads correctly

@cardsurf


cardsurf:
Good job. I approve the changes.

PR #1643, opened 2018-02-27
cryptocoin@guantanamoe: unlimited instances and multiple coins

bkmeneguello:
- Enable multiple instances
- Each instance is a pair Coin/Crypto. The list of coins and cryptos is fixed

### DIFF
--- a/cryptocoin@guantanamoe/files/cryptocoin@guantanamoe/applet.js
+++ b/cryptocoin@guantanamoe/files/cryptocoin@guantanamoe/applet.js
@@ -11,39 +11,22 @@ const Applet = imports.ui.applet;
 const Settings = imports.ui.settings;
 const Gettext = imports.gettext;
 const Soup = imports.gi.Soup;
-
-imports.searchPath.push( imports.ui.appletManager.appletMeta[UUID].path );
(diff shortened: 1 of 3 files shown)


jaszhix:
Let's ping @imDeprecated for review.

imDeprecated:
Tested, I think some work needs to be done in the usability of the settings.  The long lists take a minute to load on my dev laptop (old thinkpad).

Also there seems to be loss in functionality due to only tracking one ticker now.

Tickers would need to be searchable and individually added to a tracked set in order to bring in an additional 4k of them.  I just don't see the usability in scrolling for 30 secs to get down to XRP

bkmeneguello:
Well, about the long time, it really takes, but just a little delay for me. I'll search for a better way to list all cryptos without so much delay.
About the functionality loss, it's not missed, since now the applet allows multiple instances each with a different setting.
I've sorted the cryptos by its ticker name, to facilitate the search, since cinnamon combos are not searchable.

bkmeneguello:
Fixed with free input

NikoKrause:
@imDeprecated ?

imDeprecated:
Sorry, approved

PR #1659, opened 2018-03-07
multicore-sys-monitor@ccadeptic23: Fix applet geometry bugs.

evg-zhabotinsky:
Fixed applet graphics being slightly off-center and clipped.
(Width and height were off by a few pixels.)

Fixed incorrect vertical scaling in HiDPI.
(`panel_height` is already scaled and scaling factor was applied once more to it, leaving half the applet off-screen at 2x scale)

@jaszhix

### DIFF
--- a/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/3.4/Graphs.js
+++ b/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/3.4/Graphs.js
@@ -105,12 +105,8 @@ GraphPieChart.prototype = {
 
     //Draw Pie Chart
     let xcenter = width / 2;
-    let ycenter = (height / 2) / global.ui_scale;
-    let radius = width / 2 - 1;
-    if (height < width) {
-      radius = height / 2 - 1;
-    }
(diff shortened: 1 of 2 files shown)


jaszhix:
Part of my attempt at high DPI support was trying to get the labels to scale as well, but apparently that scaled the entire applet incorrectly. Current version:
image

PR:
image

Not sure what the solid lines at the bottom of the network and disk graphs are from, probably was always there, but hidden from the applet being positioned too far down.

Anyway, it looks good, definitely more usable for high DPI users. I don't have a high DPI display, so Cinnamon is hard to use for very long with it on. Thanks for taking time to fix this!

evg-zhabotinsky:
That was fast. Too fast, in fact.
I don't use labels so I forgot about them. Fix is the same as with the pie chart: don't divide size by scale as its already a function of height.
I guess I have to open another PR.

jaszhix:
I went ahead and merged it because it was a usability issue... the github integration adds the `In Progress` tag to every PR, I recommend adding "WIP" to the title.

evg-zhabotinsky:
It was not intended as a WIP. Just saying that after finding a less important bug introduced by a fix for more important bug it might still be worthwhile to wait for a reply, that's all.

PR #1709, opened 2018-03-18
radio@driglu4it Update for applet

Driglu4it:


NikoKrause:
Which applet? Please add it in the title.
What does the PR do?

PR #1711, opened 2018-03-18
radio@driglu4it Update ru.po

Driglu4it:


NikoKrause:
Please add the applets uuid/name in the titles, since you're opening a bunch of PRs.

PR #1716, opened 2018-03-22
Revert "IcingTaskManager@json: Fix exported variables (#1702)"

jaszhix:
The Cinnamon 3.8 xlet importer fakes GJS module export compatibility by picking out modules from a file with regex, but it currently does not work if the declaration is on the very first line of the file. Internally it is taking these namespaces and adding them to the `exports` object, so it can be fixed by appending an exports assignment at the end of the file for scenarios like these.


claudiux:
Yes, I will try to fix this, as I did in my applet sound150@claudiux : https://cinnamon-spices.linuxmint.com/applets/view/306

In the sound@cinnamon.org, the step represents 5% of the maximum volume, that is 7.5% of the volume if the maximum volume is set to 150%. Taking rounding into account, this gives steps of 7% or 8%.

PR #1722, opened 2018-03-28
Cinnamenu@json: Adapting of the position of the categories 'all appli…

claudiux:
…cations' and 'favorite applications' to that of the panel (bottom or not)

Hi @jaszhix,

I propose a small modification, so that the categories 'All applications' and 'Favorite applications' are at the bottom if the Cinnamenu is in the panel located at the bottom of the screen, and at the top in the other cases.

Hope this will suit you.

Best regards.

claudiux

### DIFF
--- a/Cinnamenu@json/files/Cinnamenu@json/applet.js
+++ b/Cinnamenu@json/files/Cinnamenu@json/applet.js
@@ -272,6 +272,14 @@ CinnamenuApplet.prototype = {
         this.refresh();
       });
 
+      this.panelEditModeChangedId = global.settings.connect('changed::panel-edit-mode', Lang.bind(this, function() {
+         if (!global.settings.get_boolean('panel-edit-mode')) {
+            this.state.orientation = this._orientation;
+            let list_of_enabled_applets = global.settings.get_strv("enabled-applets");
+            this.refresh();
+         }
+      }));
+
       // FS search
       this._fileFolderAccessActive = false;
       this._pathCompleter = new Gio.FilenameCompleter();
@@ -454,6 +462,7 @@ CinnamenuApplet.prototype = {
 
   on_applet_removed_from_panel: function() {
     Main.keybindingManager.removeHotKey('overlay-key-' + this.instance_id);
+    global.settings.disconnect(this.panelEditModeChangedId);
     if (!this.settings) {
       return;
     }
@@ -704,6 +713,11 @@ CinnamenuApplet.prototype = {
       cb: this._updateIconAndLabel
     },
     {
+      key: 'fav_and_all_apps_pos',
+      value: 'favAndAllAppsPos',
+      cb: this.refresh
+    },
+    {
       key: 'startup-view-mode',
       value: 'startupViewMode',
       cb: this.refresh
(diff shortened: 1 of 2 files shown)


jaszhix:
Hi @claudiux, thanks for the PR! This change could probably grow on me, but I think it should be an option that can be toggled, and disabled by default since its what current users are accustomed to. One thing I noticed was when the panel is moved to another side of the screen, it doesn't update the categories.

jaszhix:
Oh one other thing - you might be interested in this [discussion](https://github.com/jaszhix/Cinnamenu/issues/12) where I was considering the possibility of drag and drop re-ordering of categories.

claudiux:
Hi @jaszhix, there is now an option in settings. When this option "Position of Favorite Applications and All Applications menus" is set to "Near the Cinnamenu Icon", this feature is activated.

Now, the categories are updated whenever the "Edit panel mode" is closed.

I will think about the possibility of drag and drop re-ordering of categories. (With a button to activate/deactivate drag and drop re-ordering.)

Cinnamenu@json/files/Cinnamenu@json/applet.js
```diff
@@ -454,6 +462,7 @@ CinnamenuApplet.prototype = {
 
   on_applet_removed_from_panel: function() {
     Main.keybindingManager.removeHotKey('overlay-key-' + this.instance_id);
+    global.settings.disconnect(this.panelEditModeChangedId);
```
jaszhix: If using signalManager instead, this could be handled with the [disconnectAllSignals](https://github.com/claudiux/cinnamon-spices-applets/blob/d1d3ae27fa0303910ab0be06848bfebf55149986/Cinnamenu%40json/files/Cinnamenu%40json/applet.js#L2432) call.

Cinnamenu@json/files/Cinnamenu@json/applet.js
```diff
@@ -272,6 +272,14 @@ CinnamenuApplet.prototype = {
...
       });
 
+      this.panelEditModeChangedId = global.settings.connect('changed::panel-edit-mode', Lang.bind(this, function() {
+         if (!global.settings.get_boolean('panel-edit-mode')) {
+            this.state.orientation = this._orientation;
+            let list_of_enabled_applets = global.settings.get_strv("enabled-applets");
+            this.refresh();
+         }
+      }));
+
```
jaszhix: Some nitpicky style things:
 - Use arrow functions over Lang.bind - easier for debugging
 - Connect signal using `this.signals` with signalManager (so it will be disconnected with the other signals on removal)
 - Use camelCase variable names over under_score variable names for consistency

This doesn't work though - you can move a panel without going into panel edit mode, so it won't update in that case. We could move this to [on_orientation_changed](https://github.com/claudiux/cinnamon-spices-applets/blob/d1d3ae27fa0303910ab0be06848bfebf55149986/Cinnamenu%40json/files/Cinnamenu%40json/applet.js#L449) - orientation change is already handled there, if its different than the old orientation, then do the re-render. 

In this case, its actually re-rendering the applet twice - once when going into panel edit mode, and another time when moving the panel. I think it does't propagate on the first re-render - which is why it didn't work in `on_orientation_changed`. I would prefer to look for a more specific rendering selectivity for categories. I appreciate the effort you're putting in here and I agree some customization should be handled here, but it might be easier for me to refactor category building so drag and drop re-ordering works. When that change is implemented, it probably would be cleaner to not keep this option around, since it has a narrower scope for customization than DND category re-ordering.

claudiux:
OK. Thanks for the tips.

So, should I close this PR ?

jaszhix:
No problem. Sure, for now - I'll let you close it so you can re-open it if something changes.

claudiux:
Hi @jaszhix. Congratulations for the 3.2.0 version of Cinnamenu. The drag'n'drop category re-ordering is amazing and really practical !

Issue #1723, opened 2018-03-29
ScreenShot applet does not scale properly


CottonEaster:
What the title of this report, or some sort of title with the applet? Anyhow, that 'UUID' is: ScreenShot@tech71

```
 * Applet version/Build date (to be found in "About.." or in "metadata.json") 2017-10-22
 * Cinnamon version (cinnamon --version) 3.6.7
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...) Mint 18.3
 * Graphics hardware *and* driver used: Intel Kabylake GT1.5; 3.0 Mesa
 * Snippet or attachment of ~/.cinnamon/glass.log: I found nothing that looked relevant.
 * Snippet or attachment of ~/.xsession-errors
29/03/2018 09:37			2018-03-29 09:28:44,995 | [1;33mWARNING[0m | ulauncher.util.image_loader: load_image() | Could not load image /usr/share/pixmaps/nemo.xpm. E: g-file-error-quark: Failed to open file '/usr/share/pixmaps/nemo.xpm': No such file or directory (4)
29/03/2018 09:37			SIOCGIFADDR: Cannot assign requested address
29/03/2018 09:37			mintUpdate: no process found
29/03/2018 09:37			Cinnamon warning: Log level 16: STACK_OP_LOWER_BELOW: sibling window 0x720000b not in stack
29/03/2018 09:37			Icon theme "oxygen" not found.
29/03/2018 09:37			Icon theme "oxygen" not found.
29/03/2018 09:37			Error: standard icon theme "oxygen" not found!
29/03/2018 09:37			ksystemlog:  Log manager is not yet initialized
29/03/2018 09:37			doesn't know "/var/log/syslog"
29/03/2018 09:37			Cinnamon warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x7401ca1 (Open Locat)
29/03/2018 09:37			Cinnamon warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x7402731 (Open Locat)
29/03/2018 09:37			doesn't know "/home/nicholas/.xsession-errors"
29/03/2018 09:38			Cinnamon warning: Buggy client sent a _NET_ACTIVE_WINDOW message with a timestamp of 0 for 0x74046a0 (Save selec)

 ```

 That list says no author is assigned. (Reporting this bug could be easier, I am finding!)

**Notify author of applet**

**Issue**

[...]

PR #1728, opened 2018-04-02
Screenshot+Record Desktop@tech71 - fix icon resizing bug, allow multiple instances & clarify correct package names for dependencies.

smurphos:
…

Same bug fixed in this applet as reported and fixed for Screenshot@tech71 - #1723 & #1727.

Also amended description slightly to reflect package names of dependencies rather than the command names.

### DIFF
--- a/ScreenShot+RecordDesktop@tech71/README.md
+++ b/ScreenShot+RecordDesktop@tech71/README.md
@@ -1,0 +1,11 @@
+### Screenshot & Record Desktop
+
+#### Dependencies
+
+To use this applets recording functionality you must have the following packages installed.
+
+* ffmpeg
+* xdotool
+* x11-utils
+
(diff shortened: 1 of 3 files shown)


NikoKrause:
I wouldn't put the dependencies into the metadata.
It's better to do this in the README as e.g. https://cinnamon-spices.linuxmint.com/applets/view/179

smurphos:
Thanks, reasoning for keeping the dependencies in the description is that end-users who don't visit the spices page and download via the Applet module only will still be alerted to the dependencies via the Applet module description and via the About box once installed.

I can add a short README as-well if you prefer, but still think it's a good idea to note the dependencies in the description.

NikoKrause:
One reason why I'm against having dependencies in the metadata is because the description is translatable. Let's keep it for now and add a README as well.

PR #1738, opened 2018-04-05
bash-sensors@pkkk: Dynamic tooltips

sphh:
I wanted to have a custom tooltip for the bash-sensors@pkkk applet. Because I already touched the tooltip display I added the possibility to dynamically generate it.

I also noticed that the label is not automatically updated when the applet is (left-)clicked. This is fixed. Now it makes sense to add the option to disable the automatic updates, so that the label and tooltip only gets updated when the applet is clicked.

Is the pot file automatically updated?

- Refresh interval can be set to 0 to disable automatic updates (updates of the labels happen only if the applet is left-clicked).
- Tooltip can be dynamic using the output of a special command.
- Command output when left-clicking on the applet is trimmed on the right-hand side removing the trailing newline character which makes the output higher than required.
- Applet label and tooltip is immediately updated when the applet is left-clicked.

### DIFF
--- a/bash-sensors@pkkk/files/bash-sensors@pkkk/applet.js
+++ b/bash-sensors@pkkk/files/bash-sensors@pkkk/applet.js
@@ -25,7 +25,6 @@ MyApplet.prototype = {
     _init: function (metadata, orientation, panelHeight, instance_id) {
         Applet.TextApplet.prototype._init.call(this, orientation);
         this.path = metadata.path;
-        this.set_applet_tooltip(_("Bash Sensors!"));
         this.settings = new Settings.AppletSettings(this, UUID, instance_id);
 
         this.menuManager = new PopupMenu.PopupMenuManager(this);
@@ -35,7 +34,7 @@ MyApplet.prototype = {
         this.menuLabel = new St.Label({text: '<init>'});
 
         this.bind_settings();
-        this.update();
+        this.autoupdate();
 
         item.addActor(this.menuLabel);
         section.addMenuItem(item);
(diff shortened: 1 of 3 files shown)


jaszhix:
> Is the pot file automatically updated?

No, you can update the reference pot file with cinnamon-json-makepot, though its been renamed to [cinnamon-xlet-makepot ](https://github.com/linuxmint/Cinnamon/pull/7149) on Cinnamon git with small differences in arguments. Then use msgmerge to update the translation PO files based on the reference file. I use some scripts to automate it and make it easier/quicker for me. 

[Example cinnamon-xlet-makepot script](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/Cinnamenu%40json/locale.sh)
[Example cinnamon-json-makepot script](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/IcingTaskManager%40json/locale.sh)

Do you want this merged now, or do you want to amend it first?

sphh:
Thanks @jaszhix for the hints how to do it.

May I ask you a very big favour? Would it be very complicated and time consuming for you to update the pot files? That would help me quite a lot, because I have to learn all steps and I am currently quite busy with other projects' deadlines approaching too quickly for my liking keeping me busy for the next month or two.

jaszhix:
I've got my hands full for the next couple weeks myself, so can't promise anything right now. There's no rush on this though, approach it whenever you want. 🙂

NikoKrause:
Just run this command in the applets directory:
```
cinnamon-json-makepot --js po/bash-sensors@pkkk.pot 
```
This will update the `.pot` file. There's no need to update the other translations files `.po`. Those files are updated by the translators.

sphh:
Thanks @NikoKrause for your help. That was really easy!!

sphh:
Thanks, @NikoKrause!

PR #1754, opened 2018-04-08
IcingTaskManager@json: Update da.po

Alan01:


xearonet:
@Alan01 
Always add in description something like this "@someuser" because it alert him to review and merge your code. BR

@jaszhix

PR #1760, opened 2018-04-09
sound150@claudiux: Create da.po

Alan01:


claudiux:
Hi @Alan01,

Many thanks for this da.po file!

PR #1764, opened 2018-04-10
vpnLookOut@claudiux: add da.po

Alan01:


claudiux:
Many thanks for this PR, @Alan01 !

PR #1770, opened 2018-04-13
sound150@claudiux: v1.4.1 - Magnetic property of 100% Volume mark bec…

claudiux:
…omes optionnal

Hi,

This is the v1.4.1 of the sound150@claudiux applet.
There is now an option in settings to set/unset the magnetic property of the 100% volume mark, as requested by an user.

The .pot file and the fr.po, es.po, it.po are updated to translate the new messages.

Can you merge this sound150 branch with your master branch, please?

Best regards.

claudiux

### DIFF
--- a/sound150@claudiux/CHANGELOG.md
+++ b/sound150@claudiux/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## CHANGELOG
 
+### v1.4.1
+ * The magnetic property of the 100% mark becomes optionnal.
+ * Translations:
+   * Available languages: English, French, Spanish, Italian, Russian and now **Danish** (thanks to @Alan01!).
+
 ### v1.4.0
  * The 100% mark in sound settings becomes magnetic.
  * The applet and sound settings sliders become synchronized.
--- a/sound150@claudiux/README.md
+++ b/sound150@claudiux/README.md
@@ -79,6 +79,6 @@ The **Sound 150%** applet is designed to allow translations of some messages (in
 
 The installation of the available languages is done automatically.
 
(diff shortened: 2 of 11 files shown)


jaszhix:
Hi claudiux, I noticed a repeating St error when moving the slider after adding it to the panel.

```
(cinnamon:2634): St-CRITICAL **: 20:44:59.654: st_theme_context_get_for_stage: assertion 'CLUTTER_IS_STAGE (stage)' failed
Cinnamon warning: Log level 8: g_object_get: assertion 'G_IS_OBJECT (object)' failed
```

This goes away after Cinnamon is reloaded. I would start looking at any theme introspection calls like `get_theme_node`.

claudiux:
Hi @jaszhix, I noticed this error appeared every time the cursor went over the 100% mark.

claudiux:
@jaszhix, 
I have modified the code of applet.js. No more errors as you reported appears.
(I also removed from the panel the show-desktop applet, which seems to be buggy on Cinnamon 3.8.)

jaszhix:
> (I also removed from the panel the show-desktop applet, which seems to be buggy on Cinnamon 3.8.)

What issue did you run into?

jaszhix:
It looks like you're running Cinnamon git with cjs 3.6/2.4. I guess the update manager overwrote the version you compiled from git. CJS now uses Spidermonkey 52, which adds the class syntax from [ES2015](http://www.ecma-international.org/ecma-262/6.0/).

claudiux:
(I'm using the nightly version of Cinnamon.)

claudiux:
It's OK with the last version of cjs.

Issue #1647, opened 2018-03-03
netusagemonitor@pdcurtis, vnstat@linuxmint.com, and turbonote@iksws.com.br all need to be updated to support new NM library


coderforlife:
Distribution: Fedora 27
Author of `netusagemonitor`: @pdcurtis 
Others are unclaimed

All three of these applets cannot be loaded on Fedora 27 (and possibly future versions of Linux Mint or other distros) due to the move from `NMClient` and `NetworkManager` libraries to `NM`. Attempting to add this applet will cause Cinnamon to continually crash on Fedora 27.

More about the issue is discussed in #1616 which talks about it for multicore-sys-monitor@ccadeptic23 and has a fix in PR #1646. A similar fix would be required for each of these applets as well.

pdcurtis:
As a holding action, I have added warnings about issue #1647 to the README.md  for netusagemonitor@pdcurtis in #1729

I have not yet had time to investigated the changes needed to the code to address the problem of the move from NMClient and NetworkManager libraries to NM in Fedora 27.

 @jaszhix and @coderforlife This may take some time as I am going to be away shortly and without fast internet access for over a month. Assistance from those who have contributed to the changes on other applets would be welcome as I have no way to check under Fedora and I believe blind changes are an unacceptable risk

Do we have any idea if and when Mint will change  to NM and if any other distributions are planing a change - I have found nothing with internet searches?

NikoKrause:
> Do we have any idea if and when Mint will change to NM and if any other distributions are planing a change - I have found nothing with internet searches?

You can follow the discussion here:
https://github.com/linuxmint/Cinnamon/pull/6993

coderforlife:
I believe Ubuntu is also in the process or has already migrated, but there is maybe the possibility that it can actually load both libraries simultaneously meaning no crash, but this is word of mouth.

pdcurtis:
This is a holding action which adds warnings about issue #1647 to the README.md 

The original author of this applet was Clem but I made the most recent modifications in response to a bug which prevented it identifying the active device. The solution used code based on the code in netusagemonitor@pdcurtis so I feel responsible for eventually helping to sorti it out! In fact this is much simpler and shorter code so will be ideal for testing the changes required.

@NikoKrause Thanks for the information - it has enabled trace a lot more of the background. It is more complex and urgent than I realised but I will do what I can.

coderforlife:
[...]

PR #1771, opened 2018-04-14
[WIP] netusagemonitor@pdcurtis Update to version 3.2.5

pdcurtis:
This addresses issue #1647 where certain applet cannot be loaded on Fedora 27 (and possibly future versions of Linux Mint or other distros) due to the move from NMClient and NetworkManager libraries to NM.

 *  The fix uses the code for selecting the network manager for the multicore-sys-monitor@ccadeptic23 thanks to @jaszhix  
  *  The use of xed has been changed to xdg-open to cover more distributions.
  *  A section has been added to README.md covering use on Distributions other than Mint

I have only been able to do a very basic checks on distributions other than Mint so I have initially marked this WIP in case anyone is prepared to check further on Fedora 27 or other Distros using the libnm library before it is merged.

### DIFF
--- a/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/CHANGELOG.md
+++ b/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/CHANGELOG.md
@@ -1,5 +1,12 @@
 ### Changelog covering recent significant changes
 
+## 3.2.5
+
+  * Changes to check which network manager libraries are in use and choose which to use - addresses/solves issue #1647 with Fedora versions 27 and higher.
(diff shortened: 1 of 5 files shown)


pdcurtis:
@jaszhix Many thanks for doing some tests.I have removed the [WIP] and am happy for this to be merged

jaszhix:
I've discovered an issue with the way this checks for the new NM library. It assumes if it can't find the NetworkManager namespace through get_loaded_namespaces that the GI binding isn't installed, but apparently that's not the case on Debian. I tested the fix for that in #1780 on Mint 18.3, 19, Fedora 27, and Debian Buster. I think using try-catch (wrapped in a function called tryFn in multicore-sys-monitor), and falling back to the new lib is the most compatible way to go about this.

cc @coderforlife

PR #1775, opened 2018-04-15
Update betterlock

germanfr:
My led just broke and I installed this applet, which is very useful but it needed a paint job. I did some enhancements which I don't know if the author has to approve or not, because it is set to none.

This is a screenshot of how it looks now.
screenshot from 2018-04-15 16-19-37


NikoKrause:
Looks nice. I would shorten the name even more. Drop the `with notifications`

NikoKrause:
`Num/Caps Lock indicator`

germanfr:
I didn't do that because translations don't have the `MessageTray` part but do have the `with notifications` part. That would drop all of them.

NikoKrause:
Ok

PR #1796, opened 2018-05-04
workspace-grid@hernejj: Fix for Cinnamon 3.8

nathanalderson:
//cc @hernejj. Credit for fix goes to @Lamieur. See hernejj/workspace-grid-cinnamon-applet#22.


hernejj:
Thanks for handling this. :)

PR #1818, opened 2018-05-11
Add Spices notifier

germanfr:


jaszhix:
Nice applet, seems to work well in the little testing I did. 

Only issue I noticed was the icon isn't in the root directory (/spices-notifier@germanfr/icon.png), so it won't show on the website.

germanfr:
Oh just a sec, let me add it.

PR #1828, opened 2018-05-15
CinnVIIStarkMenu@NikoKrause: remove separators custom styles

germanfr:
They do nothing but interfere with theming.

@NikoKrause


NikoKrause:
Thanks!

PR #1829, opened 2018-05-15
CinnVIIStarkMenu@NikoKrause: remove custom padding of favorite apps buttons

germanfr:
This made the buttons' vertical paddings to be different from the horizontal ones.
Let themes define the correct padding.

screenshot from 2018-05-15 15-23-45

@NikoKrause


germanfr:
I created the same pull request for the default menu. This code comes from the applet's initial commit, there is no documented reason for it.

NikoKrause:
Thanks. I'm ok with it.

Issue #1598, opened 2018-01-09
redshift@marvel4u: Brightness level not remembered


sphh:
```
 * Applet version/Build date: 2017-12-05 20:55 UTC
 * Cinnamon version: Cinnamon 3.6.7
 * Distribution: Linux Mint 18.3 Cinnamon 64-bit
 * Snippet or attachment of ~/.cinnamon/glass.log: `info t=2018-01-09T21:26:43.142Z Error getting Cinnamon backlight power settings true,,Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown:`

 ```

**Notify author of applet**
@Marvel4U 

**Issue**
The settings of the brightness slider is not remembered.

**Steps to reproduce**
1. Open the menu of `redshift` with the primary mouse button.
2. Adjust the brightness slider to - say - 50%.
3. Click outside the applet to close the menu.
4. Move the mouse over the `redshift` icon: `Brightness: 50%` is displayed
5. Open the menu again: The brightness slider is again at 100%.

**Expected behaviour**
I would expect to see the brightness slider to show the adjusted value.

**Other information**
Bug #1597 might be related.

sphh:
Thanks @icarter09 for the pull request and @jaszhix for merging it: After updating the applet the level is remembered (but unfortunately not set as described in #1597).

PR #1834, opened 2018-05-17
Redshift: Fixed issue of Brightness level not being remembered.

icarter09:
@Marvel4U This PR is for issue [1598.](https://github.com/linuxmint/cinnamon-spices-applets/issues/1598) I added logic that allows for the Brightness level to be remembered by whatever the user selected. Both the numeric value and the slider remember the setting. Also changed a "var" to "let".

If any of the code looks incorrect or needs to be changed, feel free to change whatever or let me know and I'll make adjusts accordingly.


jaszhix:
@Marvel4U I was able to reproduce the bug, and this PR fixes it.

PR #1840, opened 2018-05-21
[sessionManager@scollins]: Choose which buttons are displayed in the …

claudiux:
…menu.

Hi @collinss,

I propose some modifications to you, so that the user can choose which buttons appear in the menu.

For example, it is convenient not to offer hibernation if it does not work on a laptop used by a child or a person who doesn't know it.

Best regards.

Claudiux

### DIFF
--- a/sessionManager@scollins/files/sessionManager@scollins/applet.js
+++ b/sessionManager@scollins/files/sessionManager@scollins/applet.js
@@ -224,7 +224,7 @@ MyApplet.prototype = {
                         }));
                     }
                 }
-                if ( found ) global.log("Unable to determine display manager");
+                if ( !found ) global.log("Unable to determine display manager");
             }
         }
     },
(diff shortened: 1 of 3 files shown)


collinss:
Ideally, it would be good to automatically remove something if it's not going to work. I have one or two entries that are only supposed to show when the conditions are right. Do you have any idea how to check if the hardware supports hibernate? That's not exactly my area of expertise. As for allowing the user to choose, I'm not sure we need that if we can be a little more intelligent about when to show something or not, though I'm not completely against the idea either. I'm just worried that it might be a little overkill.

claudiux:
Unfortunately no ; the only way to know if something is not working, like hibernation, is to try it.

I tried to display or not some options depending on the key values in dconf, but there is a bug (#7592): these keys are not used by Cinnamon; yet they exist!

claudiux:
Hi @collinss,

I close this PR, which seems useless.

Best regards.
Claudiux

PR #1880, opened 2018-07-07
sound-with-apps-volume@koutch: Add a Keybinding Option for Calling Up the Applet

WammKD:
I prefer mouseless control usage, when feasible, and, particularly for volume, it'd be super useful (especially when most of the applet is already able to be controlled via arrow keys, currently).

@koutch


WammKD:
Hmm; I'm running Cinnamon 3.8(.8), as well, and I didn't run into any problems.

I did notice a dumb mistake on my end as to why it was throwing up an error for the `UUID` variable; I fixed that and added removing the hotkey to `on_applet_removed_from_panel`.

Lemme know if `Gvc` still gives you any issues and I can try swapping it to `Cvc` and see if it still works on my end.

jaszhix:
@WammKD Sorry, not sure how I missed your response - a bit late, but yeah still hitting the Gvc error. Its probably working for you because you have some Gnome components installed.

jaszhix:
Hi @WammKD, this applet has been removed because there's a bug that caused Cinnamon to crash when attempting to open a player from the menu. Thanks for your effort.

PR #1885, opened 2018-07-09
[IcingTaskManager@json] Restore thumbnail preferred size

cardsurf:


jaszhix:
Hi cardsurf, thanks for the PR. I tested this by opening several system settings windows until they started scaling to fit. After closing all but a few, they remained at their scaled size. 

I would prefer not to change the file mode permissions (changed from 0644 -> 0755). 

This scaling issue used to not be a problem a while ago, I'll take a look at it this evening.

jaszhix:
It turns out this does work, I had an issue with my environment because I copied the js folder from cinnamon git, so it was loading as if it were Cinnamon 3.6. 

`fullyRefreshThumbnails` fully destroys all the thumbnails of an app's windows and recreates them. I'm thinking there should be a way to simply resize them, but I think I had trouble with this in the past. Doing a full recycle works, but I think we should find a more optimized solution.

cardsurf:
I have just saw that there is another PR open for this issue. I will close this one to avoid duplicates.

PR #1891, opened 2018-07-13
Old unpopular umaintained applet removal

collinss:
Removes several applets that have a score of 0 and don't have anyone to maintain them anymore.


icarter09:
@collinss Does the score indicate that no one is dowloading them or using them? Was wondering if these could be offloaded to someone that would be willing to maintain the applets. If so, I would be willing to take over these applets.

collinss:
@icarter09 Yeah, a score of 0 usually indicates that no one really cares about them. You never know though, which is why I left it open for a while. Basically, I'm trying to clean up a little bit. Most of these have been abandoned since before we moved everything to github (over a year and a half ago). Some of them haven't had a developer in several years. Combine that with the fact that they all have zero ratings and that seems like enough reason to remove them. If they had a developer or people actually showed interest, I'd leave them.

If there are any that you would like to maintain, you're welcome to, but there are a lot of more popular applets that don't currently have a maintainer, and I'd much rather focus on those. I'd much rather have a smaller number of really good ones that people can easily choose from, than a much larger number of mediocre, half-functioning, or completely broken applets that users have to wade through to find anything good. (Most of them will just give up after a little bit of wading and assume that it's all junk.)

icarter09:
@collinss Thanks for explaining that and completely agree with everything you said. Better to start culling through them now then waiting several months and all these applets that no one is using or maintaining build into a mammoth pile of junk. And better to focus on the ones that are being used. 

I'm not too worried about maintaining them since it looks like no one is really using them. Do you have a list of the ones that you mentioned that are more popular, but do not have a maintainer? I'll happily step up to maintain those applets and put my focus on them.

collinss:
@icarter09 I don't have a list right off hand, but you can look at the info.json in the root folder of each applet (here on github). If it is unmaintained, the author field will be none. Also, I noticed you did some work recently on the command runner applet. If you want, I can leave that one out of my pull request.

icarter09:
@collinss I fixed an issue with the CommandRunner applet that caused it to never start up properly. Someone posted that as an "Issue", but if the numbers say no one is using it then feel free to keep it on the list of removals.

I'll put together a list of all the applets that I can find that have no author. Do you want me to send you that list? Also, how do I go about becoming the maintainer for those applets?

icarter09:
@collinss I have the list put together. Let me know if you want me to post it here or somewhere else.

NikoKrause:
https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/CODEOWNERS

icarter09:
@NikoKrause Thanks. I had no idea that list already existed.

@collinss how do I go about becoming the maintainer of the applets with no author?

collinss:
@icarter09 Just submit a pr setting the author field to your Github user name.

NikoKrause:
@collinss ping

PR #1902, opened 2018-07-18
applet.js

joshbowyer:
This applet was abandoned and was never updated to work with SystemD. I updated and thoroughly tested it and it now works with SystemD, with the only necessary changes being in this js file.


joshbowyer:
Had to redo it since I uploaded the wrong file/revision....

jaszhix:
Please take a look at the readme, it explains the file structure needed for Cinnamon to be able to load it, and for the website to display it correctly. 

I'm not able to test this because it doesn't have a metadata.json file.

joshbowyer:
I'm a little confused. All that needs to be done is replace the current applet.js with the proposed one. I don't really use git, but shouldn't you be able to test by just installing the current spice and then overwriting the local applet.js file in ~/.local/share/cinnamon/applets ?

-------- Original Message --------

jaszhix:
Oh I see, I figured it was removed at some point because it was a file addition in the diff. Yeah, applet.js needs to be modified instead.

Issue #1597, opened 2018-01-09
redshift@marvel4u: Backlight not updated


sphh:
```
 * Applet version/Build date: 2017-12-05 20:55 UTC
 * Cinnamon version: Cinnamon 3.6.7
 * Distribution: Linux Mint 18.3 Cinnamon 64-bit
 * Snippet or attachment of ~/.cinnamon/glass.log: `info t=2018-01-09T21:26:43.142Z Error getting Cinnamon backlight power settings true,,Error: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown:`

 ```

**Notify author of applet**
@Marvel4U 

**Issue**
When adjusting the backlight with the slider provider, the backlight does not change.

**Steps to reproduce**
1. Open the menu of `redshift` with the primary mouse button.
2. Adjust the brightness slider to - say - 50%.

**Expected behaviour**
I would expect that the brightness of the screen is adjusted.

**Other information**
If I adjust the brightness level with the stock Power Manager applet (power@cinnamon.org), the backlight does get updated.

icarter09:
Cinnamon 3.8.8
Linux Mint 19 64bit
Applet version: 2018-07-15 20:03 UTC

Confirmed that this issue still exists. 

@sphh I'll start looking into this issue. Pretty sure it is closely related to issue #1598.

icarter09:
@sphh The PR containing this fix was merged. Can you verify if this fixed the problem?

sphh:
Thanks @icarter09!

**What is working:**

- I can change the brightness in the redshift applet and the backlight of the screen gets updated.
- When I change the brightness in the redshift applet, the slider in the Cinnamon power manager applet (power@cinnamon.org) gets updated, so does the OSD.

**What is not working:**

- I can change the backlight brightness with the help of the Cinnamon power manager applet or with dedicated keys on my keyboard. But the tooltip menu of the redshift applet does not show the updated value. What's more important, when I open the menu of the redshift applet, the brightness gets reset to the value which was set with the redshift slider.

Is it possible to syncronize the brightness value both ways?

icarter09:
@sphh I noticed the "what is not working" issue as well when I was debugging the PR for the initial problem. One issue I was running into with trying to apply that fix is that the system brightness would keep changing the value, regardless of what I initial set the redshift applet slider value. Best guess is that it had to be related to the hardware. Which in turn would keep readjusting the redshift slider and not remembering the value I had set it to.

[...]

PR #1919, opened 2018-08-01
redshift: Backlight not updating fix.

icarter09:
@Marvel4U: This PR is s fix for issue #1597. Feel free to change/edit however you see fit or just let me know and I'll fix it.

### DIFF
--- a/redshift@marvel4u/files/redshift@marvel4u/applet.js
+++ b/redshift@marvel4u/files/redshift@marvel4u/applet.js
@@ -175,7 +175,7 @@ MyApplet.prototype = {
 
     getSystemBrightness: function () {
         try {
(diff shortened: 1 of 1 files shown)


jaszhix:
Could be my hardware, not testing this on a laptop, but when I adjust the brightness slider I get

```Error getting Cinnamon backlight power settings true,,Error: GDBus.Error:org.gtk.GDBus.UnmappedGError.Quark._csd_5fpower_5fmanager_5ferror.Code0:```

icarter09:
@jaszhix is there anyway you can test it out on a laptop? I think I remember noticing that as well on my desktop and I'll double check when I get a moment. I'll also double check it against another laptop I have that I wasn't doing the dev work on. I'm curious on if it is the hardware or not. Thanks for the heads up.

jaszhix:
I tested on my optimus laptop and no issue there, so looks good to me.

Issue #1422, opened 2017-10-29
Idea: Remove label "Mem:" in mem-monitor-text@datanom.net


bl0ckz:
Hey,
Is it possible that you offer an option to remove the label "Mem:" in the mem-monitor-text?
Best,
Christian

PR #1920, opened 2018-08-04
Mem-monitor-text: Removed label.

icarter09:
This PR is a fix for issue #1422.


brownsr:
I don't think that is what the requestor asked for - he asked for an option to remove, not full removal.  However in the context of a multi-lingual project, having untranslated embedded English acronyms is not a great idea, so am happy to merge.  

Regards

Simon

icarter09:
@brownsr my bad. Attempted to work on another ticket before I went to bed, which in hindsight was not a good idea. I'll start looking into implementing what the requestor asked for.

Issue #1913, opened 2018-07-27
batterymonitor@pdcurtis - BAMS applet error with LM 19.0


CJConline:
```
 * Applet version/Build date 1.3.2
 * Cinnamon version  3.8.8
 * Distribution: Mint 19.0
 * Graphics Intel HD Graphic 520 (Dell Inspiron 15 5000 series laptop)
  

<!-- Author list: https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/CODEOWNERS -->

**Notify author of applet**
@pdcurtis

**Issue**
When installing the BAMS applet and also when you login after you first boot up, a screen pops up that shows a red minus sign and the error "Some Dependencies not Installed".
Error did not happen with LM18.3.
Also the percentage charge displayed doesn't update after resuming from suspend if the charge has dropped below the critical (red) level even after you plug the power in and even when the battery gets to 100%, it still remains stuck incorrectly at the critical level it was originally on.
See the attached screenshot.
![bams error screenshot](https://user-images.githubusercontent.com/27131265/43318634-12803952-91d4-11e8-9cdb-c5a72fb876e4.png)



**Steps to reproduce**



**Expected behaviour**



**Other information**

icarter09:
@CJConline the reason for the install message is that since version 1.3.2, both sox and zenity are needed for the battery monitor applet to function correctly. I got the message warning message that you stated, then installed sox (zenity was already at the latest version). After install, I removed and added the battery monitor applet back and the warning message no longer appeared. 

Let me know if installing both sox and zenity fixi the warning issue.

CJConline:
Thank you so much.  Works well now.  Will check the other issue I had with
it not updating percentage charge once it had fallen to critical when
suspended when time permits.  Chris

PS: Would be good it the need to install sox and zenity was mentioned in
the description on the Cinnamon Spices pages for less experienced user like
me.  NB:  zenity was already installed on LM 19.0 Tara (upgraded from the
beta) but not sox.   Both sox and zenity must have already been install
with LM18.3.  Cheers.

icarter09:
@CJConline Let me know if the other issue is still present. I'll try to put in a PR tonight with an updated message informing the user of how to correct the dependency problem.

pdcurtis:
I have only just found this - neither of you put the @pdcurtis in you comments so it did not reach me until #1931 alerted me!
@icarter09 I have agreed the PR you put forwards.
[...]

PR #1931, opened 2018-08-22
batterymonitor@pdcurtis: Updated warning message.

icarter09:
@pdcurtis This PR is for issue #1913. Feel free to change however you see fit or let me know and I'll make the changes accordingly.


pdcurtis:
@icarter09 Thanks for the suggestion. I am quite happy for this to be merged.
The associated comment made in the original issue does not need fixing as the requirement is already in the description on the spices page and the built in help file under Requirements - perhaps it should be higher up but it is there so please close the issue when the merge has taken place. Thanks.

Issue #53, opened 2017-02-01
weather@mockturtl: Problem loading - check looking glass


skullscream:
I get this issue when I try installing this applet via "applets" menu in system settings. This is the looking glass log:

```
info t=2017-02-01T10:36:32.678Z Cinnamon.AppSystem.get_default() started in 80 ms
info t=2017-02-01T10:36:32.685Z loading default theme
info t=2017-02-01T10:36:32.832Z PlacesManager: Updating devices
info t=2017-02-01T10:36:32.919Z loaded at Wed Feb 01 2017 11:36:32 GMT+0100 (CET)
info t=2017-02-01T10:36:33.548Z Loaded applet menu@cinnamon.org in 622 ms
info t=2017-02-01T10:36:33.831Z Role locked: panellauncher
info t=2017-02-01T10:36:33.832Z Loaded applet panel-launchers@cinnamon.org in 283 ms
info t=2017-02-01T10:36:33.900Z Loaded applet window-list@cinnamon.org in 66 ms
info t=2017-02-01T10:36:33.932Z Loaded applet removable-drives@cinnamon.org in 29 ms
info t=2017-02-01T10:36:33.985Z Loaded applet keyboard@cinnamon.org in 49 ms
info t=2017-02-01T10:36:33.995Z Loaded applet systray@cinnamon.org in 9 ms
info t=2017-02-01T10:36:34.110Z Loaded applet power@cinnamon.org in 112 ms
info t=2017-02-01T10:36:34.297Z Loaded applet network@cinnamon.org in 184 ms
info t=2017-02-01T10:36:34.470Z Loaded applet sound@cinnamon.org in 169 ms
info t=2017-02-01T10:36:34.549Z Loaded applet calendar@cinnamon.org in 78 ms
info t=2017-02-01T10:36:34.580Z Loaded applet settings@cinnamon.org in 29 ms
info t=2017-02-01T10:36:34.594Z 2 applet already loaded
info t=2017-02-01T10:36:34.595Z Loaded applet weather@mockturtl in 14 ms
info t=2017-02-01T10:36:34.596Z AppletManager.init() started in 1676 ms
info t=2017-02-01T10:36:34.601Z Cinnamon took 2054 ms to start
info t=2017-02-01T10:36:35.916Z Adding systray: parcellite (24x24px)
info t=2017-02-01T10:36:35.921Z Adding systray: pamac-tray (24x24px)
[...]

PR #1948, opened 2018-09-04
weather@mockturtl: Fix up #53

### LINKED ISSUE #1947: weather@mockturtl: Can't get weather info behind HTTP Proxy
```
 * Applet version: 1.14.0
 * Cinnamon version: Cinnamon 3.2.7
 * Distribution: Debian GNU/Linux 9.5 (stretch)
 ```
**Notify author of applet**
@mockturtl

**Issue**

When using a http proxy to access Internet servers, the weather applet does not fetch data from remote endpoint (QUERY_URL).

**Steps to reproduce**

- Filter internet access, allowing only access via proxy server.
- Set proxy info via control panel and/or http_proxy & https_proxy environment variables.
- Add weather@mockturtl applet to the cinnamon panel.
- It shows "Loading ..." and never shows weather info.

**Expected behaviour**

- Sunny or at least cloudy ~24º C  (I hope, or at least any other forecast)

**Other information**

- Seems introduced in baa54f976ddb1c2ee683e2dcfdc63fc553f456e4
- Works fine when [line 182](https://github.com/linuxmint/cinnamon-spices-applets/blame/master/weather%40mockturtl/files/weather%40mockturtl/applet.js#L182) is uncommented

jaszhix:
@mockturtl 

Closes #1947

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/applet.js
@@ -179,7 +179,7 @@ const WEATHER_CONV_ATM_IN_INHG = 33.421054e-3
 
 // Soup session (see https://bugzilla.gnome.org/show_bug.cgi?id=661323#c64)
 const _httpSession = new Soup.SessionAsync()
-//Soup.Session.prototype.add_feature.call(_httpSession, new Soup.ProxyResolverDefault())
+Soup.Session.prototype.add_feature.call(_httpSession, new Soup.ProxyResolverDefault())
 
 
 //----------------------------------------------------------------------


mockturtl:
Just to be clear, I haven't tested the change; I don't use a proxy.

jaszhix:
I don't either, still works without though. I don't remember why I commented that line out - it was the TypeError from `json.query.results` being falsey too early causing issues, so its probably a safe change if others can confirm.

maurom:
Just cloned master and tested the applet behind proxy. Works ok.
Thanks for such a prompt response!

PR #1960, opened 2018-09-12
temperature@fevimu: Resolved 'NaN' applet value when 'sensors' command returns no Package or Core temperature but shows 'CPU Temperature'

mi2ek:
NaN was caused by division by zero while trying to calculate mean temp for all Packages/Cores. On my computer lm-sensors returns only 'CPU Temperature', no Package/Core temperatures.


jaszhix:
Looks good, thanks.

PR #1989, opened 2018-10-04
hwmonitor@sylfurd: Misc. clean up

jaszhix:
- Ensures the `Mainloop.timeout_add` callback returns `true` instead of recursively
calling `Mainloop.timeout_add`, and returns `false` (stops interval) when
removed from the panel.
- Caches the actor on the GPU.
- Reduces redundant calculations and function invocation in
Graph.prototype.paint.
- Minor style fixes for more consistency.
- Removes unnecessary try-catch blocks.

Ref https://github.com/linuxmint/Cinnamon/issues/7844

### DIFF
--- a/hwmonitor@sylfurd/files/hwmonitor@sylfurd/applet.js
+++ b/hwmonitor@sylfurd/files/hwmonitor@sylfurd/applet.js
@@ -262,47 +262,47 @@ function CpuDataProvider() {
 
 CpuDataProvider.prototype = {
 
-	_init: function(){
-		this.gtop = new GTop.glibtop_cpu();
-		this.current = 0;
-		this.last = 0;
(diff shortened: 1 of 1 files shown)


cschreib:
I have just updated the applet to benefit from your changes and see if this helps with linuxmint/Cinnamon#7844. Unfortunately the applet appears now to be broken, as the CPU indicator always stays close to zero. I can see that it is trying to draw the CPU usage curve, but it's like it can only be either 0 or 1%: nothing is drawn when CPU is inactive, as one would expect, but when CPU is active (regardless of if it's just one or all cores at once), only a thin blue line is drawn on the bottom-most pixel row. You can see if on the screenshot below:

cpu

cschreib:
That did it, thanks!

PR #2031, opened 2018-10-31
vpnLookOut@claudiux: Update sv.po

eson57:
Update Swedish translation


claudiux:
Many thanks, @eson57 !

PR #2043, opened 2018-11-04
CinnVIIStarkMenu@NikoKrause: add tr.po

kelebek333:
CinnVIIStarkMenu@NikoKrause: add tr.po


kelebek333:
reviewed, no problem.

PR #2045, opened 2018-11-04
placesCenter@scollins: add tr.po

kelebek333:
placesCenter@scollins: add tr.po


kelebek333:
reviewed, no problem.

PR #2046, opened 2018-11-05
multicore-sys-monitor@ccadeptic23: add tr.po

kelebek333:
multicore-sys-monitor@ccadeptic23: add tr.po


kelebek333:
reviewed, no problem.

PR #2069, opened 2018-11-10
vpnLookOut@claudiux: Update sv.po

eson57:
New strings translated and some aprovements


claudiux:
Thank you, @eson57 and @jaszhix !

PR #2078, opened 2018-11-14
vpnLookOut@claudiux v3.2.1 - Improves display and management of Activity Log

claudiux:
Hi,

Can you merge this branch with your master one, please?

Best regards.

Claudiux

### DIFF
--- a/vpnLookOut@claudiux/CHANGELOG.md
+++ b/vpnLookOut@claudiux/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 3.2.1~20181115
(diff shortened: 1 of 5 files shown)


claudiux:
Thank you, @jaszhix !

Can you tell me if there is a signal sent to applets when Cinnamon stops or restarts? I would like this applet to write the time of its shutdown in its log file. Is it possible?

jaszhix:
Hi @claudiux, I don't think there is a signal for that currently.

claudiux:
I did not find either, @jaszhix. Thank you for your answer.

PR #2091, opened 2018-11-19
sound150@claudiux: add sv.po

eson57:
Add Swedish translation


claudiux:
Thank you, @eson57!

eson57:
No problem. Thanks for the applet.  :)

PR #2096, opened 2018-11-21
sshlauncher@sumo: Add ability to have Server Groups in SSH Launcher.  

NathanaelA:
Using `#GroupStart MyName` will create a group called "MyName".   `#GroupEnd` will close the group.  Used `#` since it is a comment character; so it won't have any negative effects on anything else that read SSH Config files.

image

So in this example image; `airey` and `story` are servers that are grouped inside the `Home` group.
Debating adding some spaces in front of the sub-menu-items...

### DIFF
--- a/sshlauncher@sumo/files/sshlauncher@sumo/applet.js
+++ b/sshlauncher@sumo/files/sshlauncher@sumo/applet.js
@@ -65,16 +65,36 @@ MyApplet.prototype = {
     this.menu.addMenuItem(new PopupMenu.PopupSeparatorMenuItem());
 
     try {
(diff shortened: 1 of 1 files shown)


NathanaelA:
@sumo - I believe you need to review this PR.

sumo:
Thanks, I'll take a look and do some testing.

NathanaelA:
Please let me know if you think it needs any more work (i.e. like spaces in front).   I use this all the time (have about 20 servers in my list in several groups and haven't even added all of them  :grinning: ).   

This won't break any existing functionality (so those who don't use grouping won't see any changes).

PR #2100, opened 2018-11-21
vpnLookOut@claudiux: Update da.po

Alan01:


claudiux:
Many thanks, @Alan01 !

PR #2109, opened 2018-11-25
vpnLookOut@claudiux: update tr.po

kelebek333:
vpnLookOut@claudiux: update tr.po


claudiux:
Thank you, @kelebek333 !

PR #2114, opened 2018-11-25
netusagemonitor@pdcurtis Update to version 3.2.6 (multiversion to support Cinnamon 4.0) 

pdcurtis:
* Changes for Cinnamon 4.0 and higher to avoid segfaults when old Network Manager Library is no longer available by using multiversion with folder 4.0
  * Remove Try-Catch as no longer required in 4.0 and associated changes.
  * It is believed that all Distributions packaging Cinnamon 4.0 have changed to the new Network Manager Libraries
  * Add cinnamon-version to metadata.json (Provides information on which Cinnamon versions can load it) 

This is in response to https://github.com/linuxmint/Cinnamon/pull/8050#issue

@jaszhix I have been unable to test this under Cinnamon 4.0 but it follows closely what you have found works for the applets you have already modified. Can you please check it before it is merged or leave the WIP until I can check myself when the beta becomes available.


pdcurtis:
@jaszhix Thanks for the quick response but I am once more confused. That does not seem to be what the only actual  documentation I could find at http://developer.linuxmint.com/reference/git/cinnamon-tutorials/xlet-versioning.html says about where earlier versions should live. To quote:

> To enable extension versioning, extensions have to add the line
> 
> {
>       "multiversion": true
> }
> 
> in metadata.json. The different versions should be put in different subdirectories. For example, the version for 2.6 should be put in `extension@uuid/2.6/`. The contents in 2.6/ should be exactly what you will normally put in `extension@uuid/`, apart from metadata.json, which should always be put in the parent directory, ie. `extension@uuid/`.
> 
> You do not need to create a subdirectory for every single Cinnamon version out there. Cinnamon looks for the most recent subdirectory that is not newer than the running Cinnamon version. For example, if you are running 2.6 and there are 2.4, 2.8 directories, the 2.4 directory will be loaded, until the you upgrade to 2.8 in which case the 2.8 directory will be used. Minor version numbers can also be used, eg. 2.6.4, and are sorted accordingly.
> 
> If no suitable directory is found, then the contents in `extension@uuid/` will be loaded. Note that Cinnamon versions prior to 2.6 will not understand this directory magic, and will always try to load the contents in `extension@uuid/`. **Hence it is suggested that extension maintainers put the version for 2.4 in extension@uuid, and create new directories if changes specific to newer Cinnamon versions are made.** Don’t make a new directory whenever a Cinnamon version is out since it is just a waste of space (and maintenance effort).
> 

I also looked at several other applets before I started. The system seems to be very tolerant and one only seemed to have the changed files in the x.y folder but most seemed to follow the above. Your applets however work with nothing in `extension@uuid/`

The applets also work fine on my Mint 19 system so Cinnamon is not ignoring them with `multiversion`
 enabled in metadata.json and are supported back to 2.2 which came with LTS version 17 which is current until some time in 2019.

Do you have some more up to date documentation that I have not found?

pdcurtis:
@jaszhix Another question.
I have noticed in https://github.com/linuxmint/Cinnamon/wiki/%5Bdevelopment%5D-extensions that metadata.json has:

```
    requiredProperties: ['uuid', 'name', 'description', 'cinnamon-version'],
    niceToHaveProperties: ['url'],
```
Is this the case as it means every applet would need to be updated every time the cinnamon version is changed?
It also means I need to add it to all my applets!

jaszhix:
Most of that documentation and examples of other applets you're seeing were from before I refactored the extension system. Mint 19 has the new bindings so you wouldn't be able to tell if the legacy version or the 4.0 version is working, did you try logging to see if the file you thought was loaded was loaded?

pdcurtis:
As far as I can tell the files that are being used are what I expect, for example the help and log files have the correct contents. There is a BUT as I am not loading as such but have a symbolic links to the git repository so any changes automatically update for testing as I change branches eg by:

`ln -s ~/cinnamon-spices-applets/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis ~/.local/share/cinnamon/applets/netusagemonitor@pdcurtis`

I will do some further tests and experiments in another branch as you suggest including adding some test code outputting to the system log files but I am pretty sure that I am running the code I expect under Mint 19 and you found it worked under Cinnamon 4.0

I can however see problems coming if even those at the heart of Cinnamon need to run experiments and when we have 148 applets which may need to be checked and potentially updated even in a trivial manner for every version - that's potentially a lot of PRs.

Would you like me to also take on netusage@30yavash.com - I have already done quite a bit of work on it in the past and keep it on my installed on my desktop and I obviously have a fork of the desklet repository.  I can make the same changes as the two I am already doing - when I know what they should be!

pdcurtis:
I will take on netusage@30yavash.com and test it out and fix the problem with filling error logs when no connection is present. I will also use it as a test bed for some of the other points you have raised before we merge them all.

As a first input I have done a search with grep and there are only 13 of 148 applets which use `'cinnamon-version'` in `metadata.json` and quite a few already have `"4.0"` included. There are also several which do not get to `"3.8"` so that updating is a much smaller task than I feared. It is also clear that a `'cinnamon-version'` is not obligatory.

pdcurtis:
@jaszhix  Pushed changes as above. It is currently 3 commits - last two could be combined but the initial two are best separate so diff of changes in  4.0 from earlier version is more clear.

I see no reason why it can not be merged so I will taken the WIP off 

Thank you for the enormous amount of effort you have been putting into this latest release. Please and let me know if there is anything I can do to help.

PR #2115, opened 2018-11-25
 vnstat@linuxmint.com Update to version 1.0.1 (multiversion  to support Cinnamon 4.0) 

pdcurtis:
* Changes for Cinnamon 4.0 and higher to avoid segfaults when old Network Manager Library is no longer available by using multiversion with folder 4.0
  * Remove Try-Catch as no longer required in 4.0 and associated changes.
  * It is believed that all Distributions packaging Cinnamon 4.0 have changed to the new Network Manager Libraries 
  * Added cinnamon-version information to metadata.json and some test code to applet which is now commented out in update to 1.0.1.2

This is in response to https://github.com/linuxmint/Cinnamon/pull/8050#issue

@jaszhix I have been unable to test this under Cinnamon 4.0 but it follows closely what you have found works for the applets you have already modified. Can you please check it before it is merged or leave the WIP until I can check myself when the beta becomes available.


pdcurtis:
Following the long discussion in #2214 with @jaszhix I have made pushed some addition changes. 

It is currently 3 commits - last two could be combined but the initial two are best separate so diff of changes in 4.0 from earlier version is more clear.

I see no reason why it can not be merged so I have taken the WIP off

PR #2119, opened 2018-11-27
feeds@jonbrettdev: Resolve scope issues, add fallback RSS reader

Centurix:
- CJS variable scoping so the app starts correctly
- Added a fallback simple RSS reader when feedparser isn't present in the Python path


Centurix:
@jake1164

Centurix:
Thanks, maybe hold off until I appease Codacy. It doesn't like Python+xml. I do use this applet a bit for feeds and felt that it should work out of the box without needing the feedparser Python dep.

Centurix:
Closing this PR. I have a tidier solution to submit.

PR #2120, opened 2018-11-27
github-projects@morgan-design.com: Fix variable scoping for CJS

Centurix:
- Variable scope was stopping the repos from listing correctly

@jamesmorgan


jamesmorgan:
👍 thanks for contributing - I dont have the time I used to anymore

PR #2125, opened 2018-12-01
Autostart programs Hungarian translation

KAMI911:


jaszhix:
Hi @KAMI911, this looks like the same as #2126. Not sure if you meant to, but PRs are supposed to change one applet at a time.

PR #2146, opened 2018-12-10
IcingTaskManager@json: Add migration functionality for moving pinned apps from ITM to GWL

jaszhix:
* Added a deprecation notice notification for Cinnamon 4.0 users.
  * Added migration functionality for moving local pinned apps from ITM to Grouped Window List in Cinnamon 4.0. This can be accessed by clicking "Migrate Pinned Apps to Grouped Window List" from ITM's settings.

### DIFF
--- a/IcingTaskManager@json/CHANGELOG.md
+++ b/IcingTaskManager@json/CHANGELOG.md
@@ -1,5 +1,10 @@
 Changelog
 
+### 6.3.14
+
+  * Added a deprecation notice notification for Cinnamon 4.0 users.
+  * Added migration functionality for moving local pinned apps from ITM to Grouped Window List in Cinnamon 4.0. This can be accessed by clicking "Migrate Pinned Apps to Grouped Window List" from ITM's settings.
+
 ### 6.3.11
 
   * Fixed an issue with the applet not initializing when an open app can't be resolved by wm_class.
--- a/IcingTaskManager@json/README.md
+++ b/IcingTaskManager@json/README.md
@@ -1,6 +1,8 @@
 Icing Task Manager
 =============
 
+**Icing Task Manager is now deprecated for all Cinnamon 4.0 users. Please add the Grouped Window List applet to your panel, go to ITM's settings, and click "Migrate Pinned Apps to Grouped Window List".**
+
 **Issues should be opened on ITM's [dedicated repository](https://github.com/jaszhix/icingtaskmanager). All PRs should be submitted to the Cinnamon Spices repository.**
 
 This is a fork of the unfinished development branch of [Window List With App Grouping](https://github.com/jake-phy/WindowIconList/) applet, originally by jake-phy who forked the code from [GNOME Shell Window List](https://github.com/siefkenj/gnome-shell-windowlist/).
--- a/IcingTaskManager@json/README.md
+++ b/IcingTaskManager@json/README.md
@@ -1,7 +1,7 @@
 Icing Task Manager
 =============
 
-**Icing Task Manager is now deprecated for all Cinnamon 4.0 users. Please add the Grouped Window List applet to your panel, go to ITM's settings, and click "Migrate Pinned Apps to Grouped Window List".**
+**Icing Task Manager is deprecated for Cinnamon 4.0+. Please add the Grouped Window List applet to your panel, go to ITM's settings, and click "Migrate Pinned Apps to Grouped Window List".**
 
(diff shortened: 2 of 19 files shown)


pdcurtis:
@jaszhix I am sure this is not the correct place but from what I have found so far your IcingTaskManager is currently much better than the GWL in Cinnamon 4.03 which is missing some of the functionality and customization I use. I immediately switched back and hope you will continue to support what I have always felt has been the most important applet and the first one I install, even ahead of my own!

jaszhix:
@pdcurtis What features in particular do you want to see in GWL?

pdcurtis:
@jaszhix What I noticed immediately was that the identification of what windows were active and how many duplicate windows were in use was different. I think what I use was set by Number Display where I use setting normal which only puts a superscript number on windows that are in use but shows how many whilst highlighting (depending on theme) shows the window in focus. Currently I have 13 different programs running some with up to 4 instances and the ITM makes it very easy. 

Overall it looks as if quite a few options were dropped. 

Also the icons seem to be much smaller and have odd bars on the bottom on several themes and no indication which window is in focus. It may be I am running with an early copy. 

Some of the options may be confusing to a new user but perhaps they could be hidden or on an advanced tab - I am sure you added them all for very good reasons.

jaszhix:
Which theme are you using? GWL relies on new class names so it doesn't conflict with other styles, and not all themes are compatible yet. This was done so the theme options could be dropped because they were buggy in some themes and hard to support. I would still get rid of them if I were to  continue ITM development - this is a transitional issue for theme maintainers.

pdcurtis:
Mint-Y, Mint-Y Dark, Linux Mint, Cinnamon, New Minty all seem odd. Mint-X looks better but icons still tiny. I am currently using Minty, Linux Mint or New MInty on my other machines but New Minty  has several  problems with text colours and backgounds. My wife still uses Void. 
I have done an update and there seems to be some improvements in GWL / Themes interaction.

jaszhix:
Have you tried adjusting the icon size in Panel Settings? That's why that option was removed - it was made global per-panel zone in Cinnamon 4.0.

pdcurtis:
No. I had GWL in left panel and ITM in right so I could compare - the defaults are quite different. I was only testing my applets on a LiveUSB so I had not looked at all the settings. 

I can now see on some of the themes what is supposed to happen but I still think that options on the way the superscript numbers are displayed is useful as it gives a very good way to separate between a pinned, in use and multiple in use programs which is consistent between themes. 

Where do I find any the reference information for the new bits in Cinnamon: themes, settings etc? Finding the latest documentation is still a problem for me.

jaszhix:
https://www.linuxmint.com/rel_tessa_cinnamon_whatsnew.php

pdcurtis:
@jaszhix  That does not give sufficient for designing or updating applets. I often end up examining your code (and others in the team) to find out the changes or how to do things. For example the only information on vertical panels I have found is in the original PR. Tabbed Settings is only in the applet example.  @NikoKrause wrote a nice explanation of localization somewhere on git which I copied onto my site. I usually get there but it must be a nightmare for anybody starting to write an applet. Sorry about the rant - I spent years hammering my staff to document. 

Enough. What I did not say above was what an excellent idea it was to build ITM into Cinnamon and I have found the original discussions which is where I should have commented.

jaszhix:
I thought you were asking about general release info since you seemed confused about 4.0 features. I suggest you look at the Cinnamon source code if you want to learn more. Install devdocs, make sure the cinnamon-doc package is installed. All the JS comments turn into API docs, and viewable through the app. We can't always document everything, and you should definitely read the parts of Cinnamon's source you're interested in, its like flying blind otherwise.

pdcurtis:
@jaszhix  That is very useful, I had found a lot of the material which is in devdocs/cinnamon-doc but not all in one place and not the information on what is depreciated. I will need to check some of my `settings-schema.json` files.

NikoKrause:
About the icons size. In Cinnamon 4.0 rightclick on the panel -> Panel settings. There you can set your preferred icon size for the left/middle/right zone of the panel. Mine looks like this:
screenshot-cinnamon-2018-12-14-201057

Firefox is focused with 2 windows open.
Those with a teal/blue line at the bottom are running apps (Writer, Nemo, Inkscape)
Terminal is pinned but has no running window

pdcurtis:
@NikoKrause Thanks, I have now got icon sizes correct - one problem was that I was running GWL in right panel to compare to ITM so it had defaulted to 24pixel. Other problem was that themes I use did not display GWL styles correctly. I still prefer what used to be Normal to the Smart that is now used but as long at one can see if a program is running it is OK and new users need limited options. In time I guess people will update their themes to have the extra styles.

PR #2152, opened 2018-12-19
netusagemonitor@pdcurtis - update stylesheet.css to look like other applets

Mintulix:
removed special font family -> takes default font now
removed green color for normal non-alert conditions -> takes default color now
removed opacity change when connected -> default opacity now
removed border radius to look more modern

### DIFF
--- a/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/4.0/stylesheet.css
+++ b/netusagemonitor@pdcurtis/files/netusagemonitor@pdcurtis/4.0/stylesheet.css
@@ -7,10 +7,9 @@ or to allow a reduced size width to save panel space.
 .numa-left {
 
 	text-align: center;
-	font-family: "Ubuntu Mono", Arial;
 	font-weight: bold;
 /*
-
+	font-family: "Ubuntu Mono", Arial;
 	min-width: 75px;
 	font-size: 95%;
 */
@@ -18,22 +17,20 @@ or to allow a reduced size width to save panel space.
 .numa-right {
 
 	text-align: center;
-	font-family: "Ubuntu Mono", Arial;
 	font-weight: bold;
-
 /*
+	font-family: "Ubuntu Mono", Arial;
 	min-width: 1px;
-    font-size: 95%;
+    	font-size: 95%;
 */
 }
 .numa-compact {
 	text-align: center;
-	font-family: "Ubuntu Mono", Arial;
 	font-weight: bold;
 /*
-
+	font-family: "Ubuntu Mono", Arial;
 	min-width: 1px;
-    font-size: 95%;
+    	font-size: 95%;
 */
 }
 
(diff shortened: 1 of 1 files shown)


Mintulix:
This stylesheet.css was identical to the stylesheet.css files in folder 3.2 and in the base folder.
I didn't know if I should change them all.

jaszhix:
Make sure you ping the maintainer @pdcurtis.

Mintulix:
Hey @pdcurtis,  what do you think about the changes? Would you be so kind to test them?
Regards.

pdcurtis:
I will certainly have a look but it will be a few days.

There was a logic to avoid the defaults and it would also affect some of my other applets so changes need a lot of consideration

That is why I gave the option for users to edit the style sheet under advanced options. The problem is that they are overwritten by updates and I will more likely look at a way to give persistence to edits.

@jaszhix Can you put a tag such as discussion or WIP on this and I would be interested if there is any policy on styling of applets

Mintulix:
This is exactly the reason why I wanted to make this change. I prefer darker themes with slight transparency, but this green is a bit prominent... And after each update (which you do very often, thanks) my changes are gone.
Persistence to edits is a good idea, but I don't know if that fits in the applet update philosophy.

pdcurtis:
I do not see there can be any objection to persistence if it is an option to use a different .css or audible warning file and I recall that settings offers a file selector. Many of the updates are translations but I have been making some other changes so just not updating is not a good option at present. 
Re light and dark themes the transparency was an compromise to give acceptable results on both. 
I will see what i can do but i only have a pad without git for the next 10 days.

Mintulix:
No need to rush, it's Christmas. And it's a matter of personal taste, not a bug.
Also, when I looked at some themes, it seems to be a good idea to drop the font-weight as well. So take the merge as suggestion.
Happy holidays!

pdcurtis:
@Mintulix you have obviously taken quite a lot of trouble including forking the repository and using git. If you are interested in working with applets the best source of information is to load `cinnamon-doc` which should pull in `devhelp` which will put Devhelp in the menu and give access to the latest (but not perfect!) documentation which will explain why there is an applet.js etc in subfolders like 3.2 and 4.0 for different cinnamon versions. There is also a Settings API Demo applet which shows how the settings (configuration) works. 
There is information, (a bit behind) on my applets on www.pcurtis.com/spices.htm and my personal use of git and Github  for applets at www.pcurtis.com/git.htm

pdcurtis:
@jaszhix I am sorry to seek advice yet again but I have been unable to find any way to get a style from a sheet other than the default stylesheet.css in the applet folder. 

I have done searches through every applet in cinnamon-spices-applets and looked in Devhelp and have found nothing helpful. 

What am I missing -  there must be a way to set a path to a different stylesheet?

pdcurtis:
@Mintulix I have a test update running which provides persistence but it lacks elegance.
I have looked at your suggestions in detail and tried them and various other versions with many themes.
The only major change which I would make when using them is to add some identification of when the interface is active. It does not have to be a background colour but could follow the way themes identify 'activity' - I rather like an underline border-bottom of a few pixels with the  Mint-Y themes. I also like a transparent top and bottom border when a background colour is in use when the corners are square.

Mintulix:
@pdcurtis Thanks for investigating. Concerning underline border - some themes like Adara-Dark already have an underline effect when you hover your mouse pointer. I think interface is active is the default condition. What about to mark only the inactive state?

pdcurtis:
@Mintulix  Reasoning is consistency with my other monitoring applets which all use a 'traffic light' type of model.
Green good/running, Orange warning/alert and Red danger/stop. I have actually got a close model of the system Themes which have slightly rounded corners and use background or edge colours. I am generating a PR with my changes included whilst adding a simple persistence as I do not believe it is possible to get a solution which matches every theme and keeps everyone happy. I have acknowledged your input.

PR #2165, opened 2018-12-29
SpicesUpdater@claudiux - Checks and automatically updates your Spices

claudiux:
Hi,

This is a new applet.

It can check for updates about the Spices installed by user in his home directory (not about spices "...cinnamon").

It can also automatically install these updates, if that is the choice of the user. Otherwise, some notifications appear, the color of the icon changes and, in the menu of this applet, a dot before any type of the Spices to update warns the user; a click on a Spices type (Applets, Desklets, etc) opens the corresponding page in Cinnamon Settings.

There are five tabs in the settings of this applet.

The first, _General_, allows the user to:
  * Select the _Time interval between two checks_ (in hours). Please note that the first check will take place 2 minutes after starting this applet.
  * Select the multiple ways to warn you : by changing the appearance of the icon of this applet and/or by messages in the notification zone.
  * Make choices about events logging.
  * Select the _Type of display_ of the icon: with or without text?

For the content of the other tabs (_Applets_, _Desklets_, etc), please look  at the screenshot.png.

The list of installed Spices is automatically filled.

The user will set to _FALSE_ all the Spices he _do not want_ to automatically update, **before** checking the _Update Applets_ box (or the _Update Desklets_ box, etc). This is useful for developers.

This applet works very well on Cinnamon 3.8 and 4.0, tested on Linux Mint 19.1, Fedora 27 and Antergos (Arch).

Happy Holidays!


claudiux:
Nothing more to add.
Happy Holidays to you, @jaszhix, the team and all the contributors !

claudiux:
To test this applet, just modify the "last-edited" field in the metadata.json file of some Spices.

SpicesUpdater@claudiux/files/SpicesUpdater@claudiux/applet.js
```diff
@@ -0,0 +1,1529 @@
...
+        dconfEnabled = "/org/cinnamon/theme/name";
+        break;
+      case "extensions":
+        dconfEnabled = "/org/cinnamon/enabled-extensions";
+        break;
+      case "desklets":
+        dconfEnabled = "/org/cinnamon/enabled-desklets";
+
+    }
+    let [res, out, err, status] = GLib.spawn_command_line_sync('dconf read "'+dconfEnabled+'"');
```
jaszhix: I think it would be better to make these spawn calls asynchronous.

SpicesUpdater@claudiux/files/SpicesUpdater@claudiux/assets/spawn.js
```diff
@@ -0,0 +1,33 @@
+const Gio   = imports.gi.Gio;
+const GLib  = imports.gi.GLib;
+const Lang  = imports.lang;
+
+var SpawnReader = function () { };
+
+SpawnReader.prototype.spawn = function (path, command, func) {
```
jaszhix: Do we need to re-invent the wheel here? Cinnamon has a lot of utilities internally that makes this more compact. If you just need to asynchronously grab stdout and be notified of success or error, a function like [this one](https://github.com/linuxmint/cinnamon-spices-applets/blob/07449f8bc6e32218be2fa9c0f532344df4efa27b/IcingWindowSaver%40json/files/IcingWindowSaver%40json/3.6/applet.js#L20), or the one in this [Cinnamon PR](https://github.com/linuxmint/Cinnamon/pull/8130) would work well, and makes handling async callbacks easier.

SpicesUpdater@claudiux/files/SpicesUpdater@claudiux/applet.js
claudiux: I replaced calls to `dconf` via `GLib.spawn_command_line_sync` with calls to `Gio.Settings`.
(Near line 1283.)
```
_interface_settings = new Gio.Settings({ schema_id: _SETTINGS_SCHEMA });
 enabled = _interface_settings.get_string(_SETTINGS_KEY);
```

SpicesUpdater@claudiux/files/SpicesUpdater@claudiux/assets/spawn.js
claudiux: I removed entirely the folder `assets`, and the file `egSpawn.js`.
Now, `tail` and `zenity` are doing the job, with a few lines of code.

Thanks for all, @jaszhix !

SpicesUpdater@claudiux/files/SpicesUpdater@claudiux/applet.js
claudiux: Now, the applet tests if metadata.json is present, possibly in the subfolder /cinnamon for themes, before opening it.

claudiux: I do not understand why the widget is not on the stage, and how to put it on. (Line 509 of applet.js.)

collinss:
Personally, I think it's a bad idea to have this as a 3rd party applet. The reason is that, as a 3rd party applet, we could potentially have multiple versions of the applet with multiple versions of Cinnamon, and so if we change something about the way applets are installed, etc (this does happen from time to time), it could potentially break something for the user which wouldn't be fixed by removing the applet.

Also, as @jaszhix mentioned, that script is not intended to be used by anyone but applet (et al) developers who wish to test their code without actually pushing their changes and waiting for it to be merged.

I would much rather see a stock Cinnamon applet, which uses Spices.py directly for all spices stuff. Such an applet would ideally do all the heavy lifting in python (where we already have all the necessary functions in one place), and then communicate status, commands, and notifications via dbus. This will be much easier to maintain, and also ensure that the version that users have always matches their version of Cinnamon. Plus it will help cut down on spurious and hard to track bug reports that might arise.

jaszhix:
@claudiux I think I agree with @collinss here. I appreciate the effort you put into this - I've attempted this myself once, but had to put it on the back burner.

Do you think it would be possible to re-purpose this into a Cinnamon PR? I think extension management should be handled by Cinnamon.

claudiux:
OK @jaszhix, I will do two things:

  * encode a "light" version of this applet, renamed "SpicesUpdate", which will only monitor updates;
  * encode a Cinnamon version, in python, using Spices.py (my first idea, in facts).

Is there a way to directly open the "Download" tab, for example in `cinnamon-settings applets`, in javascript (for applet) and in python?

jaszhix:
Yeah, if it's just notifying about updates that is probably a fair compromise. As for Spices.py, I'm not sure. @collinss usually works in that area.

claudiux:
Ok, I let @collinss work in his area ;) and I will propose an applet to just notify about updates.
I close this PR.

PR #2175, opened 2019-01-03
SpicesUpdate@claudiux v1.0.1 - Improves notifications. Added es.po.

claudiux:
Hi,

Can you merge this branch with your master one, please?

Best regards.
Claudiux

PR #2176, opened 2019-01-03
Spices update@claudiux cro lng

muzena:


claudiux:
Many thanks @muzena !
There are also two new phrases in the latest version: #2175 
Can you translate them, please?

PR #2177, opened 2019-01-03
Update cro lng

muzena:


claudiux:
@muzena can you rename this PR `SpicesUpdate@claudiux - Update of Croatian translation`, please? It will be taken care of faster.

claudiux:
Thank you, @muzena.

PR #2186, opened 2019-01-05
SpicesUpdate@claudiux - create de.po (German translation)

Mintulix:


claudiux:
Thank you, @Mintulix !

PR #2195, opened 2019-01-06
Cinnamenu@json - French translation: fr.po updated

claudiux:


claudiux:
@jaszhix The Cinnamon.pot seems out of date. You should update it.
Best regards.

PR #2198, opened 2019-01-06
SpicesUpdate@claudiux - refined German translation

Mintulix:
First translation was to literal, now it is more usual.

One off-topic suggestion: Why not set the default icon color in the case of updates to the Cinnamon Spices orange (like in icon.png) instead of blue?


claudiux:
Good suggestion, @Mintulix !

Thank you for the refined German translation.

PR #2221, opened 2019-01-20
Create nl.po

French77:
Dutch Translation Spices Update


claudiux:
Thank you @French77 !

PR #2222, opened 2019-01-20
To Fix broken applet: weather@mocturl Yahoo API support changed to OpenWeatherMap Support

Gr3q:
Pull request to make the weather applet functional again.

Added capability to support more than 1 API, but only OpenWeatherMap is implemented at the moment. 

Fully functional, 
I believe icons needs some tweaking and the translation testing/reworking because a huge amount of code was changed.


Gr3q:
Found a bug with the URIs, have to reopen the bug request

PR #2228, opened 2019-01-22
Update translation Spices-Update

French77:
Correction Dutch translation


French77:
Dutch Translation off VPN Look-Out Applet

jaszhix:
Looks like more than one translation file is included in this, the shut down applet translation was in #2227.

PR #2233, opened 2019-01-22
Remove usage of old appsys functions

clefebvre:
We want to remove these functions going forward as they are not needed.


clefebvre:
@NikoKrause @jaszhix please review these changes. I want to get rid of these functions in Cinnamon going forward.

The changes in this PR should work fine with master and past cinnamon releases.

jaszhix:
Tested all three applets and LGTM.

clefebvre:
thanks :)

PR #2234, opened 2019-01-22
Spices Update nl.po

French77:
Updated translation


jaszhix:
@French77 This is the wrong location for PO files.

French77:
> @French77 This is the wrong location for PO files.
My mistake has made a mess of it,shall try to upload the update off the Dutch translation later

PR #2242, opened 2019-01-24
Updates German translations for version 2.0.0

kghbln:


weather@mockturtl/files/weather@mockturtl/po/de.po
```diff
@@ -1,453 +1,402 @@
...
+
+#. settings-schema.json->apiKey->tooltip
+msgid "Copy this from your account on OpenWeatherMap and paste it here."
+msgstr "Kopiere ihn von Deinem Konto bei OpenWeatherMap und füge ihn hier ein."
+
+#. settings-schema.json->location->description
+msgid "Location"
+msgstr "Vorhersageort"
+
+#. settings-schema.json->location->tooltip
```
kghbln: @Gr3q This is the only thing I was unsure of since it differs from the rest. Since the source language works for the applet I just imitated the translation.

Gr3q: I will have a look at it after work, but if assume it works for you

kghbln:
@Gr3q No worries. I very much appreciate the work you put into this even more as I understand that you do it in your spare time, too. Anyways - great collaboration, I believe.

PR #2255, opened 2019-01-27
SpicesUpdate@claudiux - update German translation

Mintulix:
You update to often... ;-)


claudiux:
Yes, I know, but I am being asked for so much change ... And I have a hard time refusing when it is an improvement. ;-)

https://cinnamon-spices.linuxmint.com/applets/view/309

PR #2262, opened 2019-01-29
InternetTimeApplet@stefan : new

stefan12O:
I would like to publish this Applet.


InternetTimeApplet@stefan/files/InternetTimeApplet@stefan/applet.js
```diff
@@ -0,0 +1,39 @@
...
+		let h = d.getHours();
+		let m = d.getMinutes();
+		let s = d.getSeconds();
+		let tzoff = 60 + d.getTimezoneOffset();
+		let beats = ('000' + Math.floor((s + (m + tzoff) * 60 + h * 3600) / 86.4) % 1000).slice(-3);
+		this.set_applet_label(_("@" + beats));
+		
+		// set update intervall:
+		//  1 beat = 86.4 sec.  Shorter update interval means more precise changing time of the .beat
+		this.timeout = Mainloop.timeout_add_seconds(10, Lang.bind(this, this.refresh));
```
jaszhix: Hi @stefan12O. This will keep looping after the applet is removed from the panel. To have finer control of the loop, it should not recursively create new loops. The handler can return true to keep looping, or a falsey value (functions default to returning undefined) to loop once. To fix this you will need to add an `on_applet_removed_from_panel` method that stops the loop, so it doesn't impact the thread after its removal. An example, feel free to borrow.

```js
NetBeatApplet.prototype = {
    __proto__: Applet.TextApplet.prototype,

  _init: function(orientation, panel_height, instance_id) {
    Applet.TextApplet.prototype._init.call(this, orientation, panel_height, instance_id);
	  this.set_applet_label(_("@Ste"));
	  this.set_applet_tooltip(_("Internet Time\n.beat"));

    this.refresh();

    // set update interval:
    //  1 beat = 86.4 sec.  Shorter update interval means more precise changing time of the .beat
    this.timeout = Mainloop.timeout_add_seconds(10, Lang.bind(this, this.refresh));
    this.shouldUpdate = true;
  },

	refresh: function() {
		let d = new Date();
		let h = d.getHours();
		let m = d.getMinutes();
		let s = d.getSeconds();
		let tzoff = 60 + d.getTimezoneOffset();
		let beats = ('000' + Math.floor((s + (m + tzoff) * 60 + h * 3600) / 86.4) % 1000).slice(-3);
		this.set_applet_label(_("@" + beats));
		
		
		return this.shouldUpdate;
	
	},

  on_applet_removed_from_panel: function() {
    this.shouldUpdate = false;
    if (this.timeout) Mainloop.source_remove(this.timeout);
    this.timeout = 0;
  }
};
```

Starting with Cinnamon 4, there are now conventional `setTimeout` and `setInterval` methods that wraps Mainloop and behaves the way you would expect in the browser or nodejs. You can use these, but will need to specify 4.0 as the supported version in your metadata.json file using the `cinnamon-version` key with a string array of keys. E.g., `["4.0"]`.

stefan12O: Thank you very much @jaszhix . Currently my biggest challenge is handling with GitHub  :-)
I updated the applet.js file similar to your suggestions: moved timeout to the _init function and added on_applet_removed_from_panel function.
I am not sure if I need to launch a new Pull Request, somehow I don't manage to do so.
I am on Debian strecth/stable, so my Cinnamon Version is 3.2.7.

PR #2266, opened 2019-01-29
SpicesUpdate@claudiux: Update Croatian language

muzena:
@claudiux 
Can you add option to auto update spices?


jaszhix:
You might be interested in the discussion that took place in https://github.com/linuxmint/cinnamon-spices-applets/pull/2165.

PR #2276, opened 2019-01-31
SpicesUpdate@claudiux: Update German translation

Mintulix:
#2274 

It's a new dawn
It's a new day
It's a new pot for me, yeah
...
And I'm feeling good

Hi @claudiux,
perhaps you could refresh the pot applet description (about...):

metadata.json:7
msgid "Warns you when your Spices (applets, desklets, extensions, themes) require an update." 

It actually should be something like
"Warns you when installed Spices (applets, desklets, extensions, themes) require an update or new Spices are available." 

I already did this in German within this po file, so I don't need another pull request for this.


claudiux:
Hi @Mintulix,

I will do it!

Many thanks.

PR #2285, opened 2019-02-01
weather@mockturtl: Make compatible with previous Cinnamon versions

### LINKED ISSUE #2275: Weather@mockturtl 2.1.0 does not work below Cinnamon version 3.8
Opening the issue to get it some visibility.

Indicator: applet does not load, the reason is SyntaxError in the LookingGlass log. (Because of using async functions)

Example of Error:
```
error t=2019-01-30T19:41:59Z missing } after property list 
error t=2019-01-30T19:41:59Z [Applet "weather@mockturtl"]: Error importing applet.js from weather@mockturtl 
error t=2019-01-30T19:41:59Z Could not load applet weather@mockturtl
```

It won't be fixed in the next update, but I will provide version 2.0.1 in the readme.

jaszhix:
@Gr3q @mockturtl 

This moves version 2.0.1 to a 3.6 folder, and 2.1.1 to a 3.8 folder. This allows users on Cinnamon <3.8 to continue using the weather applet without needing to stop using newer JS features on newer versions of Cinnamon.

For reference, Cinnamon 3.8 and 4.0 uses mozjs52, 3.4-3.6 uses mozjs38, and 3.2 uses mozjs24. I kept 3.2 in there for Debian stretch users. If unsure about what is supported, MDN has a compatibility table at the end of most of their docs.

Closes #2275


Gr3q:
I'm up for it.

the one in 3.6 folder currently supports 1.8 as you know, and I don't really know how that will change just yet.

PR #2296, opened 2019-02-04
weather@mockturtl Update 2.1.3

Gr3q:
* Multiversion fix. Everything below 3.8 should use the previous version now (3.6 version moved to the root folder), because it does not even work with version 3.2 at the moment.
* small addition for DarkSky condition processing for shorter conditions.

My system correctly uses version 3.8.

This structure needs to be tested with Cinnamon version less or equal 3.2


Gr3q:
Confirmed to work on 2.6.11

PR #2304, opened 2019-02-06
Update ISSUE_TEMPLATE

French77:
Add screenshots to issue


germanfr:
Please don't recommend imgur in the issue template. Images are harder to reach and they tend to disappear over time. It's better to just post photos here when needed.

PR #2308, opened 2019-02-08
Display the alarm time in menu and panel

sphh:
I always wanted to know the time, when the time will finish. I added this with this pull request. It is even possible to format the alarm time. Because some of the alarm times will be in the far future, I believe it's not necessary to show too much detail, hence two time formats can be specified.

I also changed the order of the default Preset Times to be strictly in descending order (the defaults of 2 and 3 hours were in the wrong order).

I still have one question: Is it possible to add a link in the label of a setting? It would be great to point the user to the website with the available format specifiers. At the moment this is included in the tooltip, but that is far from ideal.

### DIFF
--- a/cinnamon-timer@jake1164/README.md
+++ b/cinnamon-timer@jake1164/README.md
@@ -15,6 +15,10 @@ All settings can be changed by right clicking on the icon and selecting configur
 ## TODO:
 
 ## Changelog
+* 1.2.0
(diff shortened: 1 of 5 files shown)


jaszhix:
@jake1164

jake1164:
This looks good, Thanks @sphh!

PR #2331, opened 2019-02-23
weather@mockturtl Update: Adding OpenWeatherMap Conditions

Gr3q:
* Added condition descriptions for translations to support not supported languages on OpenWeatherMap, courtesy of @jorgenqv 
* Removed the Translations section from the older versions as one place is enough for the translations to work.
* Updated template

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/3.6/applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/3.6/applet.js
@@ -1284,12 +1284,7 @@ MyApplet.prototype = {
   }
 };
 
-//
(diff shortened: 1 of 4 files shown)


jorgenqv:
@Gr3q There seems to be some missing line breaks that puts around 8 snow conditions in weather@mockturtl/files/weather@mockturtl/3.8/applet.js on row 1262 and are because of this missing in the pot template file?

Gr3q:
@jorgenqv Thanks, fixed. Visual Studio Auto formatted it incorrectly and I did not notice.

PR #2345, opened 2019-03-02
github-projects@morgan-design: Update sv.po

eson57:
Update Swedish translation


github-projects@morgan-design.com/files/github-projects@morgan-design.com/po/github-projects.pot
```diff
@@ -1,209 +1,436 @@
-# SOME DESCRIPTIVE TITLE.
+# Swedish translation for github-projects@morgan-design.com.
```
jaszhix: Shouldn't this be in an sv.po file? This is the template pot file.

eson57: Sorry, I don't understand. It is sv.po as far as I can see.

PR #2358, opened 2019-03-10
sysmonitor@orcus: polish translation; new screenshot; new default settings

pyroxar:
@OrcusCZ 
I added polish translation; new screenshot; new default settings


OrcusCZ:
Hi, thank you for the translation, but I have to say I don't agree with the rest.

What is the reasoning for changing default values and screenshot? The proposed default values only change the applet padding and whether the border is drawn, which I consider subjective if a user likes this more than the current default values and if not, they can always change it in the settings. Custom padding also overrides the default Cinnamon applet padding, which I believe should not be overridden by default.
The proposed screenshot shows a horizontal applet with proposed default settings, but the current screenshot shows both horizontal and vertical applet while the vertical is being drawn with borders, therefore I think it is more representative of the applet.

You can create pull requests for translations alone without applet owners agreeing, so if you want, you can either close this and create a new pull request for the translation or revert changes to default values and screenshot in this pull request and then I will agree to it.

pyroxar:
@OrcusCZ 
I restored the settings and the screenshot.

OrcusCZ:
This looks good, thank you. I approve this PR.

PR #2375, opened 2019-03-18
wireguard@nicoulaj.net: initial import

nicoulaj:
This is a simple [WireGuard](https://www.wireguard.com/) systray applet.

image

From https://github.com/nicoulaj/cinnamon-applet-wireguard tag 1.0.1


wireguard@nicoulaj.net/files/wireguard@nicoulaj.net/applet.js
```diff
@@ -0,0 +1,212 @@
...
+
+    on_applet_added_to_panel(userEnabled) {
+
+        if (!this._net_monitor) {
+            this._net_monitor = Gio.network_monitor_get_default();
+            this._net_monitor_id = this._net_monitor.connect('network-changed', (monitor, network_available, user_data) => this._on_net_changed());
+        }
+
+        if (!this._wg_monitor) {
+            this._wg_monitor = Gio.file_new_for_path("/etc/wireguard").monitor_directory(Gio.FileMonitorFlags.SEND_MOVED, null);
```
jaszhix: Might be smoother to check if wireguard is installed, something like this would work.

```js
let path = Gio.file_new_for_path("/etc/wireguard");
if (!path.query_exists(null)) {
  Main.notify('Please install wireguard')
  // ... notify the user and let the applet load, otherwise they may never know and the applet will remain enabled on cinnamon restart.
  return;
}
```

nicoulaj: thanks, updated

PR #2384, opened 2019-03-23
windowlist@cobinja.de Improve performance for monitor names

Cobinja:
This improves performance when opening the context menu by only fetching monitor names when the monitor configuration is actually changed


Cobinja:
Addition: Follow-up for #2383

PR #2391, opened 2019-03-26
Weather@mocturtl Update 2.1.8

Gr3q:
Before I start, just a disclaimer. 
I had problems with back-porting 3.8 (and a lot of work), so after a couple tries I was like 'fuck this' and went through on it with Babel. added the required polyfill for async and Array.includes, commented out the TimeZone dependent code because .toLocaleString does not have it in the previous versions. 

Tell me if this is not ok, otherwise it will stay like this so I can focus on implementing new providers or features instead.

3.6 (from now on 3.4) compatibility version:
* All functionality is now included from 3.8 except Timezones.

Changes Tested on Cinnamon 3.4 and 3.6, versions below and above are tested as well, everything seems functional.


weather@mockturtl/files/weather@mockturtl/3.4/applet.js
```diff
@@ -0,0 +1,2345 @@
...
+
+// Polyfill for Babel async: regenerator-runtime
+/**
+ * Copyright (c) 2014-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+var runtime=function(g){function D(a,b,c,e){b=Object.create((b&&b.prototype instanceof y?b:y).prototype);e=new z(e||[]);b._invoke=K(a,c,e);return b}function A(a,b,c){try{return{type:"normal",arg:a.call(b,c)}}catch(e){return{type:"throw",arg:e}}}function y(){}function t(){}function q(){}function E(a){["next","throw","return"].forEach(function(b){a[b]=function(a){return this._invoke(b,a)}})}function u(a){function b(c,d,l,f){c=A(a[c],a,d);if("throw"===c.type)f(c.arg);else{var e=c.arg;return(c=e.value)&&
```
jaszhix: I would wait until Cinnamon is using mozjs60 or newer before using async/await. regenerator-runtime will add quite a bit of overhead, and we can't forget the process is also managing compositing and file I/O.

weather@mockturtl/files/weather@mockturtl/3.4/applet.js
```diff
@@ -0,0 +1,2345 @@
+"use strict";
```
jaszhix: If you're going to transpile with Babel, please include the original source file somewhere, along with the build scripts used to generate it.

weather@mockturtl/files/weather@mockturtl/3.4/applet.js
Gr3q: I will not use babel and regenerator but I could transpile the code from typescript. It uses its own generator (smaller).

This is obviously not ideal, but people's main concern that OpenWeatherMap is not accurate enough, and there is no other option below 3.8. Hence the pr as I know I don't have time/won't be able to do the same with Promises reliably.

weather@mockturtl/files/weather@mockturtl/3.4/applet.js
Gr3q: i will do that, don't know why I did not include them

Gr3q:
I am going to open another Pull request and close this one when I get some free time.

PR #2403, opened 2019-04-05
asl@santiago: Information to American Sign Language vocabulary, addin…

SamBurgos:
…g ASL applet

This cinnamon applet shows some vocabulary of American Sign Language (commonly known as "ASL") so that people can learn about it


NikoKrause:
Thanks for contributing to cinnamon spices. I have a few remarks though:

(1) The applet is broken. When I try to add it to the panel, the applet doesn't show up. -> The uuid in the metadata is not correct.
(2) It's in Spanish ("Abecedario" instead of "Alphabet", "Letra X" instead of "Letter X"). -> The applet should be written/coded in English and then should be translated with gettext to any other language.
(3) The applet is just opening a website, where a short video is showing the letter in ASL. I like the idea of an ASL applet, but what is the advantage of this applet in contrast to googling?

SamBurgos:
Greetings Niko, I hope that this can answer your questions
(1) I guess that I forgot to do a minor change on the metadata, let me see if it is what I think and change it, else if you can help me out since I am kind of new to this world of Cinnamon spices and after testing I guess it worked previously but I may have failed one last thing

(2) I didn't know that particular situation about coding in English only and then translate it, I was hoping to do that in a future version, but right now I am having issues on my applet regarding the translate and gettext part; because if I add those (import and code per the documentation) the applet doesn't run at all

(3) My original idea was to somehow show a small video from Youtube or other site that opens on the very same applet which shows how to do the sign, since I need to work with some people I know and/or find a place to upload the video and show it. The current behavior of the applet is as I intend to work it for now (probably plan something else if I am able to keep learning about xlets). About the googling is really hard to explain but the idea is to make the search easier, instead of typing "letter A ASL", "movie in sign ASL" or something alike as examples you just open the applet, select the word depending of the category and once you click on it, it would show the site and how to do the sign; specially since it is really hard already (or at least for me, considering that I have been studying ASL for 10 months) to find a website among the many on how to say a particular word and take it as a reference (my mentor sometimes does words with variations that until now I have been unable to understand why it is different)

Hope this can help you and if you need something else from me please let me know, also I will try to keep reading documentation (which so far I have been between googling, the information of mint-doc + devhelp package, the current developer site and "reverse engineering" the behavior of other applets as well) and request some help from you, so we can work this out. Kind regards

PR #2414, opened 2019-04-14
printers@linux-man add Nl.po

French77:
Dutch translation


printers@linux-man/files/printers@linux-man/po/Nl.po
```diff
@@ -3,100 +3,100 @@
 # This file is distributed under the same license as the PACKAGE package.
```
jaszhix: It looks like you renamed the base template POT file to Nl.po. This will break translation maintenance, please copy instead.

PR #2418, opened 2019-04-15
printer@linux-man nl.po

French77:
Dutch translation Applet printer@linux-man


French77:
nl.po file is not placed in the po folder.
Screenshot

PR #2421, opened 2019-04-24
weather@mockturtl Update 2.2.0

Gr3q:
General changes:
* Source converted to Typescript.
* Added some basic Typescript declarations of Cinnamon js libraries (minimum needed for the applet)
* Added Build scripts and Build guide
* Text 'Today' and 'Tomorrow' is used for forecasts in 3.8+
* Version 3.6 changed to 3.4 and using the Typescript source compiled to es5 with some extra changes after:
   * All files moved into One file
   * code regarding importing other files was removed
   * .ToLocaleString() does not support Timezones, removed
   * Array.includes polyfill added (it is used in DarkSky)

Fixes: 
* !!! the panel was not rebuilding at the correct time on refresh (I honestly don't know how the app was working before at all), now rebuilding is part of the data refresh function.
* Applet was crashing when there was no internet (Debug line was outside try/catch)
* Big performance increase in version 3.4, plus DarkSky support
* OpenWeatherMap forecast conditions were always translated, fixed.
* Several fixes where undefined variables were referred to, now fixed (found while converting to TS)

New stuff: 
* 3.8+: Added Refresh button to context menu.


Gr3q:
@jaszhix If you disagree with how 3.4 version is in this PR, I will accept and revert that to the previous version. Either it will be reverted or not, I will consider that version final.

Gr3q:
I can also remove the 3.4 build script, I can see Codacity is still complaining.

Gr3q:
@jaszhix This one can be pulled now.
All versions tested, 3.4 version functioning identically to 3.8+, source included, build script for 3.8 included. Took out 3.4 build script because it was failing checks.

PR #2426, opened 2019-04-28
 [calendar@simonwiles.net] Adds feature to let the applet check and apply system clock setting 'use-24h' and 'show-seconds'

jakesmolka:
This change introduces honoring of the system clock settings 'use-24h' and 'show-seconds'. The changes are ported back from the current version of the system's calendar applet, see:
https://github.com/linuxmint/cinnamon/tree/ec6fb27452da571142bd39fccd8dbb5c64c066c6/files/usr/share/cinnamon/applets/calendar%40cinnamon.org

I usually don't code in JavaScript, so please someone have a quick look at the code for anti-patterns and such.

@simonwiles Thanks for the applet! Would you have a look at proposed changes and approve them if you agree?

### DIFF
--- a/calendar@simonwiles.net/files/calendar@simonwiles.net/applet.js
+++ b/calendar@simonwiles.net/files/calendar@simonwiles.net/applet.js
@@ -29,6 +29,7 @@ if (typeof require !== 'undefined') {
     Calendar = AppletDir.calendar;
 }
 const GLib = imports.gi.GLib;
+const Gio = imports.gi.Gio;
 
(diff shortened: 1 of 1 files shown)


simonwiles:
This looks sound to me -- I think the nature of the change means that no labelling or documentation changes are required too, which is a bonus; the new behaviour is better in every respect and much more likely to be what a user expects.  Thanks very much!

I'm not in a position to merge this or approve it "officially", but I do whole-heartedly approve of it in my capacity as the original author of this applet.

PR #2429, opened 2019-05-01
 printers@linux-man Added polish translation

pyroxar:
Added polish translation


pyroxar:
Please, remember to update printer@linux-man in /appsys

PR #2438, opened 2019-06-02
fix for unformated value

coolcow:
The value is not formatted if `Show currency` is set to `name`.

Values like `1234.5` are show as `1234.5 USD` instead of `1,234.50 USD`.


jaszhix:
Author @imDeprecated is no longer on Github, merged.

PR #2444, opened 2019-06-09
New Redshift UI Applet

raphaelquintao:
I wrote a new Redshift UI Applet, maybe can be useful for someone else.

screenshot


Gr3q:
Hello, 
I've got some problems with the app:

* If I use any of the sliders on the applet, the brightness/color temp only get reduced, so I can make my screen completely black/red in just a couple of clicks.
* If I change the brightness on the keyboard, the applet resets it if I click on it (I expected the app to respect manual change)
* Applet structure does not look like in the picture (If it is intended its ok, it's just I expected the header and footer part to be there)

raphaelquintao:
I wrote that on Debian with Cinnamon 3.8, so maybe really does not works below 3.8.

jcmljunior:
Hello,
It was very good, I find it interesting to consider an icon change as well.
You can use the Papirus icons for example.
Papirus: https://github.com/PapirusDevelopmentTeam/papirus-icon-theme

In case you want to make the code compatible with previous versions of cinnamon you can use a javascript compiler.
I know this compiler: https://babeljs.io/

raphaelquintao:
@jcmljunior The icon change the color =). I will consider a better icon. Thanks.

@Gr3q How i can do that? The metadata has some specific key to specify the minimal version?

Odyseus:
Hello, everybody.

@raphaelquintao: You could use the `cinnamon-version` key on the applet metadata.json file. It's a mandatory key for extensions, but optional for applets/desklets. But you cannot specify a minimum Cinnamon version; you will have to specify every version the applet works on. For example:

```json
"cinnamon-version": [
    "3.8",
    "4.0",
    "4.2"
]
```

And the installation and activation of the applet on a non listed Cinnamon version is not blocked; a warning message is displayed instead:

> Extension <UUID> is not compatible with current version of cinnamon. Using it may break your system. Load anyway?

***

> I will consider a better icon.

You could make them configurable by the user using the `iconfilechooser` setting type; if there are custom icons, use them, else use the default ones. Just an idea.

brownsr:
I found the gamma stopped working after a while, worked fine at first, might be a bug there.  Would be good if it worked in a vertical panel too, but as you are using an optional label to display information that limitation might be deliberate.  Otherwise seems to work OK, so will merge.

PR #2446, opened 2019-06-11
weather@mockturtl: Update sv.po

eson57:
Update Swedish translation from POT


weather@mockturtl/files/weather@mockturtl/po/sv.po
```diff
@@ -669,52 +678,161 @@ msgstr "Väder"
...
 #. 3.8->settings-schema.json->dataService->options
+#. 3.4->settings-schema.json->dataService->options
 msgid "OpenWeatherMap (no key needed)"
 msgstr "OpenWeatherMap (ingen nyckel behövs)"
 
 #. 3.8->settings-schema.json->dataService->options
+#. 3.4->settings-schema.json->dataService->options
 msgid "DarkSky"
-msgstr "DarkSky"
+msgstr "Mörk himmel"
```
Gr3q: I think the API name should not be translated

eson57: you're right.

PR #2450, opened 2019-06-14
SSH Connect Panel applet

foobar-beer:
This cinnamon panel applet can be configured via JSON config file (instead of the usual .ssh/config) file which is supported by other applets - this allows for grouping connections into submenus and configuring custom options.


Odyseus:
Hello, @foobar-beer.

I have a couple of suggestions for you.

You have hard-coded the configuration file to be inside the applet folder. This will cause the configuration file to be overwritten every time the applet is updated by the user, loosing all configurations. You can simply implement the Cinnamon's settings system on your applet. You can read the documentation about how to implement it (see at the end).

And here comes the nightmare part (LOL). You are using a terminal extracted from a setting. But you are using CLI parameters that are used very differently by different terminals.

- The `-t` (or `--title`) parameter in **Gnome Terminal**, depending on the distribution and the version of **Gnome Terminal**, may not exist at all because it was removed at some point upstream and added back by some forks.
- The `-t` parameter in **Xterm** means a totally different thing.
- In **Terminator** is used the `-T` parameter (uppercase).
- In **Tilda** the `-t` parameter is used to set transparency.

The point is, you can't rely on the `-t` parameter to set the title of a random terminal. There is a somewhat universal way to set a terminal title *on-the-fly*; by executing the command `echo -ne "\033]0;SOME TITLE HERE\007"`.

And the nightmare continues. The `-e` parameter is also problematic because in **Gnome Terminal** the `-e` parameter was removed (around 3.32 I think).

And lastly, you can avoid all function bindings (the use of `.bind(this)`) by simply using arrow functions. This isn't an issue at all, I'm just informing you that you could (not all JavaScript changes are useless syntactic sugar after all (LOL)).

### Documentation

For an updated documentation, install the packages `devhelp` and `cinnamon-doc`. Then open Devhelp from your applications menu. It not only contains a more up-to-date documentation that the one on-line, but also has documentations about the **St** library and Cinnamon's JavaScrit modules.

For documentation about GObject libraries (Clutter, GLib, Gio, etc.), exists [this on-line documentation](https://lazka.github.io/pgi-docs/) that can also be [downloaded for off-line use](https://github.com/lazka/pgi-docs) as a static HTML site, and also can be downloaded as [Devhelp books](https://github.com/pygobject/pgi-docs-devhelp).

sshconnect@foobar-beer/files/sshconnect@foobar-beer/applet.js
```diff
@@ -0,0 +1,115 @@
...
+    let addFlag = (flags !== undefined && flags !== '' ? flags : '');
+    
+    Main.Util.spawnCommandLine(terminal + ' -t "' + name + '"' + addProfile + ' -e "ssh ' + addFlag + ' ' + host + '"');
+
+    let notification = new MessageTray.Notification(this.msgSource, 'SSH Connect', _('Connection opened for ') + name);
+    notification.setTransient(true);
+    this.msgSource.notify(notification);
+  },
+
+  editConfig: function() {
```
Odyseus: You don't need to call an external script here. You can directly call `xdg-open` like this:

```javascript
editConfig: function() {
    Main.Util.spawn_async(["xdg-open", PATH_TO_CONFIG], null);
},
```

brownsr:
@foobar-beer I don't think there is anything that need stop me merging this.  What are your plans ?  Do you want to address some of @Odyseus very pertinent comments ?  If so then would you want to do this after merge, with a PR or would you want to make further changes before merge ?

foobar-beer:
@brownsr I defo want to have the configuration override issue @Odyseus mentioned addressed before merging (it's a really good point about quite annoying behavior) - might take another "nightmare" problems with it :-) I should be able to make some time to fix those during the upcoming week (weekend probably).

foobar-beer:
Hi @Odyseus,

thank you very much for your comments! Very helpful points! I've moved the config to Cinnamon Settings (even though the layout is not ideal this way (yet) and created configuration options for everything terminal related. This way it should be much more user-friendly.

@brownsr Thanks for the merge! That was quick :-)

PR #2455, opened 2019-06-18
spice name: MessagingMenuV3@blub, added more apps

igorer88:
- Spice name: Messaging Menu V3
- Author: @clefebvre (I think is the author, sorry if I'm mistaken)
- Description: I just added more apps to the array so the applet would have more apps compatible with.

### DIFF
--- a/MessagingMenuV3@blub/files/MessagingMenuV3@blub/applet.js
+++ b/MessagingMenuV3@blub/files/MessagingMenuV3@blub/applet.js
@@ -20,7 +20,7 @@ function _(str) {
 
 /* global values */
 let icon_path = "/usr/share/cinnamon/theme/";
-let compatible_Apps = ["skype", "pidgin", "empathy", "xchat", "kmess", "gajim", "emesene", "qutim", "amsn", "openfetion", "gwibber", "qwit", "turpial", "birdie", "pino", "fbmessenger"];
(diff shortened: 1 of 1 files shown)


brownsr:
I will merge this, but I can't verify correct operation for all the added messengers and I suspect at least one is not right. If the commands noted are not right nothing shows so I am not risking showing faulty menu items to a user, hence why I am happy to merge.

Slack works fine.  Telegraph works fine.  WhatsApp does not work, perhaps that is what you intended to get launched by whatsdesk or whatsie, anyway that does not work.  I've not tried fbmessenger or skype.  I suggest you install all the messengers you want to add and check correct operation, and PR any further changes required.

igorer88:
Hi @brownsr. 
My menú looks like this. 
messaging menu

- Whatsapp 
  - Let's get rid of __whatsie__ it is an abandoned project as the author says in his repo: 
  "Whatsie is not available anymore. Please use the official client. The following is a final version of the abandoned, unofficial _ Web client Whatsie. This project is/was not affiliated with Whap Inc. in any way. It will not be updated._"
There are some forks, but is better to stay with the working one as it is __whatsdesk__
  - I'm using __whatsdesk__ right now and I don't have any problem with it.
- Facebook Messenger
  - I will change __fbmessengerfordesktop__ to __messengerfordesktop__ sorry I couldn't test that one I just used before, now I have installed it and it works. 
-  I will create another branch to PR the __Spanish__ translation.

PR #2459, opened 2019-06-25
notification-mute: Initial commit of a new applet for toggling muting of notifications and showing the status

jgillula:
This is my first PR with the Cinnamon project. Please let me know if I messed up on the PR format, and I'll fix it right away!

Thank you so much for your review--I look forward to working with you on this.


brownsr:
Well it worked fine when tested, but there must have been something amiss with the metadata as I can't see it available for download.  I'll check tonight as to what that was, and either let you know or fix up directly as seems appropriate.

Regards

Simon

jgillula:
Thank you so much! I'm happy to do whatever needs doing on my end.

smurphos:
@brownsr 

There is an issue with Cinnamon Spices at the moment that is probably the reason why the new applet isn't available to download..

https://github.com/linuxmint/cinnamon/issues/8712

brownsr:
@jgillula The metadata needs correcting even if there is a problem with the Spices site (which I can't confirm is causing the download problem).  Could you please change the metadata.json, and put the uuid in the first line, name on the second, description in the 3rd, and get rid of that max-instances line entirely.  It may be that a download script is expecting the uuid in the first line, and I would like to eliminate that as a cause.  Secondly in the info.json could you please put your name in the author field - we only blank this out when the author is no longer around, or not wanting to maintain the applet, and we want to see that the applet is available for adoption.

I would want this as a new PR please.  Any issues doing this just let me know.

Regards

Simon

@smurphos thanks for the information

jgillula:
@brownsr Done! https://github.com/linuxmint/cinnamon-spices-applets/pull/2478

Thanks so much for your help!

claudiux:
@brownsr There is not a problem with the Spices site, where all the pages are up to date, but with the service updating the json files. https://cinnamon-spices.linuxmint.com/json/

brownsr:
@jgillula Should be downloadable now
@claudiux Seems that a non-ascii character in the metadata on an applet commit message jammed things up.  The script has been modified now, so all the applets bar the one with the problem should now be processing OK.

Regards

Simon

claudiux:
Many thanks @brownsr. All is OK now.

Issue #2462, opened 2019-06-27
cpu-monitor-text@gnemonix Applet creates way too much empty space that it will never use..


t3pfaffe:
### NON CRITICAL!

```
 * App Build -2017-07-17
 * Cinnamon version - 4.0.10
 * Distribution - Manjaro (latest rolling)  
 * Graphics - NVIDIA 980ti, 430.14
 ```

@gnemonix

**Issue**
 This applet creates a decent amount of space to its right that it will never even use. 
![image](https://user-images.githubusercontent.com/16065962/60275507-953b3c80-98e9-11e9-9fe6-e5460e442eff.png)
   You can see it in comparison to mem-monitor-text@datanom.net

**Steps to reproduce**
 Placing this applet anywhere in a panel. 

**Expected behaviour**
Applet should be only as long as is needed and would preferably be adjustable by the user.  

**Other information**

PR #2463, opened 2019-06-27
edited width of cpu-monitor-text@gnemonix

ThePfaffanater:
In solution to #2462 Changes the width to the right of the applet to be more reasonable leaving less dead space. It still adapts to the max percentage length, there is just less buffer. It was 3.5em and is now 1.5em.
Old:
image

New:
image

I am sort of new to this using github pullrequests so let me know if I did anything wrong.


brownsr:
@gnemonix  OK with you ?

gnemonix:
Seems fine. I've been trying to remember why I had it wider to begin with. Probably to work around an issue with font size or because I felt it looked better.

As I presently do not have Cinnamon installed, I am unable to test it myself.

PR #2479, opened 2019-07-19
Update rwall applet

zou000:
* update wallheaven URLs
* update to work with new cinnamon version


brownsr:
I couldn't get this to work at all.  Do you have to be logged into wallhaven or something else like that ?  I presume the changes are tested ...

Regards

Simon

zou000:
No, you don't need to be logged in. I am on a newly installed Debian 10
buster with cinnamon 3.8.8. Those are the changes I had to make to
.local/share/cinnamon/applets/rwall@typicalfoobar/bin/rwall to make the
applet work. I am not familiar with cinnamon development process, just
thought it might be useful.

-- Li

zou000:
> There is "alpha" prefix in the url on line 116. If you remove it the script works.

Thanks. Don't know how I missed that.

PR #2480, opened 2019-07-19
cpu-monitor-text@gnemonix: Fixed width of cpu monitor label

diogofcosta:
Changelist:

- Fixed a bug where the width of the label of cpu monitor had a space at the beginning which pushed it outside the visual box of the applet.

@gnemonix

As it was, the label for the CPU monitor applet had an extra two white spaces that were making it override other applets, I simply removed those whitespaces and set the min-width to be more consistent with for example _[mem-monitor-text@datanom.net](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/mem-monitor-text%40datanom.net)_


brownsr:
treated as bugfix, no reply from author, merged.

PR #2486, opened 2019-07-24
Remove usage of old appsys functions

mushroomzz:
We want to remove these functions going forward as they are not needed.


brownsr:
it would be helpful if you could split into a separate PR for each applet, as each applet has a different author and may react to your request differently.

Also there are conflicts that need resolving.

Regards

Simon

brownsr:
Sorry, I really need a PR per applet, particularly as the owners are active.  Closing for now.

Regards

Simon

PR #2495, opened 2019-07-31
hwmonitor@sylfurd : Added settings for scale_factor (width)

Hultan:
Added a setting for scale_factor. Can be set to a value between 1 and 5 (3 is the default value).


Hultan:
Just wanted to clarify, the scale factor I am talking about is the factor you multiply the panel height with to get the width of the applet.

Hultan:
I made a mistake, will make a new PR tomorrow...

PR #2504, opened 2019-08-03
metadata.json - Make compatible with Cinnamon 4.2

claudiux:
@jake1164 

Make this applet compatible with Cinnamon 4.2 just modifying metadata.json.

Regards.
Claudiux


claudiux:
This PR is similar to #2503. I close it.

Issue #1708, opened 2018-03-18
hwmonitor@sylfurd: Needs changes for vertical panel to work correctly


attila123:
```
 * Applet version/Build date: 2017-Nov-20 (basically the same as in git)
 * Cinnamon version (cinnamon --version): 3.6.7
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...): AntergOS / does not matter
 * Graphics hardware *and* driver used: Not sure / does not matter
 * Snippet or attachment of ~/.cinnamon/glass.log
 * Snippet or attachment of ~/.xsession-errors
 ```

**Notify author of applet**
Applet has no issued author/maintainer as per https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/CODEOWNERS @brownsr ?

**Issue**
The applet was designed to horizontal panel. It does not work well on vertical panel. Only the leftmost part of the CPU graph is shown (which gives a delayed values of CPU usage, so not really useful). I hate horizontal panels. They are simply not practical. Please make it work on vertical panels.

**Steps to reproduce**
Add the applet to a horizontal panel. It works correctly showing the CPU and MEM graphs from left to right.
To make the panel vertical: right click on your panel -> Modify panel -> Move panel. Click it on either the left or right part of the screen (where the possible new places for the panel are shown). Only the CPU graph is shown and I believe it is the leftmost part of it, so it shows delayed data.
For vertical panels it make sense to make them wider, e.g. max out the Panel width in Panel settings.

**Expected behaviour**
This applet should check whether it is running on a horizontal or vertical panel (hopefully there is some notification about that). For a horizontal panel, it would render itself the usual way.
For a vertical panel, it needs to get the current panel width and draw the CPU and MEM graphs in such a width, and put them on top of each other (e.g. the CPU could be on top of MEM). If the width of the panel changes, the applet should redraw itself accordingly (so that the user does not have to remove/re-add it because of that).
Bonus feature: the hight (in vertical panel) and the width (on horizontal panel) should be configurable. Looks a bit tall-ish on vertical panel.

**Other information**
I saw that the workspace switcher applet is aware of the horizontal/vertical panel thing. See /usr/share/cinnamon/applets/workspace-switcher@cinnamon.org/applet.js on filesystem. For the first sight, the required tricks are:
this.setAllowedLayout(Applet.AllowedLayout.BOTH);
Define handlers:
on_orientation_changed
on_panel_height_changed

[...]

PR #2509, opened 2019-08-03
hwmonitor@sylfurd : Now works in a vertical panel.

Hultan:
The applet does now work in a vertical panel. The user can now control the size of the applet better.


claudiux:
Fixes #1708.

PR #2537, opened 2019-08-19
Finnish translation

mahtiankka:
I hope I got the translation right now. Thank you for your patience.


claudiux:
Many thanks @mahtiankka !

Issue #2476, opened 2019-07-14
mem-monitor-text Inconsistent RAM occupation


olimexsmart:
Hi to everyone, on my system the RAM percentage displayed by the applet is different than the one I can find in System Monitor.

Specifically, the value displayed by the applet is lower (around 15%-20%) than the one in System Monitor. 

To understand which of the two value was wrong, I disabled swap and then I started opening tabs until Firefox crashed. It crashed when in System Monitor RAM was at 99% occupation, while the applet was displaying just 83%. 

I think there may be a bug at line 69 of the applet.js file, but I have no idea how it should be. 

Thanks

olimexsmart:
Being tired with it, I fiddled around with the code and found the solution.

`    _update: function() {
        try {
            GTop.glibtop_get_mem(this.gtop);
            this.maxmem = this.gtop.total;
            this.user = this.gtop.user;

            let percent = Math.round((this.user * 100) / this.maxmem);
            this.set_applet_label(this.mem_label + " " + percent.toString().slice(-3) + "%");
            this.set_applet_tooltip(_("Click to open Gnome system monitor"));
        }
        catch (e) {
            global.logError(e);
        }

        Mainloop.timeout_add(2000, Lang.bind(this, this._update));
    },
`

I'm a noob regarding to contributing in open source project, what do I do now? A pull request? Could someone help me with this?

Thank you

brownsr:
A pr would be great. If the git side is unfamiliar I put a very simple
primer in the reader of the cinnamon spices themes repo

olimexsmart:
> A pr would be great

I'm on it. Git is not a problem.

Since I don't have write permission to push a new branch for the Pull Request, is forking the default procedure?

brownsr:
Yes

PR #2548, opened 2019-08-26
mem-monitor-text: issue #2476, fix RAM occupation calculation

olimexsmart:
This is relative to issue #2476 

Thanks @brownsr for the hints

This my first pull request to an open source project, this feels nice!


brownsr:
@icarter09

brownsr:
Treated as a bugfix, you will probably want to have some correspondence with @icarter09 about further changes, if he is around.

PR #2559, opened 2019-08-28
weather@mockturtl: New updated fi.po file in Finnish.

mahtiankka:
weather@mockturtl: New updated fi.po file in Finnish.


brownsr:
best to get into the habit of a new git branch per change.  This is probably repeatedly pulling down the same change, and there are multiple different applets updated.  Not a problem for translations really, but good practice generally.

Regards

Simon

PR #2564, opened 2019-09-03
[placesCenter@scollins] Shows help in tooltip about Right click on a Recent File

claudiux:
@collinss 

Hi Stephen,

I propose an option to show an help message about the Recent Files section (default value: true).

Capture du 2019-09-03 09-43-27

I did this because I manage the computers of several old people and some of them told me that they would not easily remember the use of the right click.

I also updated the .pot and fr.po. files

I hope this will suit you.

Best regards.
Claudiux

### DIFF
--- a/placesCenter@scollins/files/placesCenter@scollins/applet.js
+++ b/placesCenter@scollins/files/placesCenter@scollins/applet.js
@@ -121,6 +121,13 @@ class PlaceMenuItem extends FolderTypeMenuItem {
     }
 }
 
+class RecentFileMenuItem extends IconMenuItem {
+    constructor(text, icon){
+        super(text, icon);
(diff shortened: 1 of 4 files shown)


claudiux:
Hi @collinss,

Like that? It is done. And it's obviously better than what I was proposing.

Capture du 2019-09-04 01-41-17

collinss:
Yeah, that's much better. Could you also move the "activate" handler to the new menu item class as well? Now that recents have their own class, it really makes more sense to have it there.

claudiux:
@collinss I moved he "activate" handler to the new menu item class. And now, when the folder containing the right-clicked file is opened, this file appears selected. To do that, the default app that opens folders (chosen by the user in Cinnamon Settings) is used. Usually, it's `nemo`.

PR #2570, opened 2019-09-05
sshlauncher@sumo: Fix: Use correct settings for terminal

t-oster:
Right now the plugin is using `org.gnome.desktop.default-applications.terminal` setting, which is not present on my machine. I guess the correct value should be `org.cinnamon.desktop.applications.terminal`


t-oster:
nevermind.... seems to work with the original line and not with mine. Gnome-Terminal is confused with the `-T` parameter though. after removing that in the original code, it works.

PR #2571, opened 2019-09-08
temperature@fevimu: Updated PT-pt translation

leplastic:
I updated the translation for new alerts functionality and corrected some other issues as suggested by POEdit.


leplastic:
I think the translations have some problem. After updating the applet, the strings in configuration panel are still in English. I tried to reset settings to factory defaults, uninstalling the applet and restarting the computer, but nothing seems to have effect. Do you know if the translations are cached somewhere?

@claudiux did you have any problem with french translations for the strings of your new alert functionality?

leplastic:
Sorry for the confusion, I was able to make it working. I removed the extension again and used the gulp script + cinnamon-spices-makepot utility and it is working fine. I then noticed that the downloaded version does not contain the translated file yet, although I believe this should be a cache problem.

PR #2572, opened 2019-09-08
placesCenter@scollins: Improvement of Recent Files section

claudiux:
@collinss 

Hello Stephen,

I propose you some improvements to the Recent Files section.

1. No more files without an icon!
2. Whenever possible, replace the mime type icon with the icon of the last application that opened the file.
3. Do not show deleted files.
4. Do not show the Clear button when the list is empty.
5. Use default recentManager instead of a new recentManager, to be synchronized with  the `menu@cinnamon.org` applet.
6. In settings:
   * the type of the `recentSizeLimit` parameter becomes `combobox` to avoid to call 10 times the `buildRecentDocumentsSection()` method when the user wants to display 10 more files in the Recent Files section.
   * a new button appears, to open the Privacy window of Cinnamon Settings.

To see the differences:

1. Example of Recent Files section in `menu@cinnamon.org`:
menu-recents
2. Example of New Recent Files section in `placesCenter@scollins`:
placesCenter-recents

Can you tell me if all this suits you, before I try to optimize and clean the code?

Best regards.
Claudiux

### DIFF
--- a/placesCenter@scollins/files/placesCenter@scollins/applet.js
+++ b/placesCenter@scollins/files/placesCenter@scollins/applet.js
@@ -28,6 +28,10 @@ function _(str) {
    return Gettext.gettext(str);
 }
 
+function icon_exists( iconName ) {
+    return Gtk.IconTheme.get_default().has_icon( iconName );
+}
+
 class IconMenuItem extends PopupMenu.PopupBaseMenuItem {
     constructor(text, icon){
         super();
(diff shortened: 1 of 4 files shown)


collinss:
Yeah, I don't see anything I have a problem with.

claudiux:
OK. I'll review my code to clean up and optimize it.

PR #2574, opened 2019-09-09
[sound150@claudiux] Now compatible with Cinnamon 4.2

claudiux:
Hi @brownsr,

This is a simple update of `metadata.json` to make this applet compatible with Cinnamon 4.2.

Best regards.
Claudiux


claudiux:
Thank you @brownsr.

PR #2586, opened 2019-09-16
SpicesUpdate@claudiux: Update German translation

Mintulix:
Hi Claudiux, I thought it was time to rework some sentences.
Greetings


claudiux:
Great! Thank you @Mintulix.

PR #2596, opened 2019-09-28
update

French77:


brownsr:
Sorry, I think the outcome of this PR is not what you may intend.  It will turn the readme into German, not add a translation.

Regards

Simon

PR #2612, opened 2019-10-03
SpicesUpdate@claudiux: Update croatian language

muzena:


muzena:
@claudiux 
Can you add option to: Automatically update spices, or for maybe each spices separate
Automatically update aplett
Automatically update desklet
Automatically update theme
Automatically update extensions

claudiux:
@muzena 
I will think about this...
Thank you for your update!

PR #2616, opened 2019-10-04
Adds the ability to increase up to 30 days the time between two checks

### LINKED ISSUE #2610: SpicesUpdate@claudiux: Almost daily updates
```
 * Applet version/Build date: 4.2.0
 * Cinnamon version: 3.2.7
 * Distribution: Debian 9.11
 ```

**Notify author of applet**
@claudiux 

**Issue**

Almost every day Spices Update notifies me of an update for itself. The applets menu does indeed let me update the applet, but the version number doesn't change, there is nothing new in the changelog, I see no different functionality and I generally doubt that a relatively simple applet like this needs daily updates anyway. So I assume that this is a bug, that the versions are indeed identical.

claudiux:
Closes #2610.


claudiux:
Thank you @brownsr.

PR #2623, opened 2019-10-11
SpicesUpdate@claudiux - Update de.po

Mintulix:
Hi claudiux,
even if I have to make a new Pot file - I think the forced reinstall of a spice is handled a bit illogical.

Why do I have to check the first box if I just want to repair/reinstall a spice? Perhaps I don't want it to be checked regularly? So I think checking the second box should be enough. What do you think?

Kind regards

### DIFF
--- a/SpicesUpdate@claudiux/files/SpicesUpdate@claudiux/po/de.po
+++ b/SpicesUpdate@claudiux/files/SpicesUpdate@claudiux/po/de.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: SpicesUpdate@claudiux v1.0.1\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2019-09-10 20:22+0200\n"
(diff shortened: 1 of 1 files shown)


claudiux:
Hi @Mintulix,

You are right. I will change that.

Kind regards

Mintulix:
Hi @claudiux, kind reminder: you said you would change the behavier of the second checkbox. Did you? At least you didn't change the hint
If you want to get the latest version of an xyz now, check both boxes.
After that you get a new German translation ;-).

claudiux:
Hi @Mintulix,

I tried to do that. It was source of many bugs. To implement this feature, I need to rewrite much of the code. I do not have time now. I'll think about it.

Thank you for your patience and your understanding.

Mintulix:
@claudiux, ok, then it might not be worth it. Thanks for the information. And don't worry about patience - I have to thank YOU! Hey, it will be in Cinnamon 4.4, congratulations! Translation is following...

claudiux:
@Mintulix Thank you very much!

PR #2632, opened 2019-10-14
Updated the list and icons of cryptocurrencies

AleksiMagner:


AleksiMagner:
In the repository the files were updated, the applet itself in the system was also updated. But there are no changes to the applet files in system. What could be the problem?

PR #2638, opened 2019-10-19
Update sshlauncher@sumo

Gr3q:
Changes:
* Added some options for using Theme icon
* Getting the default application from the correct gsettings
* Moved Force file recheck to context Menu
* Generally making sure that the correct arguments are used for terminal emulators

### DIFF
--- a/sshlauncher@sumo/files/sshlauncher@sumo/applet.js
+++ b/sshlauncher@sumo/files/sshlauncher@sumo/applet.js
@@ -195,7 +195,7 @@ MyApplet.prototype = {
   connectTo: function(hostname) {
     let command = "";
     let terminal = "";
-    let notifMessage = _("Connection opened to ") + hostname + _(" using ") + terminal;
+
     if (!this.empty(this._customCommand)) { // Custom command set in settings
       command = this._customCommand;
       terminal = command.split(" ")[0]; // Get terminal name from custom command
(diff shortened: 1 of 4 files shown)


sshlauncher@sumo/files/sshlauncher@sumo/applet.js
```diff
@@ -106,31 +190,48 @@ MyApplet.prototype = {
...
+        this.sendNotification(_("SSH Launcher"), _("Error: I can't open the terminal.\nYou need to set a default terminal first in Preferred Applications!\n\n(More help in Right Click->About->More Info)"));
+        return;
+      }
+
+      terminal = term;
+      command = term + " ";
+      command += this.buildTermFlags(this.getTermOptions(term), hostname);
+
+      let termArg = this.gsettings.get_string("exec-arg");
+      if (!this.isExecArgCorrect(term, termArg)) {};
```
Gr3q: What is the correct action when we know that the argument to run a command for the terminal is incorrect (obtained from gsettings)?
Some options:
* Notify about it in the notification and do nothing
* Notify about it in notification and override the argument to make sure the terminal is opened
* Do not use the gsettings exec-arg at all (or only as a fallback)?

Atm there is no feedback if the command has run successfully in script.

brownsr:
ping @sumo

Gr3q:
Thanks @brownsr, forgot about that

brownsr:
ping @sumo .  Will assume all OK unless I hear otherwise.

Regards

Simon

sumo:
Sorry, I am not able to test this till next week. Don't want to hold this up so looking at it, it seems all good.

PR #2640, opened 2019-10-20
[SpicesUpdate] v4.3.0: Takes better account of manually installed spices.

claudiux:
Hi @brownsr,

This is the version 4.3.0 of Spices Update, tested on many versions of Cinnamon. It gives a better behavior when many spices need to be updated and takes better account of manually installed spices.

Can you merge this branch with your master one, please?

Best regards.
Claudiux


claudiux:
Thank you @brownsr !

PR #2646, opened 2019-10-23
[vpnLookOut] Update settings-schema.json - Bugfixes for Cinnamon 4.2

claudiux:
For Cinnamon 3.8+:

  * Removes "dependency" option of "clientsList" (bug in Cinnamon 4.2).

  * Changes all "checkbox" types to "switch".


claudiux:
Thank you @brownsr.

PR #2648, opened 2019-10-24
[stopwatch@pdcurtis] FEATURE Make it more flexible, add some features, style to better fit in panel

l00k:
Add ability to modify time (buttons in contextmenu), add ability to modify name in context menu, show stopwatch name in applet actor, always display hours, style adjustment to better fit content


brownsr:
ping @pdcurtis

brownsr:
ping @pdcurtis

pdcurtis:
@l00k  Thanks for the suggested features and @brownsr for alerting me to the PR. I haven taken a while to comment as I wanted to try it out for a little while and to get my thoughts together.

@100kI am impressed - there are a number of interesting ideas and you have made some additions to the context menu that i did not realise were possible, although there are still a few anomalies which would need sorting out.

I do however have a number of concerns with some of the concept. The original applet was deigned to have a very low footprint on the toolbar so it could always be present and to be very simple and quick to activate as a timer during expense critical activities such as phone calls, time chargeable internet access etc. I therefore kept the count to minutes and seconds only with an indicator that it had over-run so the tooltip could be checked. I resisted adding many 'Bells and Whistles'  quite deliberately. This was made clear in the Rationale statement.

It was also made clear in the header that It was also intended to not only useful in its own right but is also provides a 'tutorial' framework for other more complex applets - for example it provides a settings screen and a 'standard' right click (context) menu which opens the settings panel and a Housekeeping submenu accessing help and a version/update files and also the gnome system monitor program in case you want to find out how much machine power this applet is using at various update rates. It also provides an example of how to implement l10n localisation/translation support.

So I would like to include some of the additions you have put forwards for the context menu when the anomalies have been removed but have concerns over changing the panel display to occupy addition space unless it is available as an option.

I currently do not understand the rationale for the ability to change the timer 'on the fly' unless it is intended as a substitute for a timer which brings me to another point - I do not thing there is currently a good combined timer, stopwatch clock applet giving the functionality we have come to expect in mobile devices running android etc.

I would appreciate your and perhaps @brownsr 's comments on whether  a useful way ahead would be for you to 'fork' my applet with your Features into  a new stand alone applet which could in the longer term provide not only a stopwatch but also a timer to which alerts etc could be added and perhaps even offer some clock functions such as global times.

In the meantime I have seen some anomalies in the modification of the title in the context menu where changes initially displayed incorrectly - sometimes a cinnamon restart then displayed the correct text but not every time. I also see problems which did not previously exist on vertical panels.

l00k:
@pdcurtis I personally use it as time tracker.
That is why I have added feature to be able to quickly rename each applet.
I have multiple applets added (as multiple tasks) - it is usable to move time between them - that is why I have added features to edit time.
Style is really subjective - I have to admit I didn't tested vertical panels.

pdcurtis:
Firstly I am happy for you to fork my applet and make a version which satisfies your specific requirements but to some extent it is up to the Cinnamon team whether an extra applet is accepted.

In any case I am working on integrating most of your modifications either in the baseline or as an option. 

The first phase would be changes to the context menu to give access to title and make a single line for the various start pause and resume options. However I still see anomalies in when title changes are applied even with an extra update u call

AleksiMagner:
Heading to the timer can be implemented in a tooltip. Then the applet will occupy the same space, but will become more informative. For example, add entry in the settings. If the field is empty, display standard header

pdcurtis:
@AleksiMagner The timer already displays in the tooltip as you suggest and can be longer than would be practical in the panel and it is set up in settings. THere can be up to ten instances of the timer with different headers. It also displays the 'status' and current count.

Stopwatch_Screenshot from 2019-10-30 11-39-18

I agree that a more comprehensive applet with a timer with various alarms would be nice but would be a lot of extra work although some of the groundwork exists in my Network Usage and Monitoring Applet (NUMA to the left) and the Battery Applet with Monitoring and Shutdown (BAMS to the right) .

pdcurtis:
@l00k I have implemented most of the changes for a new major version 2.2.0
### Version 2.2.0
  * Adds option option in Configure... to display Counter Heading in Panel.
  * New single line layout for Start, Pause, Reset, Continue and Continue from Start Time in Context Menu
  * Adds ability to edit Counter Heading in Context menu as well as Configure..
  * Adds option in Configure... to add experimental functions such as an ability to modify time by buttons in Context (right click) Menu - a possible precursor to a flexible timer function.
  * Implements changes proposed by @100k (Łukasz Dobrowolski)in #2648 and uses code provided by him to implement his suggestions.
  *  Minor changes in stylesheet.css to match the other applets in the suite.
Also
  * Updated CHANGELOG.md much as above
  * Updated README.md including acknowledging @100k major contributions by @100k to ideas and code.
  * Updated metadata.json to version 2.2.0

I suggest the time modifications should be changed to +- 1 Min +- 10 Mins and +- 1 hour as precursor to including a flexible timer. @100k  Any views as it a feature you use extensively?

l00k:
Looks nice! I appreciate fast reaction :D

brownsr:
So I am waiting for a new version to come through from @pdcurtis and then we can close this, if I understand correctly ?

Regards

Simon

pdcurtis:
@browns

Just a few tests to make sure. 24 hrs and I will provide a new PR??

@100k Any views on time series yet?

Peter

l00k:
@pdcurtis everything is ok.

PR #2658, opened 2019-10-28
Do not use temporary files

sphh:
The original applet calls the external script `batteryscript.sh` every 2 s (this is the default). This script writes two temporary files into the user's home directory. With the advent of SSDs, this might wear out the disk. This pull request captures and parses the output of the `upower` command directly.

**Up for discussion**: Should the applet check for the old temporary files in the home directory and removes them, if they exist?

Maintainer: @pdcurtis 

* Removes dependency on 'batteryscript.sh'. This script writes two files on
  every update, maybe wearing out the harddisk. Now it uses asyncronous calls
  to execute the 'upower' command directly and captures its output.
* Checks for dependency on 'upower'.
* Increases maximum refresh interval to 5 min (300 s).
* Simplification of the logic to select the icon to display.
* Removes (outdated) changelog.txt.


sphh:
Sorry, I made some mistakes where I put the files. I'll close this one and open another with the right file locations.

PR #2659, opened 2019-10-28
Batterymonitorwithoutlocalfiles

sphh:
The original applet calls the external script `batteryscript.sh` every 2 s (this is the default). This script writes two temporary files into the user's home directory. With the advent of SSDs, this might wear out the disk. This pull request captures and parses the output of the `upower` command directly.

**Up for discussion**: Should the applet check for the old temporary files in the home directory and removes them, if they exist?

Maintainer: @pdcurtis

Changes:
- Removes dependency on 'batteryscript.sh'. This script writes two files on every update, maybe wearing out the harddisk. Now it uses asyncronous calls to execute the 'upower' command directly and captures its output.
- Checks for dependency on 'upower'.
- Increases maximum refresh interval to 5 min (300 s).
- Simplification of the logic to select the icon to display.
- Removes (outdated) 3.2/changelog.txt.


pdcurtis:
@sphh Just to let you know I am aware of your PR and the changes look useful and positive!

I obviously need time to look in detail and run tests on my machines. How much testing have you done and has it been on a number of machines and flavours of Linux?

I note that you have also updated many (all) of the associated files such as changelogs and have spent a lot of time on this PR.

Only immediate comment is that we should update cinnamon-version to include 4.4 as we must be getting towards the early versions being available for test.

I have no strong views on removing old temporary files - "Perfect is the enemy of Good" (Voltaire)

sphh:
Glad that you like this PR (so far).

Since I only have one machine (Linux Mint 19.2 Cinnamon), I am sorry that I could test it only on this flavour. Here it is running without any problem. I tried to mimic the `batteryscript.sh` as closely as possible, so if `upower` does not change its output, it should work (AFAIK).

If you are so inclined, please feel free to advance the cinnamon version to 4.4!

My opinion is that "A program should always clean up behind itself" (???). But it is a overhead to check for those files every time the applet starts.

pdcurtis:
I have several machines I can access and also use persistent LiveUSBs for test but it takes time.

I wonder if there is a clean-up mechanism in Cinnamon when one uninstalls? I think not.

 I will try it out in the next few hours.

sphh:
I noticed that the applets forget their settings when they were uninstalled ...

Adding a button to the configuration to delete these two files could be a way forward.

claudiux:
@sphh @pdcurtis Maybe this will help you? https://github.com/linuxmint/cinnamon/pull/8919

pdcurtis:
Agreed - it avoids any overhead.

I have not seen any problems yet so you just need to acknowledge yourself, change to include 4.4 and do above and I am happy to proceed.

I am not sure how one alerts the system that I have agreed the PR. Failing anything else change title to include something like [AGREED] to attract attention as I can not change any flags.

sphh:
@pdcurtis I now added a button and updated the translation files. I also updated the cinnamon version to 4.4.

If you want to credit me, please feel free to add it wherever you seem appropriate.

The last time I made a pull request, I asked @brownsr if he could push it to the main repository. And he did.

PR #2687, opened 2019-11-17
Update nl.po

French77:
Spices Update Version 5.1.2


claudiux:
Thank you, @French77.

Issue #2698, opened 2019-11-25
Weather UUID: weather@mockturtl


Padre2:
* Applet version/Build date Version 2.3.2 (2019-10-30 05:43:33)
 * Cinnamon version 4.4
 * Distribution - Arch
 * Graphics hardware AMD
error t=2019-11-25T15:06:41Z [weather@mockturtl]: Extension is not compatible with current Cinnamon version
Cjs-Message: 14:57:22.346: JS LOG: [LookingGlass/error] [weather@mockturtl]: Extension is not compatible with current Cinnamon version
logError@/usr/share/cinnamon/js/ui/extension.js:134:17
Extension.prototype.validateMetaData@/usr/share/cinnamon/js/ui/extension.js:322:19
Extension.prototype._init/<@/usr/share/cinnamon/js/ui/extension.js:242:17

 ```

<!-- Author list: https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/CODEOWNERS -->

**Notify author of applet**
weather@mockturtl 

**Issue**



**Steps to reproduce**



**Expected behaviour**



**Other information**

Padre2:
With 1 instance loaded and **+** still showing in white colour, clicking the **+** changes the applet into the loading icon and adds a 2nd loading icon and the computer freezes (see attached photo), requiring Ctrl+Alt+Esc/Del to reboot .
On rebooting, it has many multiple instances across the panel (see attachment).
I cannot understand why it is necessary to allow multiple instances for this applet. As far as I am aware, applets change the **+** and **-** from white to grey or vice versa after using and only permit loading 1 instance as, I believe was was the case before.
![Weather_2x_loading_freezing](https://user-images.githubusercontent.com/38849451/69794917-bfc55480-11fd-11ea-85d2-552d21e34a42.jpg)
![Multiple_weather_after_reboot_2019-11-28 15-59-53](https://user-images.githubusercontent.com/38849451/69794981-d8ce0580-11fd-11ea-986d-f78dbb1bac4e.png)

Padre2:
*

Padre2:
weather@mockturtl
Sorry, was closed. Must have clicked the wrong button

Gr3q:
It was limited to 3 applets (for multiple places). I do not know why it is not enforced (anymore? I never really checked).

Padre2:
@Gr3q  weather@mockturtl
[...]

PR #2699, opened 2019-11-25
Update compatibility for 4.4

Gr3q:
I did this 1 month ago to prepare for the next version, then decided to write _4.3_ into the array.

I must be retarded.

Resolves this: https://github.com/linuxmint/cinnamon-spices-applets/issues/2698

Edit:

I Include some other Applets as well, as they work on 4.4


brownsr:
not sure about the 4.6 compatibility noted there, a little premature !  However of no consequence right now.

Regards

Simon

Gr3q:
This happens on every release, and I always forget to add the next version to the array on time. And it seems I'm not alone.

I wish I could get notifications about these things before they happen.

PR #2702, opened 2019-11-25
Weather@mockturtl: Separate weather provider classes

Gr3q:
No functional changes.

The purpose of this PR is to increase separation  between weather service classes and the main app.

This change makes it straightforward to add new weather providers for other people, as the return object for the weather provider and the access of the applet functions are clearly defined.


brownsr:
Regression tested ?

Gr3q:
The PR is ready, I've tested it on 4.4, 3.6 and 3.0. It seems everything works as before.

PR #2726, opened 2019-12-01
mem-monitor-text: added warning on wigh memory usage

olimexsmart:
My laptop has very little RAM (4Gb) I if I don't watch my browser tabs I end up with the system crashing on low memory. 

So I made this simple addition that the spice text start flashing red when the RAM usage is above a threshold specified in the spice settings. By default the feature is disabled. 

Have a nice day,
Luca


brownsr:
ping @icarter09

icarter09:
@brownsr can you merge this PR for me? I don't have the admin rights. Thanks.

PR #2728, opened 2019-12-01
Added german translation to the shutdown applet

desperateCoder:
Hope, you can use that, my dad would be happy with that i guess :smile:


icarter09:
@brownsr can you merge this PR for me? I don't have write access to this repository. Thanks.

PR #2732, opened 2019-12-04
[SpicesUpdate@claudiux] Update de.po

Mintulix:
Hi @claudiux , I waited for further po-changes, but there are none... ;-)

To update this applet in a lower frequency - would it be possible to combine all po-updates at once and merge/release it only then? But I don't know how to manage that, and I fear it creates too much extra work.


claudiux:
Hi @Mintulix, thanks for all the German-speaking users who were waiting for a better translation.

This applet will not have any changes for a moment (except to fix possible issues). I'm working on other projects.

I plan to change notifications management... in a few weeks... or a few months.
The next

claudiux:
@brownsr Can you merge this branch with your master one, please?

PR #2734, opened 2019-12-05
Update metadata.json

Padre2:
Change max-instances from 3 to 1 to prevent crashing Cinnamon 4.4.3

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/metadata.json
+++ b/weather@mockturtl/files/weather@mockturtl/metadata.json
@@ -2,7 +2,7 @@
   "uuid": "weather@mockturtl",
   "name": "Weather",
   "description": "View your local weather forecast",
-  "max-instances": 3,
+  "max-instances": 1,
   "version": "2.3.4",
   "multiversion": true,
   "cinnamon-version": ["2.2", "2.4", "2.6", "2.8", "3.0", "3.2", "3.4", "3.6", "3.8", "4.0", "4.2", "4.4", "4.6"]


claudiux:
@mockturtl

Gr3q:
I do not think that this is the right way to solve the issue, this only works around the problem that the applet manager does not enforce max-instance setting correctly nor it prevents the user to spam the Add button. 

The issues related to the applet manager should be raised in the main cinnamon repo [here](https://github.com/linuxmint/cinnamon/issues) first as it affects all applets.

Also giving users the option to have multiple weather applets is a feature already enabled for the applet, and I usually vary to take features away even they are broken (I'm pretty sure we will break at least a few people's workflow disabling it).

 If the others think the feature should be disabled for the weather@mockturtl applet, then I won't say no either.

Padre2:
@Gr3q  @mockturtl @claudiux
I have only recently started using Linux and am trying to get up to speed with Arch and my chosen DE Cinnamon. I am getting the latest updates daily and Cinnamon is already at version 4.4.3, whereas even Manjaro (Arch) is still on Cinnamon 4.2.4.

An objection to removing the ability to allow users to load multiple instances of this Weather applet is puzzling. I cannot conceive of any need to display 2 or 3 instances of this applet on a desktop panel. What am I missing?

I must accede to your decision,  because my knowledge of the coding in Cinnamon is very limited and this is my first PR.  As far as raising this issue in the the main Cinnamon repo, I will leave this to you, as I don't understand how this change would affect all applets.  This is a popular applet and I think there will be a more issues raised as other distros catch up. I have had no issues at all with Weather since I made the suggested change, so will be ok until there are updates.

mockturtl:
>  I cannot conceive of any need to display 2 or 3 instances of this applet on a desktop panel.  What am I missing?

Some users like to know the weather in multiple cities.  I can imagine a better interface to support it, but multiple instances are an easy workaround.

Likewise, if multi-instance is broken in a recent Cinnamon release, editing the metadata file feels like a "hotfix" until the underlying problem is resolved.

I suggest we use votes on this PR to decide which is causing more pain.  In any case, it will help if someone with more information [opens an issue upstream](https://github.com/linuxmint/cinnamon/issues).

Padre2:
@mockturtl
> Some users like to know the weather in multiple cities. I can imagine a better interface to support it, but multiple instances are an easy workaround.
> 
> Likewise, if multi-instance is broken in a recent Cinnamon release, editing the metadata file feels like a "hotfix" until the underlying problem is resolved.
> 
> I suggest we use votes on this PR to decide which is causing more pain. In any case, it will help if someone with more information [opens an issue upstream](https://github.com/linuxmint/cinnamon/issues).

Ok, now I understand.  No problem for me right now, if not merged. We can wait and see how this develops.

Gr3q:
https://github.com/linuxmint/cinnamon/issues/9004 opened

Padre2:
@mockturtl
> Some users like to know the weather in multiple cities. I can imagine a better interface to support it, but multiple instances are an easy workaround.
> 
> Likewise, if multi-instance is broken in a recent Cinnamon release, editing the metadata file feels like a "hotfix" until the underlying problem is resolved.
> 
> I suggest we use votes on this PR to decide which is causing more pain. In any case, it will help if someone with more information [opens an issue upstream](https://github.com/linuxmint/cinnamon/issues).

Ok, now I understand.  No problem for me right now, if not merged. We can wait and see how this develops.

brownsr:
Closing as any issue appears to lie elsewhere.

PR #2757, opened 2019-12-18
Fixed some issues reported by the users.

raphaelquintao:
* 1.4
  - Fixed icon scale.
  - Added support to version 4.3 and 4.4


raphaelquintao:
This update has not yet appeared on the site.
Has something happened? How does the site update work?

PR #2774, opened 2019-12-26
gnote@brett-smith: Gnote Applet, Initial commit of the Cinnamon Applet

brett-smith:
This first release features include :-

 * Auto-activation of Gnote via DBus
 * Lists the 15 most recent notes
 * Allows creation of new notes
 * Allows inline search of notes
 * Dynamically updates when notes change externally


brownsr:
Happy to merge this. The icon could ideally do with being square (the validate spice script picks this up, see the readme at the head of the repository).  There are also some log messages that get shown in .xsesssion-errors, presumably from your debugging.  You might want to turn those off, as I expect some users will report this as a bug otherwise.

Regards

Simon

brett-smith:
Thanks for merging and your comments. I've removed all the debug output fine (it seems there was actually a problem with how i was rebuilding menu items anything, and this exposed that). 

However, I am not sure I understand what you mean by 'square icon'. I had run validate-spices as instructed in the README and it did not report any problems with the icon. Another user reported that I wasn't using a symbolic icon correctly, is this what you mean? https://github.com/linuxmint/cinnamon-spices-applets/issues/2775

remicle:
Joining this conversation as similar error seen.  It seems that for the icon file: "add-note-symbolic.svg" , the "xmlns:osb=..." and related references were omitted.  This caused the reference "osb:paint ..." under "<linearGradient" to error as not defined.

After adding the same references used in "gnote-panel-symbolic.svg", the error not longer occurs.

Regards.

PR #2776, opened 2019-12-27
gnote@brett-smith: Fixes #2775 and other comments from initial merge.

### LINKED ISSUE #2775: [gnote@brett-smith] Applet icon not symbolic
System data irrelevant.

@brett-smith

### Issue

Applet not using symbolic icon. This causes the icon to be light when the applet is placed in a panel with light background.

### Steps to reproduce

Install and enable applet.

### Expected behaviour

The applet should use a symbolic icon.

### Other information

The issue can be fixed by just renaming the icon file at **icons/gnote-panel.svg** to **icons/gnote-panel-symbolic.svg** and editing its name [where the icon is used](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/gnote%40brett-smith/files/gnote%40brett-smith/applet.js#L339).

Not an error but a side note. There is no need to use paths when the icon files are stored inside a folder called **icons** inside an xlet folder. [Cinnamon automatically *appends that folder*](https://github.com/linuxmint/cinnamon/blob/master/js/ui/extension.js#L378) so the icons can then be used by name (for example: `this.set_applet_icon_symbolic_name("gnote-panel-symbolic")`).

brett-smith:
* Different icon.
 * Symbolic icon.
 * Got rid of debug output and other error messages.
 * Changed how menu is rebuilt on changes.

### DIFF
--- a/gnote@brett-smith/README.md
+++ b/gnote@brett-smith/README.md
@@ -14,3 +14,18 @@ If not already installed by your distribution, install the gnote itself.
 ```
 sudo apt install gnote
 ```
+
+## Changelog
+
+### v1577449518
+
+ * Different icon.
+ * Symbolic icon.
+ * Got rid of debug output and other error messages.
+ * Changed how menu is rebuilt on changes.
+
+### v1577357545
+
+ * Initial release
+ 
+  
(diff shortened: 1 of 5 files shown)


claudiux:
@brett-smith Thank you for this applet. FYI, the field "last-edited" is useless in metadata.json. This field will be created by the server producing the zip file allowing the installation of this applet.

brett-smith:
@claudiux ah ok thanks, noted for next time.

PR #2783, opened 2019-12-30
[SpicesUpdate@claudiux] v5.2.1: Make it configurable (Cinnamon 2.8 -> 4.0)

claudiux:
Hi @brownsr,

When trying to install this applet under Debian 10 (Cinnamon 3.8), I noticed that the settings-schema.json file should not contain a "units" field. This is fixed.

Best regards.
Claudiux


claudiux:
Many thanks @brownsr !

PR #2803, opened 2020-01-09
Weather wind text translation fix

dmiwell:
Wind units and direction were not translated, so it needs fix. Line was too long to read, so I formatted with prettier


brownsr:
A significant change on this applet has just been merged.  Could you look at your changes on top of the new code and check it is still OK, please ? Let me know either way.

Regards

Simon

Issue #2816, opened 2020-01-19
weather@mockturtl automatic location not working at certain locations


Gr3q:
Issue:
Weird payload, here is the sanitized ipapi payload:
```
info t=2020-01-17T18:52:35Z weather@mockturtl#12: API full response: {
    "ip": "<REMOVED BY OWNER>",
    "city": <REMOVED BY OWNER>,
    "region": "California",
    "region_code": "CA",
    "country": "US",
    "country_code": "US",
    "country_code_iso3": "USA",
    "country_capital": "Washington",
    "country_tld": ".us",
    "country_name": "United States",
    "continent_code": "NA",
    "in_eu": false,
    "postal": "Sign up to access",
    "latitude": "Sign up to access",
    "longitude": "Sign up to access",
    "timezone": "<REMOVED BY OWNER>",
    "utc_offset": "-0800",
    "country_calling_code": "+1",
    "currency": "USD",
    "currency_name": "Dollar",
    "languages": "en-US,es-US,haw,fr",
    "country_area": 9629091.0,
    "country_population": 310232863.0,
    "message": "Please message us at ipapi.co/trial for full access",
    "asn": "<REMOVED BY OWNER>",
    "org": "Comcast Cable Communications, LLC"
} 
```

I will make the applet switch to a different provider.

Issue #2808, opened 2020-01-11
weather@mockturtl: "Cloudiness" missing colon and German translation


Kabuse:
Linux Mint 19.3 Tricia x86_64
Cinnamon 4.4.8
Weather (weather@mockturtl) Version 2.3.5 (2020-01-09 21:29:23)

If you open the applet, it shows you the current values in the upper right corner.

There is also the value "Cloudiness".

First issue, I want to mention is, that all other descriptors, like "Temperature" or "Humidity" are followed by a colon, but not "Cloudiness". I think this should be done, so it is more consistant.

Second issue, I am in Germany here, the translation of "Cloudiness" is missing. I would like to propose "Bewölkung" as the corresponding term in German language.

Thanks.

Gr3q:
Hello,

 I will look into it. In the meantime can you do a screenshot about the problem? as a reference for me.

Thanks

Kabuse:
Thank you for the reply, sure, no problem!

![weather_applet_screenshot](https://user-images.githubusercontent.com/13647006/72224349-db9c7400-3579-11ea-8f2a-a4c2ff3d3f38.png)

Kabuse:
Btw, what I just noticed. For the wind speed, the German translation for kph (km/h) is also not working. In the settings, I can choose "km/h", but in the applet, it still says "kph" :-)

![Bildschirmfoto vom 2020-01-12 20-40-20](https://user-images.githubusercontent.com/13647006/72224607-831aa600-357c-11ea-81c5-9e0969c19669.png)

Odyseus:
Hello, everybody.

@Gr3q: The problem is that the German localization needs to be updated. It contains the translation for the string `Cloudiness:` but not for the string `Cloudiness` (without colon). The retrieved string in [this call](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/weather%40mockturtl/files/weather%40mockturtl/3.8/applet.js#L537) is `Cloudiness`, which will not be translated to German.

As for the wind speed unit not being translated, the German translation does contain the localization, [but it isn't used by calling `_()`](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/weather%40mockturtl/files/weather%40mockturtl/3.8/applet.js#L533).

Gr3q:
[...]

PR #2817, opened 2020-01-19
weather@mockturtl Update 2.3.7

Gr3q:
Changes:
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2816, changed to different geolocation API
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2808
* Debug can be enabled by placing a file named DEBUG to the applet folder
* Added complete TS declaration files for some import modules (the easy ones...)
* Only import whats needed (hopefully it speeds loading the applet the first time, but I don't really know)

### DIFF
--- a/download-and-upload-speed@cardsurf/files/download-and-upload-speed@cardsurf/applet.js
+++ b/download-and-upload-speed@cardsurf/files/download-and-upload-speed@cardsurf/applet.js
@@ -89,6 +89,8 @@ MyApplet.prototype = {
         this.gui_sent_icon_filename = "";
         this.hover_popup_text_css = "";
         this.hover_popup_numbers_css = "";
+        this.minimum_bytes_received_to_display = 0;
+        this.minimum_bytes_sent_to_display = 0;
         this.bytes_start_time = AppletConstants.BytesStartTime.START_OF_CURRENT_SESSION;
 
         this.gui_speed = null;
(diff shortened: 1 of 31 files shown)


Gr3q:
What version of Cinnamon this style of importing came in, does anyone know?
```Javascript,
var {aaa, bbb} = imports.gi.etc;
```

Odyseus:
Hello, @Gr3q.

That's called [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) and it doesn't depend on a Cinnamon version, but in a SpiderMonkey version (Firefox). For all intent and purposes, you can use it in all Cinnamon versions in existence (from 3.0.x onwards).

Gr3q:
Hey @Odyseus 
Thanks, that's good news, meaning I can use it and still keep one codebase for both versions

Gr3q:
@brownsr Hold on merging this, its non-functional at the moment

Gr3q:
@brownsr All good now, tested as well on Cinnamon 3.0

brownsr:
What are the implications of these .ts files ?  Is there an implied dependency with versions of the glib etc. libraries that needs to be managed ?  They look generated to me with all that whitespace, is that right ?  if so what commands and scripts are being used - that will need some sort of basic documentation if so so others can make changes.

Regards

Simon

Gr3q:
So ok, I can put some documentation in there, but I'll explain now as well.

Purpose of the files:
These are Typescript declaration files (.d.ts), simply describing libraries already present in (my and probably every) system whats using Cinnamon. They help with static type checking and get offered in autocomplete, they are never transpiled into JS.

**They are optional.** I  - or anyone - could just as well remove them and declare ```imports: any``` but then I would have to look up documentation all the time. It gives me a head start how to use the libraries.

File versions:
Simply these versions were present on my system. I don't think the call signatures change that much, anyway its better to be explicit about those.  

File generation:
 I used my fork of [GIR2TS](https://github.com/Gr3q/GIR2TS), It generates Typescript declarations from from .gir files. They are included in the _gobject-introspection_ package.

It is far from perfect (Typescript can't handle overloaded functions for example), especially with libsoup and some other libraries, but it is much better to read these than the .gir files or the GTK documentation.  Wish I had time make it include the docstrings too.

PR #2828, opened 2020-01-24
Update bing-wallpaper@starcross.dev: recover image name

### LINKED ISSUE #2825: bing-wallpaper@starcross.dev is loosing wp-name
@Starcross 
Hi, I noticed that the WP-Applet is loosing the mouse-over name of the current picture when rebooting, logging off and in again or just restarting Cinnamon. After loading a new picture its name is loaded normally.

As I did'nt know how to save the name I tried a different solution: get the name again (only the name, not the picture and only if it's different from the default). That happens after the next loop (5 min.).
Do you mind if I provide a pull request? Of course you could decline or change it as you like.

Greetings

Mintulix:
Hi Alex,

this fixes #2825 

Two remarks:
- I personally set the loop to 10 minutes - a wallpaper that changes once a day isn't so important. But I don't know if that makes any tiny bit of difference; may be the 5 min. loop is "the same" related to CPU power, so I let it unchanged in the pull request.
- _done_ [GLib.get_user_special_dir(GLib.UserDirectory.DIRECTORY_PICTURES) doesn't work, at least in Linux Mint 19.3. The picture is saved in $HOME. I don't know how to fix that. Do you have a clue?]

### DIFF
--- a/bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
+++ b/bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
@@ -12,6 +12,8 @@ Soup.Session.prototype.add_feature.call(_httpSession, new Soup.ProxyResolverDefa
 
 const bingHost = 'https://www.bing.com';
 const bingRequestPath = '/HPImageArchive.aspx?format=js&idx=0&n=1&mbl=1&mkt=';
+const default_file_name = 'Bing Desktop Wallpaper';
+var file_name;
 
 function log(message) {
     if (logging) global.log(`[bing-wallpaper@starcross.dev]: ${message}`);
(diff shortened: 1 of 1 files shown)


brownsr:
pin @Starcross

claudiux:
@Mintulix I am not using this applet, but try to replace GLib.UserDirectory.DIRECTORY_PICTURES with its value: 4. I had the same problem with other GLib constants.

Mintulix:
Thanks @claudiux , but 4 doesn't work either (at least in German); may be it's locale/language specific?

Mintulix:
@claudiux @Starcross 
The problem with the picture directory seems to be MY problem: I'm not using real directories for pictures, music and videos, but soft links to other directories. GLib doesn't recognizes that. I tried it with "GLib.UserDirectory.DIRECTORY_DOCUMENTS" (which is a real directory in my $HOME) and it works (also with "1", so claudiux is right - "4" is related to the pictures directory).

So this should work for most users and is no global problem. I'll change it locally to use my own pictures directory.
No further questions ;-)

Starcross:
@brownsr Happy with this pull request thank you

Regarding the refresh time, I considered a user resuming from standby when a new image is available. They would have to wait up to this time until the new image is fetched; I though 5 minutes a reasonable maximum for this.

The user Pictures directory may not be the best location to store the image. Perhaps the ~/.cache directory would be more appropriate.

claudiux:
@Starcross @Mintulix Instead of checking the contents of a directory every 5 minutes, prefer to monitor it. Take a look at the `monitor_png_directory` method, starting at line 1490 of `4.2/applet.js` of the `SpicesUpdate@claudiux` applet.

brownsr:
Am assuming any further changes will be in a new PR

Regards

Simon

Issue #2835, opened 2020-01-26
weather@mockturtl minor language things


jorgenqv:
```
 * Applet version/Build date 2.3.7 (2020-01-21 00:16:39)
 * Cinnamon version 4.4.8
 * Distribution - Mint 19.3
 * Graphics hardware Intel Corporation Haswell-ULT Integrated Graphics Controller i915
 ```
@Gr3q

**Issue**
The text "Feels Like" is not translated to your chosen language in other languages than English.
The Configuration page also need some work regarding localization?

**Steps to reproduce**
Click on applet logged in with a language not English with DarkSky as provider and see Feels Like.

**Expected behaviour**
The text should be translated to your chosen language.

**Other information**
The workaround seems to be adding a colon character in the po-file to:
```
#: 3.0/darkSky.js:135 3.8/darkSky.js:87
msgid "Feels Like:"
```
Have not looked any further about the configuration page.

Gr3q:
Yes, it is my mistake changing the string a couple of times.

I moved the colon out from the string, it shouldn't really be there and I went through all the translation files this time to make sure all of them are correct.

It will be in the next update.

Issue #2831, opened 2020-01-25
weather@mockturtl applet is not unloading/reloading properly


Gr3q:
```
 * Applet version/Build date: Irrelevant, the bug persisted over multiple versions.
 * Cinnamon version: 4.4.8
 * Distribution - Arch
 * Graphics hardware *and* driver used
 ```
**Issue**
On Moving/Updating/Removing/Adding **any** applet on the Panel where the Weather applet is, there is a chance of the weather applet's main loop remain, this can cause the applet updating several times for van update interval, causing API call limits to saturate.

**Steps to reproduce**
1. Move/Update/Remove/Add any applet on the panel where the Weather applet is.
2. Wait and check the logs that the weather is updated multiple times over the set "update interval setting"
(I've added DEBUG more logs for my branch to observe my RefreshLoop function since then)

**Expected behaviour**
Only one of the applet's loop remain.

**Comment by Gr3q** [comment]
The problem is [here](https://github.com/linuxmint/cinnamon-spices-applets/blob/7d4f73ae3e55de257caa0c702c0d339993a9f089/weather%40mockturtl/src/applet.ts#L452). Multiple loops can remain and work in parallel. 

On my branch I switched to an async version of this function but it still relies on the same function so it behaves the same.

Any ideas what I can do do prevent this happening?

Some more info: applet ID is the same for all remaining loops.

What I came up with as a workaround:
So, multiple instances exist with the same instanceID and still work. I generate a uuid4 for the applet and save it to a global variable and in the applet class for each instanceID.

Every time the applet gets reloaded it saves that uuid4 to the global variable.
Then I check in the loop if the uuid4s match. If not, I stop the loop. I also have to use a global lock variable to prevent race conditions.

Log when 3 instances are running on top of each other:
```
info t=2020-01-25T19:22:56Z [weather@mockturtl#78]: Loop began 
 On Line: applet.js:334:17
info t=2020-01-25T19:22:56Z [weather@mockturtl#78]: 55ee3cab-07a0-4054-8f70-de90b6ec3281 
 On Line: applet.js:335:17
info t=2020-01-25T19:22:56Z [weather@mockturtl#78]: 8fd11675-f32d-42e8-a4c7-0e7d5d2192b6 
 On Line: applet.js:336:17
info t=2020-01-25T19:22:56Z [weather@mockturtl#78]: No need to update yet, skipping 
 On Line: applet.js:359:21
[...]

Issue #2305, opened 2019-02-07
openweathermap data not accurate for Australia


Normmron:
[weather@mockturtl] Include BBC for weather data

```
 * Applet version/Build date 2.1.4
 * Cinnamon version 3.8.9
 * Distribution - Mint 19.1

The weather data for Australia from OpenWeatherMap is not accurate, very different from the local weather forecasts.  Would it be possible to include the BBC weather for the data source - its forecasts are accurate for Australia.

much appreciated and thanks for your great applet, 

Norm

**Comment by Gr3q** [comment]
Hello,

I agree, BBC weather data is accurate, when they provide data.

For example, Cambridge did not have current weather data for the whole last weekend in their rss feed, and the often omit weather conditions from there too.

I'd like to include it but it may end up more unreliable than the others.

The other option is that I can fill up the holes from the forecast, but it won't be accurate then.

Bit more technical:

The best would be get the data from the HTML page (would need care often because webpages change constantly) but I can't find a JS HTML parser, or have hard time convert one to be usable.

Its really easy to do in python, but the only way I know to get the data back into JavaScript atm is write/read files, what I really don't want to do.

**Comment by Odyseus** [comment]
Hello, everybody.

@Gr3q

> ...but the only way I know to get the data back into JavaScript from Python atm is write/read files...

You could look at the Feeds Reader applet for [how does it calls a Python script](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/feeds%40jonbrettdev.wordpress.com/files/feeds%40jonbrettdev.wordpress.com/feedreader.js#L134) to get [the return data](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/feeds%40jonbrettdev.wordpress.com/files/feeds%40jonbrettdev.wordpress.com/getFeed.py#L96).

It basically does this:

```javascript
Util.spawn_async(["python-script.py", "arg1", "arg2"], (aStdout) => {
    let data = aStdout;
});
```

Even Cinnamon does it this way to open a Save/Open file dialog from JavaScript code (the [fileDialog.js](https://github.com/linuxmint/Cinnamon/blob/master/js/misc/fileDialog.js) module uses the [cinnamon-file-dialog Python script](https://github.com/linuxmint/Cinnamon/blob/master/files/usr/bin/cinnamon-file-dialog)).

[...]

Issue #780, opened 2017-05-28
weather@mockturtl: Request: Better Weatherdata


MegaV0lt:
weather@mockturtl

The actual data for the weather is very poor. Usualy up to 8 °C off odf thew real value. There is much better data from meteogroup (www.wetter24.de). 

Is there a way to get data from that source?
What about other sources?

NikoKrause:
You should mention the author on GitHub by typing @ plus his/her username to trigger a notification and bring his/her attention to an issue or pull request. You find the username of the author on the cinnamon spices website: https://cinnamon-spices.linuxmint.com/

jaszhix:
@mockturtl

SerafDosSantos:
@MegaV0lt @mockturtl http://openweathermap.org/ would maybe be a better fit. Weather is accurate to 1 degree for me. Weather data from Yahoo! is quite horrible in my opinion.

MegaV0lt:
Openweathermap does not even find my Town (Gelnhausen). Almost no data for germany

mockturtl:
Abstracting the data provider would be wonderful, and a lot of work.  I won't have time to do it myself.

Have a look at the [bbcwx@oak-wood.co.uk](https://github.com/linuxmint/cinnamon-spices-desklets/tree/master/bbcwx%40oak-wood.co.uk) desklet.  

http://www.oak-wood.co.uk/oss/other/bbcwx

PR #2836, opened 2020-01-27
Weather@mockturtl Update 2.4.0

Gr3q:
Changes:
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2835
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2831
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2305 and https://github.com/linuxmint/cinnamon-spices-applets/issues/780 I guess?
* Add Met Norway as a weather provider
* Add sunrise/sunset calculator to support Met Norway.

Not tested on Cinnamon 3.0 yet, I'll do it soon.


weather@mockturtl/files/weather@mockturtl/3.0/settings-schema.json
```diff
@@ -8,6 +8,22 @@
   , "description": "Set the keybinding to call the menu"
   , "default": "<Super><Shift>w"
   },
+  "descriptions":
```
Gr3q: I need an opinion on this section of the Settings, is it ok to do something like this?
It looks like this:
image

weather@mockturtl/src/met_norway.ts
```diff
@@ -0,0 +1,674 @@
...
+    }
+
+    private IsNight(): boolean {
+      if (!this.sunTimes) return false;
+      let now = new Date();
+      if (now < this.sunTimes.sunrise || now > this.sunTimes.sunset) return true;
+      return false;
+    }
+
+    private ResolveCondition(icon: string, checkIfNight?: boolean): Condition {
```
Gr3q: I need help on writing text for conditions, it has much to be desired at the moment and I don't want to edit much of the text later because of translations.

I can't go out in this state in my opinion.

weather@mockturtl/files/weather@mockturtl/3.0/settings-schema.json
brownsr: Looks fine to me

weather@mockturtl/src/met_norway.ts
brownsr: I am struggling to find the code fragment you've highlighted in the files.  That question mark looks like a mangled ternary comparison, but as I don't use .ts I may be missing something.  When I expand the files looking for it a) I'm not finding it b) it's taking a lot of time not to find it c) what I am finding looks reasonable.

Gr3q: The code above is shown but I wanted to select the code below, crap.

In typescript, for variables, paramaters and anything else you write types like this ```<varname>: <vartype>```, the ```?``` inditcates thet its an optional parameter.

Gr3q: I meant the cases inside the function:
image

(I'm unable to link it properly)

For each weather symbol (condition/icon, whatever) I'm need text what is not provided ```main and description fields```, so I have to write them based on the icon name, what is challenging because I don't know the intent behind some of them and English is not my first languge.

For example this one has a typo as well.

I contacted MET Norway for some explanation on the intent behind them, but they still need a rework.

Gr3q: They don't answer.

I resolve it for now, and I'll make changes once I come up with something better.

I want the update out because some of the fixes are actually urgent.

Gr3q:
Tested on Cinnamon 3.0, @brownsr it is ready to merge.

PR #2848, opened 2020-02-05
SpicesUpdate@claudiux: Update Coatian language

muzena:


brownsr:
your previous PR had languages for two applets, which is why I assume there is a conflict here.  Closing on the assumption this is not needed. I am sure you will re-raise if it's needed.

muzena:
@brownsr 
Can you open and merge this?
I don't know what I messed with first merge request, but translation for SpicesUpdate@claudiux was not updated see date in [hr.po](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/SpicesUpdate%40claudiux/files/SpicesUpdate%40claudiux/po/hr.po), old hr.po file.

PR #2854, opened 2020-02-11
Add Support for new Mint's sudo

Aminur670:
gksu is not supporting in newer mints. so changed gksu to sudo.


Aminur670:
thanks Linuxmint/Cinnamon-Spices-Applets

PR #2864, opened 2020-02-17
download-and-upload-speed@cardsurf Update 1.6.5

Gr3q:
@cardsurf 

Changes:
* Add symbolic icon option to settings
* Make symbolic icons the default (I thought they look better, but let me know if I need to change it back)
* Add new icons for displaying default symbolic setting.


Gr3q:
Looks like this
image

cardsurf:
@Gr3q 
Thanks for the effort you have put in the pull request. Having symbolic icons to choose from is a good addition.

I would prefer to make the symbolic icons optional by default. Please modify `settings-schema.json` and remove `Known issues` section from `README.md`.

In `applet.js` the `received_path` and `send_path` variables will remain non-empty after color icons are picked by a user with a `filechooser` dialog. We can use the `gui_symbolic_icon` flag to determine which icons to load like this:
```
...
let received_path = this.get_icon_path(this.gui_received_icon_filename, "arrow-pointing-down-symbolic.svg");
let sent_path = this.get_icon_path(this.gui_sent_icon_filename, "arrow-up-symbolic.svg");
...
get_icon_path: function(icon_filepath, symbolic_icon_name) {
    return this.gui_symbolic_icon ? 
        "~/.local/share/cinnamon/applets/download-and-upload-speed@cardsurf/icons/" + symbolic_icon_name :
	icon_filepath;
},
```

In `settings-schema.json` there is a tooltip for a new flag which states to add `-symbolic` suffix when using own icons. This can still be insufficient amount of information to figure out from it all the necessary steps to replace the default symbolic icons. Please change description/tooltip to shorter texts:
```
"description": "Show symbolic icons",
"tooltip": "Show symbolic icons instead of colored ones"
```

`Flaticon.com` website is very concrete about how to attribute an icon author as described in this [support article](https://support.flaticon.com/hc/en-us/articles/207248209-Attribution-How-when-and-where-). If the symbolic icons are available to a Free user please modify attribution in `credits.txt` to something as follows:
```
Symbolic icon up made by [author] from www.flaticon.com
Symbolic icon down made by [author] from www.flaticon.com
```

Gr3q:
@cardsurf 
I agree on all counts (I admit I only skimmed the Flaticon crediting page) and I will fix them as soon I have time. I will change the setting's description to something more specific from "Show symbolic icons" to something what conveys that this is a specific icon set and does not make the users custom icons symbolic.

There is one more thing I want to discuss:
> In settings-schema.json there is a tooltip for a new flag which states to add -symbolic suffix when using own icons. This can still be insufficient amount of information to figure out from it all the necessary steps to replace the default symbolic icons.

The tip might not be clear, first it should be a tooltip for the filepath choosers, second its aim is to inform that any icon can be displayed as symbolic as long as "-symbolic" is appended to its name (and should be svg I guess?). Nevertheless this information cannot be found anywhere and it is critical if someone wants to use custom icons.

cardsurf:
@Gr3q 
Thanks for the changes!

I understand now what you meant initially in the description text. After changing suffix of default icon names to `-symbolic.svg` the icons appear as symbolic in a panel:

icons

This is something unexpected and it is not mentioned anywhere at a screen.

My suggestion will be to modify tooltips of `Download icon` and `Upload icon` settings to include information about how to handle symbolic icons. A user needs to click on these to use custom symbolic icons. The tooltips can be displayed like the following:
```
An icon displayed next to download speed in the panel.\nTo display a symbolic icon make sure its name ends with '-symbolic.svg'."
```

When the new setting is enabled `arrow-pointing-down-symbolic.svg` and `arrow-up-symbolic.svg` will be used as icons so we can keep this setting as a shortcut. These icons will be used regardless of the ones selected in `Download icon` and `Upload icon` filechoosers. To avoid confusion that the new setting is responsible for enabling/disabling of usage of custom symbolic icons please change the description to:
```
"description" : "Show a default set of symbolic icons",
```

Gr3q:
@cardsurf it is done

cardsurf:
@Gr3q 
Great!

The last thing that I noticed is that tooltips for `Download icon` and `Upload Icon` settings are now the same. Please change `download` to `upload` in `Upload Icon` tooltip.

Gr3q:
Yes, sorry, I really should not copy-paste mindlessly.

cardsurf:
@Gr3q 
Great job with the PR!
I approve the changes.

PR #2869, opened 2020-02-19
jenkins@backuity.org: Jenkins Folders showed up as failure

mrbesen:
folders of Jenkins showed up as failure.
I removed them completely from the view. Not the best solution, but the easiest and fastest for now.
Its really annoying to see the read icon even when its all building correctly.


brownsr:
ping @vitaliis

vitaliis:
Hi guys, I no longer use this applet (many years now). You will have to find someone who will do changes and maintain it.

mrbesen:
I can do that, but i have no real experience with cinnamon applets or anything related.
But i guess better than nobody, right?

PR #2875, opened 2020-02-23
update german translation

to-ba:


Mintulix:
Hi @to-ba, I have no problem that you help translating this applet, but it has to be reasonable.
You changed Aktualisieren to Auffrischen, but update to Aktualisierung.
And you use Gewürze, Kästchen and Auffrischenknopf - really? Noone says Auffrischenknopf in German, checkbox is in common usage and spices is a proper name.
I can live with the rest of the changes, but that's a bit too much Germanization in my opinion.

to-ba:
Hi @Mintulix we should use the same words that are used in https://translations.launchpad.net/linuxmint/latest/+lang/de 
Auffrischen:
Bildschirmfoto vom 2020-02-24 15-15-15
https://translations.launchpad.net/linuxmint/latest/+pots/cinnamon/de/+translate?batch=10&show=all&search=refresh

--You changed Aktualisieren to Auffrischen, but update to Aktualisierung.
--And you use Gewürze, Kästchen and Auffrischenknopf - really?
Why not?
Checkbox do only computernerds understand. Do you think Ankreuzfeld is better than Kästchen?

Issue #2801, opened 2020-01-08
hwmonitor@sylfurd: New applet - I/O Disk usage


ArnaudDovi:
Actually for me hwmonitor@sylfurd is the perfect replacement to the system tray icons of process explorer under Windows except it lacks the informations to monitor I/O disk usage.

I believe it helps to not charge the system queue too much when there is extreme disk works.

@claudiux @brownsr

Hultan:
I have implemented two new graphs, showing disk read and write of the **root disk (/)**.

I hope you like it :-)

ArnaudDovi:
Thanks you a lot @Hultan 

Definitely the best hardware applet outta there, I use it at home and work everyday ;)

Keep up the good work

ArnaudDovi:
Hey @Hultan, is there any requirement to the disk usage to work ? because it does a few minutes I installed it and both graph are showing 0 usage

Hultan:
I am using **fusage** in **libgtop** (https://developer.gnome.org/libgtop/stable/libgtop-fsusage.html) to measure how many blocks have been read and written. I might have misunderstood how it works though.

Try copying a large file and see if you see anything. I noticed on my machine too that the disk wasn't used very much, but whenever I copy a large file I see the graph move. Maybe this graph should have the lin/log option too, like net in/out have?

Hultan:
And remember, that once the disk has reached its max performance once, that value will be used for the rest of the applets life time as a max value, so all percentages will be based on that max value. In the beginning it does not have that max value, so it will seem like it is being used more.

I implemented it like this, to avoid having to add a setting for disk speed, that I felt that most people don't know, or know how to find.

ArnaudDovi:
Not better, yet I'm trying on work computer with normal disk but still nothing moves even with a large file transfer.
![Screenshot from 2020-02-28 16-10-06](https://user-images.githubusercontent.com/52439/75560244-17659d00-5a45-11ea-9703-69d0d27ef596.png)

Hultan:
Are you sure you are copying that file from the root disk (/), to the root disk(/)?

I get the behavior that you describe if I copy a file from a secondary disk to a secondary disk. But every time I involve the root disk (/) in the process, I get reasonable graphs. If I copy from the root disk (/) to a secondary disk, I get some action in the DISK (read) graph, and vice versa.

[...]

PR #2881, opened 2020-02-27
hwmonitor@sylfurd : Version 1.1 - Disk read/write

Hultan:
Version 1.1 - Two new graphs, showing disk read and write.


Hultan:
Fixes #2801

PR #2886, opened 2020-03-01
hwmonitor@sylfurd : Version 1.1.1

Hultan:
Added a setting for mount dir


claudiux:
This works very well! I can now monitor my /home. Greats! Many thanks, @Hultan.

PR #2891, opened 2020-03-02
fix missing python(2) appearances in feeds spice

norbusan:
There are still two instances of Python version 2 in the applet.js code,
one of which is functional (calls `python` instead of `python3`), one is
only informational. Fix both.


brownsr:
ping @jake1164

norbusan:
Thanks for merging

PR #2895, opened 2020-03-07
Weather (weather@mockturtl) nl.po

French77:
updated Dutch translation Version 2.4.5 (2020-03-05 08:38:14)


FabienRCT:
Is it not a problem that the nl.po file is at the root and not in the files/weather@mockturtl/po subfolder while the old one is still here ?

PR #2899, opened 2020-03-08
temperature@fevimu - Added missing translations to German translation

OzzieIsaacs:
@claudiux: This PR adds the last missing strings (German.po file didn't have all strings from pot file) to the German translation of the temperature applet


claudiux:
@OzzieIsaacs Thank you !

PR #2930, opened 2020-03-27
Updated Icelandic translation of weather@mockturtl

sveinki:
Updated Icelandic translation of weather@mockturtl applet for Cinnamon.


brownsr:
this has carried in a translation for another applet, best to branch first as it's easy for this to happen if you work off master.  Should be harmless, so am merging regardless.

PR #2937, opened 2020-04-01
Fix missing images for Spotify Player

Rodrigo-Barros:
Hi I use linux mint as my distro, and recently I started writing a music script when I realized that the cover art of music on sound applet was missing, so I tried fix by myself, its a little hard for me understand the code but I did it, sorry if I open this pull request just update one line of code, my skills with Gobject are weak yet.
     This little fix just only update the url because spotify does not show cover art anymore by the old url.

### Additional notes: 
On old mode the artUrl could return a link like this for example: https://open.spotify.com/image/ab67616d00001e027ca89320e10bdc2bd343cdee and no image is returned.

Now with the change return something like this: https://i.scdn.co/image/ab67616d00001e027ca89320e10bdc2bd343cdee.

PR #2945, opened 2020-04-07
Fix missing images for Spotify Player

Rodrigo-Barros:
Hi I use linux mint as my distro, and recently I started writing a music script when I realized that the cover art of music on sound applet was missing, so I tried fix by myself, its a little hard for me understand the code but I did it, sorry if I open this pull request just update one line of code, my skills with Gobject are weak yet.
This little fix just only update the url because spotify does not show cover art anymore by the old url.

**Additional notes:**

On old mode the artUrl could return a link like this for example: https://open.spotify.com/image/ab67616d00001e027ca89320e10bdc2bd343cdee and no image is returned.

Now with the change return something like this: https://i.scdn.co/image/ab67616d00001e027ca89320e10bdc2bd343cdee.

Ref to [see](https://github.com/linuxmint/cinnamon-spices-applets/pull/2937#issuecomment-610103944) why open a new PR 
Spoiler: my mistake :(

This time I will try not blow up evereything.


brownsr:
ping @claudiux

PR #2950, opened 2020-04-13
First commit to add rust-menu applet

jerrywham:
This applet give user shortcuts to many Rust documentations


brownsr:
Installs with no icon visible in the applet, I suggest you take a look at other applets that use custom icons to see how they work.

PR #2961, opened 2020-04-19
weather@mockturtl Update 2.4.8

Gr3q:
Changes:

* Update Openweathermap to their unified API call
* Add information on DarkSky aquisition to Readme
* Fix translation issues in files other than applet.js
* Add Yahoo Weather to the available services
* Fix issue when system time is changed backwards, the applet stops updating until time gets to the time previously set
* Change 'Updated' to 'As of' to have a clearer meaning.

Tested on Cinnamon version 3.0 as well.

### DIFF
--- a/weather@mockturtl/CHANGELOG.md
+++ b/weather@mockturtl/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Update Openweathermap to their unified API call
 * Add DarkSky aquisition to Readme
 * Fix translation issues in files other than applet.js
+* Add Yahoo Weather to the available services
 
 ## 2.4.7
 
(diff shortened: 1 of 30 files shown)


muzena:
@Gr3q 
After update Openweathermap do not work anymore.

Gr3q:
@muzena Do you have any (Debug) logs you can share with me?

muzena:
@Gr3q 
Where can I found debug logs?

muzena:
I fix this by removing:
/.cinnamon/configs/weather@mockturtl/54.json

Gr3q:
@muzena Glad to hear that, it seems the settings become incompatible between the 2 versions (not that I changed them in a big way)

PR #2966, opened 2020-04-20
Add fr translation

jerrywham:


brownsr:
Still want this ignored ? the changes to the markbokil applets don't look as if they really belong.

brownsr:
I'm sorry, I can't merge this.  You have a PR which affects multiple applets, some not your own (though not adopted by anyone).  I have no issues with the changes that I see, but they should be separated out please. 

Can I refer you to the contribution guidelines, excerpt below:

One Pull Request - One Applet. This avoids the risk of having to revert changes to multiple applets if any problem is subsequently found with the PR. It also means that reviewing your changes is more straightforward.
Pull Requests that don't follow this format will be closed.

Regards

Simon

PR #2973, opened 2020-04-23
ShutdownApplet@shelley: Fixed broken Log Out command

amptrofa:
@icarter09

gnome-session-quit is not part of cinnamon-core, but cinnamon-session-quit is


icarter09:
@brownsr can you merge this in? Thanks.

PR #2988, opened 2020-04-29
timeshift-spy@nicog60: Add new applet Timeshift-Spy

NicoG60:
Hi everyone,

First contribution here, I've been working on a simple applet that allow the user to know the current state of Timeshift snapshot process.

It changes the icon depending on the state: 
- faded if idle
- plain if device mounted
- spinning if making a snapshot
 And it has a menu shortcut to create a new snapshot quickly.

Hope you guys will like it!


brownsr:
crashes if timeshift not installed, would be good if it handled that more gracefully.   Thanks very much for the new applet, merging now.

Regards

Simon

clefebvre:
Ah sorry, I arrive too late :)

I think it's really cool and it's something that we need, but as you said I think it should be cross-DE. With XappStatusIcon we can make something that works in all 3 DEs with symbolic/HiDPI support in the tray.

We've got this in the roadmap somewhere actually. Other than Timeshift, we also want to warn the user when automatic updates are being applied. There's cache update detection and xapp tray icon support in mintupdate already, I'm not sure whether it should be part of mintupdate or something else, but it's better off as a cross-DE thing than a Cinnamon applet imo.

NicoG60:
> crashes if timeshift not installed, would be good if it handled that more gracefully. Thanks very much for the new applet, merging now.
> 
> Regards
> 
> Simon

Cheers I'll have a closer look at that, I was pretty sure Timeshift was shipped anyway and didn't checked for it.



> Ah sorry, I arrive too late :)
> 
> I think it's really cool and it's something that we need, but as you said I think it should be cross-DE. With XappStatusIcon we can make something that works in all 3 DEs with symbolic/HiDPI support in the tray.
> 
> We've got this in the roadmap somewhere actually. Other than Timeshift, we also want to warn the user when automatic updates are being applied. There's cache update detection and xapp tray icon support in mintupdate already, I'm not sure whether it should be part of mintupdate or something else, but it's better off as a cross-DE thing than a Cinnamon applet imo.

Thanks for your answer. I think you're right, if a cross DE thing can be made that'd be great. I just worked on that both as little challenge and because I was tired of getting errors because of a snapshot running in the background.

clefebvre:
Thanks Nico, your code will help no matter what happens.

NikoKrause:
As long as we don't have a cross DE solution, it's nice to have something at least for Cinnamon.

NicoG60:
Cheers guys

> crashes if timeshift not installed, would be good if it handled that more gracefully. Thanks very much for the new applet, merging now.
> 
> Regards
> 
> Simon

Fixed in #2990

clefebvre:
Something worries me here... in most cases we'll be in WAIT_FOR_SNAP mode, with the volume mounted and we'll be running check_if_snapping() every single second.... with I/O read and CPU wakes. This is really bad for laptops.

I'm also worried about Cinnamon performance since it's single-threaded.

How about we optimize this by getting an event from timeshift itself? If it was running as user we'd simply add a XappStatusIcon in it.. job done. It's a bit more complicated than that since it's running as root (usually started by CRON). Maybe we can tell Timeshift to send a dbus signal on the system bus at the start and at the end of a snapshot? That way we don't need to check every single second, we just get told by dbus whenever something happens.

NicoG60:
Yeah I was thinking the same. I was working on using file monitors to get event from Gio instead of polling it manually. I'm still testing that...

mtwebster:
A dbus interface on timeshift would be ideal, then it's all just reponses, and easily made asynchronous.

In the meantime, I agree having a signal to rely on would be better than a timer - my timeshift drive is usually mounted for other reasons, so I'd be constantly polling.

Another thing is to try and use asynchronous operations wherever possible.
- The JSON library has Parser.load_from_stream_async (Open a GInputStream with gfile.read_async).
- Use gfile.enumerate_children_async
- Use Gio.file_test() instead of gfile.query_exists

 If this becomes something that is status-icon-based, the these things becomes less important since they'll be in their own process, but it's still good practice.

clefebvre:
Yeah, an outside process with XAppStatusIcon won't affect the Cinnamon perf, but you still harm battery power on the long run by doing I/O and CPU wakes. It doesn't really matter if you do it every 10 minutes or so... but for something to check stuff every second... hmm..

PR #2989, opened 2020-04-30
[hwmonitor@sylfurd]: Workaround for glibtop disk I/O for LVM bug (#2801)

### LINKED ISSUE #2801: hwmonitor@sylfurd: New applet - I/O Disk usage

jacobwills:
This is a work in progress. I'm submitting this primarily for review/feedback. By the time I got this far, I realized that **there is a significant desktop performance impact when performing large transfers** to or from the device being monitored although I'm not sure if it is my crappy JS coding or something inherently wrong with the approach I'm taking here. Maybe if I were using async calls when reading the sysfs kernel exports it would perform better? **My plan is to rework this using the libgudev library** which I believe utilizes non-blocking, asynchronous I/O system calls which might be the ticket to speeding this up. Feedback and additional performance testing results welcome. Thanks.

* Enumerates block devices by reading directly from sysfs kernel exports.
* Provides combobox for device selection in settings menu. (Thoughts?)
* Disk I/O detail displayed as **throughput** with unit scaling (B/s, KB/s, MB/s, etc.) mirroring NET stats implementation. (Thoughts?)
* Graph history is also scaled by keeping track of max throughput achieved within time frame based on the graph display width and update frequency.
* confirmed working to display statistics for both physical block and virtual block device mounting facilities including those auto-mounted with device mapper framework and across several filesystems. Gvfs, dm-crypt, LVM on LUKS, ext3/4, BTRFS, NTFS, FAT, etc. across SATA and USB devices that I had available for testing. (I haven't tested with any software/hardware RAID configurations but it should work depending on the way they are mounted)
* tested on latest Manjaro and Arch Cinnamon Desktop environments

Fixes #2801 
@Hultan @ArnaudDovi @claudiux @brownsr

### DIFF
--- a/hwmonitor@sylfurd/README.md
+++ b/hwmonitor@sylfurd/README.md
@@ -9,6 +9,12 @@ Issues can be reported here : [Issues](https://github.com/linuxmint/cinnamon-spi
 
 ### Changes
 
+**Version 1.3**:
+* **Bug Fixes and Feature Enhancements**: Disk Usage Stats
+    * Fixed a bug that prevented device mapper volumes (LVM, dm-crypt LUKS, etc.) from displaying usage statistics details or graphs.
+    * Disk usage graphs and details statistics are now displayed as throughput with unit and graph scaling (B/s, KB/s, MB/s, etc.).
+    * It is now possible to select the disk to monitor from a drop down menu in the settings UI that lists the available storage devices on the system.
+
 **Version 1.2.1**:
  * **New setting**, a new minor UI setting, you can now choose how to draw the grid lines in the graphs. You can choose between:
     * Don't draw grid lines at all
@@ -83,3 +89,4 @@ Vertical pane examples:
 *  giwhub
 *  andreevlex
 *  clefebvre
+*  jacobwills
--- a/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.2/applet.js
+++ b/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.2/applet.js
@@ -29,6 +29,7 @@ const Clutter = imports.gi.Clutter;
 const St = imports.gi.St;
 const Gettext = imports.gettext;
 const GLib = imports.gi.GLib;
+const Gio = imports.gi.Gio;
 const Settings = imports.ui.settings;
 const UUID = "hwmonitor@sylfurd";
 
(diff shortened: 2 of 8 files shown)


Hultan:
@jacobwills Thanks for this PR, here are a few suggestions though:

How about an extra setting where you can choose the old way (libgtop) or the new way (/sys/block/sdX/stat)? Since the old way does seem to work for most users, it might not be good to remove it right away, until we know that your way works even better for more users. The easiest way to do that would probably be to create two different disk-providers, and select the right one in applet.js depending on the user setting.

We probably need more than SDA,SDB and SDC though, personally I have a four disks in my computer, so potentially if I wanted to monitor the fourth disc I would need SDD. 

Can you tell the kernel which SDX you want to use, or is that chosen by the kernel and cannot be changed by the user in anyway? If the user can change that, we might need way more options...

About the performance issue you mentioned, have you tried different frequency settings, to see if that is the issue? Right now there is only one frequency setting for all graphs on the general tab...

I will take a deeper look into your code over this weekend.

jacobwills:
@Hultan 
I don't think it's unreasonable to have both options available, at least for some time to ensure something works for everyone. I'm thinking I could add a checkbox option to switch between the two disk-providers, leaving the current options as the default, and adding a label explaining to users that they can try switching to this new method if they have issues getting disk usage info.

---

> We probably need more than SDA,SDB and SDC though, personally I have a four disks in my computer, so potentially if I wanted to monitor the fourth disc I would need SDD.
> 
> Can you tell the kernel which SDX you want to use, or is that chosen by the kernel and cannot be changed by the user in anyway? If the user can change that, we might need way more options...

These settings entries are only place holders in this case, which seem to be required in order for the combobox key type to "support" options. In _applet.js_ the **getAvailableDisks** function populates these options with the available disks on the system, replacing those listed here in the default settings completely. If these placeholders aren't here we get an error saying that it doesn't support options when trying to write those new values. I have quite a few disks/partitions on my system as well which makes for a pretty solid demonstration which I've attached at the end. 😃

---

As for performance issues I've addressed these in subsequent commits by using GUdev instead of reading from the sysfs files directly. This seems to do the trick, supporting the narrowest update interval available to choose from here (2Hz) very well without any desktop stuttering as I was experiencing before.


Peek 2020-05-01 01-39

Hultan:
It seems to work fine on my machine, and I love the combo where you select the drive. That is awesome...

Would it be possible to add the size of the individual partitions? In my case I get sdd, sdd1 and sdd2, and having the size of the individual partitions would make it easier to identify which one is boot, efi, home etc. That might make the combo to wide though...?

Anyway, I am super happy with this, although I can't test it against LUKS/LVM since I don't use that. But if you say it works, I am happy with that...

BTW does this add an extra dependency? It worked fine on my Manjaro Cinnamon without adding any other packages, but for other distros?

I am starting to think that we don't need a setting for using the old way instead, what do you think?

A couple of minor things, increase the version number in metadata.json (1.3?) and add your name to the readme.md (contributing section), and also add a section describing whats new for this version.

jacobwills:
@Hultan 
Good! I'm glad you like it. 😄
I have added the volume capacities to the the combo display and refined the format a little.
You can actually make these comboboxes as wide as you want (the settings UI window width expands automatically) but it doesn't look pretty. Based on some experimentation with the different device/volume types and what information is available from each, I thought it looked best to limit the display to one of the following options in order of precedence:

- **ID_FS_LABEL** - via UNIX-mounts for formatted devices with FS labels _(i.e. "Media Storage")_
- **ID_MODEL** - fallback if no FS label reported by firmware data _(i.e. "ZTC-MS001-512G")_
- **DM_NAME** - fallback for device mapper (LMV, LUKS, etc.) _(i.e. "luks-7ccf3a9f..." or "vg_1-lv_2")_

---

> BTW does this add an extra dependency? It worked fine on my Manjaro Cinnamon without adding any other packages, but for other distros?

The `libgudev` appears to be required by Cinnamon, so it's expected this works for all Cinnamon Desktop users without any additional package installation or configuration changes.

---

> I am starting to think that we don't need a setting for using the old way instead, what do you think?

I think it should no longer be necessary to provide the `libgtop` implementation for rendering disk usage statistics, since these changes should cover everything for all Cinnamon users regardless of how the user decides to manage their storage devices. Additionally I think `libgtop` is poorly documented which makes it hard to know what works or how to use it in this project. It would still be relatively easy to revert back to using `libgtop` in the future if/when they sort out some issues with unsupported devices, etc. All of the other _Provider_ classes still depend on `libgtop` (except BATT) and while I'm certainly NOT suggesting that an effort to purge it from this project be made, I think it's possible to do; it's worth mentioning that it doesn't appear to be required by the Cinnamon packages that I looked at (probably why there is a warning if the import fails), making this applet just a little less user friendly. Anyway, that's my two cents.

---

I've appended my contributor info, added details to the change log, updated the version number to **1.3**, squashed a few commits to clean things up and marked this PR as ready for review.
Thanks for your review!

Hultan:
> Additionally I think libgtop is poorly documented 

I agree...

> All of the other Provider classes still depend on libgtop (except BATT) and while I'm certainly NOT suggesting that an effort to purge it from this project

Can libgudev give the information required for CPU, MEM and NET? I have personally no issues with switching to libgudev if that is better documented (do you have a link to the official docs?), when I switched to Linux, and started to make a few PR:s to this applet, a year or so ago, only CPU and MEM existed and they used libgtop. So I have just continued to use it...

jacobwills:
No, libgudev doesn't provide the functionality to replace `libgtop`. In fact I'm not so sure there is a good alternative API for the other stats we want to observe.

Here is the official libgudev docs: [libgudev](https://gitlab.gnome.org/GNOME/libgudev/-/tree/master/docs)

[Devhelp](https://wiki.gnome.org/Apps/Devhelp).
If you don't use this give it a try. It integrates with other GNOME dev tools like Glade.
Some packages include docs but you can try this on Manjaro to see those packaged separately:
`pacman -Ss ".*-doc$"`

[Valadocs](https://valadoc.org/index.htm) - These are by far the most comprehensive docs for GNOME APIs that I've found for explaining how things work, though not specific to Cinnamon's JavaScript bindings.

More Links:
[GNOME JavaScript Docs](https://gjs-docs.gnome.org/)
[GNOME Development Center](https://developer.gnome.org/)
[DevDocs](https://devdocs.io/)

PR #3027, opened 2020-05-25
New applet request: calendar@ccprog

ccprog:
Another calendar: Calendar with public Holidays. Obtains public holiday data from a webservice, http://kayaposoft.com/enrico/. (I have no connection to that service.)

This is a fork of the original Cinnamon calender, without any attempt to change design. All that happens is that holidays get the class `calendar-nonwork-day` and a tooltip with the name of the holiday.

Supported countries are hardcoded into the settings-schema. There are limitations from both ends that impede more dynamic approaches. This means I'll have to commit to following [updates](http://kayaposoft.com/enrico/release_notes.php) by the service provider.

Please review version requirements, as I have not taken much precautions for supporting old cinnamon/cjs versions. Do I need to tag this somehow?


brownsr:
Hi

Needs some attention to the file format.  It's missing the applet directory under files.  There should be a folder named calendar@ccprog under files, and then all the applet files under that rather than under files directly.  Have a check against, say, weather@mockturtl which is probably the most active applet.

Regards

Simon

ccprog:
oops

PR #3030, opened 2020-05-26
Cinnamenu@json - Piroritize non-fuzzy search over fuzzy, ignore case

Gr3q:
**_This PR is not ready_**

@fredcw 

I'd like to have a mixed-traditional-fuzzy search method in the applet, as the normal fuzzy is really inaccurate at times.

The commit has an example implementation for application entries prioritizing search the beginning of application names, then then searches inside the name, finally uses fuzzy on the remaining entries. It is faster and more accurate.

The question is, is there any preference how do you want this implemented? As an option in config, using it on the other type of entries etc (might need some refactoring in that case)?

I'm happy to work on it or just let you use the sample if you'd like to work on it yourself.


fredcw:
@Gr3q I took this applet over from the previous author about a month ago and tbh, I'm still trying to figure out how it works - I haven't even looked at the search algorithm yet. If you can improve it then please, be my guest! I'll be happy to approve any changes you make. I've taken a quick look at your changes and it definitely looks like a great improvement to me.

>how do you want this implemented? As an option in config,

I think the less options the better so as not to make things too complicated for the user is better, just the best search for the most people. My priority atm is to choose better defaults and simplify the user options.

>using it on the other type of entries 

Yeah, I think it would be better with other types and you're right, it needs refactoring because similar search code seems to be used in 6 different methods.

Gr3q:
@fredcw ok, I'll start work on it like so

fredcw:
@Gr3q Are you still working on this? I think the fuzzy.js algorithm is a bit pointless and it should just match substrings. Also, the markup tags (<b></b> etc) are unnecessary.

Gr3q:
@fredcw I started yesterday. It's definitely a huge amount of code and not that effective, but I would keep it in for now and write search modularly so it's easy to remove if it's deemed completely useless.

fredcw:
@Gr3q Thanks, but I've fixed it now.

PR #3034, opened 2020-05-31
Corrected URL included inside README

Simonmicro:


Simonmicro:
I don't know why the checks failed, but as you can see the only line I changed should not be relevant for that. Sooo, could you merge it anyways?

PR #3051, opened 2020-06-05
KDE Connect Applet

Severga:
Propose to add a KDE Connect applet for the Cinnamon. It is intended to be convenient, easy-to-use and Cinnamon-style.


brownsr:
Thank you for the applet.  It matches the file structure, loads up OK, and has nothing untoward, so I am happy to merge it.  However it would be good if it can also run in vertical panels.  More significantly I could not get it to work.  I'm not a KDE connect user, so I installed KDE connect and tried to get the applet to work, with my android phone paired with my laptop using the standard Mint blueberry applet.  Without any good result I am afraid.  I suspect that may be down to how to set up KDE connect rather than this applet.  But in any case I suspect you will need to give some much more explicit instructions if you want this to be effectively used by people.

Regards

Simon

Severga:
Hi! It is honour for me to contribute to Linux. And I'm happy it is successful. Especially because it is my first time of coding for Cinnamon, using GitHub and sharing something with others. I still know little, and do it like an amateur, so I can consider not everything. Thanks for the review and pointing out some moments for me. I will take them into consideration and try to explain more in README. The point is that some time ago I really was a Kubuntu user, so KDE is familiar to me and I could fail to take something into account thinking it was an evident thing. I really liked KDE Connect in Kubuntu, and it is great we can have it in Mint. KDE Connect must be installed in both Mint and your phone, it must be paired (request and accept) and the KDE Connect plugins must also be activated and set up in the settings both in your computer and your phone. In the applet, I placed the KDE Connect Settings item into the right-click popup menu. Pairing is also carried out in the KDEC Settings.

Thanks once more!

PR #3079, opened 2020-06-18
vnstat@linuxmint.com Version 1.0.3 Fix for Cinnamon 4.6

pdcurtis:
* Addition to cinnamon-version in metadata.json for Cinnamon 4.6 (Mint 20)


pdcurtis:
@brownsr That should be the last 4.6 fix commit from me!
  * I have noticed that none of them are reflected in updates in the applet 'store' as updates. When i go to the cinnamon spices web site the latest commits are reflected on the site text but again the actual download zips do not reflect those commits. So none of the changes are available to anybody testing the beta.
  * Is this delay normal? I have not noticed it before but I might have missed it as I normally use links from my local git repository on my test machines so they are always up to date, whilst I am now on a LiveUSB of 20/4.6.
Peter

Issue #2874, opened 2020-02-23
weather@mockturtl - TCP connection stays in CLOSE_WAIT


Mintulix:
@Gr3q 

Check the connection state with
lsof -i

During and after updating weather data the TCP state is ESTABLISHED, after a few seconds it changes to CLOSE_WAIT. So far ok, but it stays there "forever".

Shouldn't the connection be closed until the next refresh in x (e.g. 30) minutes?

Gr3q:
@Mintulix There are possibly 2 issues here, assuming you are talking about the connection ip-api.com.

First (and this is the main issue), it seems the connection to ip-api.com does not immediately closes on request completion, it just hangs there. 

Now this can be caused by the applet (maybe I use the call improperly, I will double check), the underlying library I'm using or their end, or the combination of some of this. I know the connection is closed properly when I curl to the url.

Second, on my system the connection times out after 1 minute and it is dropped. This means the behavior you see is at least specific to your system. But the route cause is the first issue.

I will look into further when I will have time.

Mintulix:
Thanks for the info.
I use the latest LM 19.3 with all updates and nearly no special installs.
There are only 2 connections hanging in CLOSE_WAIT - this weather applet and the weather desklet bbcwx@oak-wood.co.uk. All other connections are closed properly.
I don't know anything about TCP connections, just saw there is a constant download stream (~120 B/s) from these two spices.
If you think this should be changed/fixed, thanks in advance and take your time.

Gr3q:
@Mintulix The issue is closed, but please let me know if the issue is gone or still there.

Mintulix:
@Gr3q Works great, thanks a lot! I managed to provide your fix for the weather desklet bbcwx@oak-wood.co.uk also (https://github.com/linuxmint/cinnamon-spices-desklets/pull/556).

PR #3134, opened 2020-07-24
weather@mockturtl Update 2.6.3

Gr3q:
Changes:

* Add Met Office UK as a weather provider
* Add US National Weather Service as a weather provider
* Add support for entering manual location as an address
* During and after entering a location, the applet waits 3 seconds until user finishes typing then refreshes (prevents spamming)
* Applet locks itself during a refresh pass, any input during this period will trigger another refresh **after** the previous one finished
* Some of the custom icons were resized (to be bigger) to fit in panel and applet better
* 3.8 Settings is using tabbed layout, added help section with explanation for providers and entering manual location
* Add button for version 3.8 settings to submit issue
* Fixes issue where gray text is hard to read on light theme (grey text is changed to darker color in light themes)
* Distance unit is added to the settings, metric or imperial, precipitation, visibility and site distance uses it.
* Status bar shows observation site distance for site-based providers (Met Office UK and US National Weather Service)
* Improve handling API specific errors what need to be shown in UI (incorrect API key and such)

Fixes:
* Making sure calls and idle calls time out after 10 seconds so they don't lock the applet forever (hopefully fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/2874 as well)
* Fix issue when the applet would break if the locale is set to C

Tested with Cinnamon 3.0 as well.


brownsr:
That's a seriously chunky upgrade, good going !

PR #3155, opened 2020-07-30
Translated from POT

fusentast:
Danish translation for Cinnamenu Version 4.2.10
"da.po"


brownsr:
sorry, needs to go in the right place in the file hierarchy.  As it stands it is not even in the cinnamenu applet code.  

Regards

Simon

PR #3165, opened 2020-08-03
[mailnagapplet@ozderya.net]: Fix time difference error

AKorezin:
Fix time difference string error when time > 2 hours and < 1 day
Add more time constants


AKorezin:
The main problem is `time_diff / 60*60` it must look like `time_diff / (60*60)`.
Magic numbers + math error.

PR #3172, opened 2020-08-08
Screencapture+Recorder use DISPLAY for capture references, mpdecimate to reduce capture file sizes

octurite:
* use DISPLAY environment variable instead of \:0
  * resolves capture errors when DISPLAY is \:1 or other
* add mpdecimate as a video filter to significantly reduce recorded video sizes
  * mpdecimate only removes duplicate frames
  * mpdecimate has low overhead
* quote vars and convert parameters to an array to prevent potential whitespace issues


brownsr:
ping @icarter09

brownsr:
ping @icarter09

brownsr:
The directory name syntax is very cryptic, would prefer longer and more legible.  You might want to revisit that in a future change.

Regards

Simon

PR #3189, opened 2020-08-13
[New Applet] kdecapplet@joejoetv

JoeJoeTV:
This is an Applet for interacting with KDEConnect from the Cinnamon panel. It shows basic information about connected Devices and implements most of the features that KDEConnect offers for interacting with them. It uses DBus to connect to KDEConnect.
It is inspired by the KDE Connect Applet from Severga and the fact that the KDEConnect Indicator has not been updated for a longer time.

I am new to Cinnamon Spices, so I am not sure about version compatibility. What do I have to know to to make this Applet compatible with multiple versions? How can I check if it's compatible with an older version?


Gr3q:
> I am new to Cinnamon Spices, so I am not sure about version compatibility. What do I have to know to to make this Applet compatible with multiple versions? How can I check if it's compatible with an older version?

I usually fire up a VM and check everything is working, but I can share some of the info I gathered over time about JS compatibility.

Cinnamon under the hood uses ```mozjs``` package what is the Mozilla JS engine (to put it simply). It's version corresponds to Firefox's version, so you can check features on [caniuse](https://caniuse.com/).

I know that Cinnamon 3.0 and up uses ```mozjs24```, 3.6 uses ```mozjs38?```  (I think), and 3.8 uses ```mozjs52``` (what has classes and async/await). I don't know how it changes above because ES2017 support is perfectly fine for myself, whats in 52. _Note that not all distros follow this, for example Debian uses ```mozjs24``` for Cinnamon 3.6 as well, what sucks._ 

By looking at your code it's fairly compatible except the use of ``let`` and ```Array.includes()```, what puts a minimum requirement of ```mozjs52``` and Cinnamon version 3.8.

brownsr:
Thanks for the applet @JoeJoeTV .  

Regards

Simon

PR #3192, opened 2020-08-17
Advanced options for opening the menu on selected categories

erwn16:
@fredcw   
Users are now able to select a category which receives the focus when Cinnamenu is opened.
Valid options are: last opened category, favorites, recent, places, web bookmarks (when enabled).
This advancement makes sense because users may have different ways to access their objects.

Confusing: the comparisons of "applet.js" as well as of "settings-schema.json" show in Github more changes than I made. I derived both advanced versions of the files from my actual installation of Cinnamenu.

I'll do the rest of the work (update of POT-file and descripition of Cinnamenu) if my contribution is accepted.


fredcw:
@erwn16 Sorry, I just did another update PR #3193 an hour ago before I saw this PR. Also, this PR seems to be based on version 4.3.1 (#3184) but there's been another update since (#3190 two days ago.) So this PR can't be merged as is. I'll take a look at it now ...

fredcw:
@erwn16 At the moment, if someone selects recents, places or bookmarks as the startup category and that category isn't enabled then the menu will open with no category selected (it will be blank). While it can easily be fixed to select a default category, there's another problem - too many options. I don't like to add an option unless I think there is a high demand for it because it has the downside of making the options too complicated.

erwn16:
@fredcw One task of a menu is to provide an easy access to the work area of a user. I organised my menu that way for my daily work in the office and my experience is positive. I considered my proposal as a small step of such an approach.

The five options for selecting a category and the fact that three of them can be disabled is a minor flaw, but the combination of including Web bookmarks as category on one hand and setting the focus on it after the menu is opened on the other is a consequent concept which contains a significant value added for users. 

I agree with extending the options on a basis of a given demand, but was there really a demand for the usage of "session buttons box style"? What is its value added for me as user of Cinnamenu?

PR #3200, opened 2020-08-24
[kdecapplet@joejoetv] KDE Connect 1.14 Support

JoeJoeTV:
- Added support for new features in KDE Connect 1.4
  - SMS App
  - Photo Plugin
- Added multiversion support(1.3 and 1.4)
- Added more error messages


brownsr:
The conflicts need to be resolved before I can merge this.

Regards

Simon

JoeJoeTV:
Oh, ok, I'm still new to using pull requests on github and I also just noticed, that the first two commits shouldn't be in this pull request, because they were already in the first one. I just forked the repo and pushed commits to that fork and then I just created a pull request from that, but I don't know why these commits are in here. I know that this is probably not the right place to ask for this, but I'm not sure what I did wrong.

PR #3228, opened 2020-09-22
[Cinnamenu@json] Add option to reverse the favorite box elements order

bomdia:
I've added an option to reverse the elements order of the favorite box the default value is the previus behavior


brownsr:
ping @fredcw

fredcw:
@bomdia Hi, if we changed the order of the buttons automatically under the following circumstances it would obviate the need for the extra option in settings:

1) when the button placement is "top" or "bottom"
2) when the button placement is "left" or "right" and the applet is placed on the top panel

Also, the arrow keys no longer work as expected when the order is changed, that would have to be fixed as well.

brownsr:
I suspect that this PR isn't relevant now, after the applet owner's recent PR.  I'll close for that reason. Feel free to fix conflicts and reopen if I've not understood correctly

Regards

Simon

PR #3233, opened 2020-09-25
Fix crashing bug in Sensors@claudiux where there is no LANGUAGE environment 

rmonk:
If there's no LANGUAGE variable when running Sensors@claudiux, it causes a replacement on a null variable. This makes sure there is at least a blank string in the variable so the replace always "works"

### DIFF
--- a/Sensors@claudiux/files/Sensors@claudiux/applet.js
+++ b/Sensors@claudiux/files/Sensors@claudiux/applet.js
@@ -679,7 +679,7 @@ class SensorsApplet extends Applet.TextApplet {
       //_minimumIntegerDigits = 3;
     }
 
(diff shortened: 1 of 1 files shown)


brownsr:
Ping @claudiux

claudiux (commented):
Thank you. This is a good workaround. But I chose another way to solve this problem. See #3236. 

Mint has LC_NUMERIC, LANG and LANGUAGE environment variables.
Debian, Fedora and Arch have only the LANG one.

Regards.

brownsr:
Thanks for your PR, judging by the comment above it has helped the author to include a more comprehensive solution to missing environment variables.  Do keep the changes coming, they are much appreciated !  Closing this now.

Regards

Simon

PR #3299, opened 2020-11-07
Sensors@claudiux: Czech translation

Fotobob1:
Czech translation for Sensors monitor applet


claudiux:
Thank you, @Fotobob1!

Issue #3314, opened 2020-11-18
calendar@ccprog: UTC time zone missing in World Clocks


sphh:
```
 * Applet version/Build date: 1.1.3 (2020-11-16 20:54:10)
 * Cinnamon version: Cinnamon 4.6.7
 * Distribution: Linux Mint 20
 ```

**Notify author of applet**
@ccprog 

**Issue**
I want to add the UTC time zone to my World Clocks, but cannot find an appropriate entry. Where would I find it?

Thank you for your help!

**Comment by ccprog** [comment]
This is actually an upstream issue with [pytz](https://pypi.org/project/pytz/).

**Comment by sphh** [comment]
Thanks for looking into this.

I do not understand, why this is an upstream issue:
```python
>>> from pytz import common_timezones
>>> 'UTC' in all_timezones
True
```
It looks, like there are more timezones without a category (it's not a continent!) prepended:
```python
>>> from pytz import common_timezones
>>> [tz for tz in all_timezones if '/' not in tz]
['CET', 'CST6CDT', 'Cuba', 'EET', 'EST', 'EST5EDT', 'Egypt', 'Eire', 'GB', 'GB-Eire', 'GMT', 'GMT+0', 'GMT-0', 'GMT0', 'Greenwich', 'HST', 'Hongkong', 'Iceland', 'Iran', 'Israel', 'Jamaica', 'Japan', 'Kwajalein', 'Libya', 'MET', 'MST', 'MST7MDT', 'NZ', 'NZ-CHAT', 'Navajo', 'PRC', 'PST8PDT', 'Poland', 'Portugal', 'ROC', 'ROK', 'Singapore', 'Turkey', 'UCT', 'UTC', 'Universal', 'W-SU', 'WET', 'Zulu']
```
could they be summarized under the "continent" `Weird and wonderful` (or simply `Miscellaneous`? I just created PR #3322  implementing this. You might want to translate the variable `TZ_NO_CATEGORY` though.

sphh:
Implemented by PR #3322

PR #3322, opened 2020-11-19
Puts uncategorized timezones into the list of World Clocks

sphh:
Includes uncategorized timezones in the list of World Clocks. (Issue #3314)

### DIFF
--- a/calendar@ccprog/files/calendar@ccprog/settings_widgets.py
+++ b/calendar@ccprog/files/calendar@ccprog/settings_widgets.py
@@ -5,7 +5,7 @@ from xapp.SettingsWidgets import ComboBox, Entry
 import pytz
 from gi.repository import Gtk
 
-TZ_NO_CATEGORY = 'Miscellaneous'
+TZ_NO_REGION = 'Etc'
 
 def list_edit_factory(params):
     is_combo = 'options' in params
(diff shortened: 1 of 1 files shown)


brownsr:
ping @ccprog

ccprog:
Thank you for your work on this. This is the right direction.

There are actually more occurrences where you need to take care of string splitting: lines 57, 71 and 78. Currently, after setting the UTC timezone, the settings window won't open again.

I would propose a utility function `split_name`:

```python
def split_name(str):
    try:
        r, c = str.split('/', maxsplit=1)
    except ValueError:
        r = TZ_NO_CATEGORY
        c = str
    return r, c
```

ccprog:
You should propose the same sort of change to the Cinnamon Date and Time Settings dialog, which my settings code was inspired from. Having different lists for timezones in different dialogs would be confusing.

https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py

Especially, the "Miscellaneous" string should be the same in both use cases, should they accept the change.

sphh:
@ccprog: You are right, I missed the other occurrences of the splitting code (that's why I didn't make a dedicated method for it). Thanks for pointing that out. I also changed the `'Miscellaneous'` string to `'Etc'`, because that's what `pytz` uses.

@brownsr: What do you think of @ccprog's suggestion? If you also think, that's a good idea, I can create a PR along the same line as for this applet.

ccprog:
Resolves #3314. Please merge.

brownsr:
Feel free to propose the PR to the main cinnamon code.  It won't be me reviewing or merging though, so I can't say if it will be accepted.

Regards

Simon

Issue #3313, opened 2020-11-18
calendar@ccprog: Move "Date and Time Settings" to context menu


sphh:
```
 * Applet version/Build date: 1.1.3 (2020-11-16 20:54:10)
 * Cinnamon version: Cinnamon 4.6.7
 * Distribution: Linux Mint 20
 ```

**Notify author of applet**
@ccprog 

**Issue**
When displaying the calendar, there is an item `Date and Time Settings` between the calendar and the world clocks. This setting is not needed on a day to day basis. Would it be possible to move it into the context menu (right-click menu) as the World Clock Calendar does it? This would unclutter the display. Thanks.

Issue #3317, opened 2020-11-18
calendar@ccprog: Display holidays/weekends in bold even outside of the selected month


sphh:
```
 * Applet version/Build date: 1.1.3 (2020-11-16 20:54:10)
 * Cinnamon version: Cinnamon 4.6.7
 * Distribution: Linux Mint 20
 ```

**Notify author of applet**
@ccprog 

**Issue**
The calendar shows the days before and after the currently displayed month in dimmed numbers. These do not show the weekends and the public holidays in bold.

Would it be possible to mark these days in dimmed and bold?

PR #3324, opened 2020-11-20
Moves "Date and Time Settings" to the context menu

sphh:
Moves the "Date and Time Settings" menu item to the context menu to unclutter the calendar display.

The idea was shamelessly stolen from [Wold Clock Calendar](https://cinnamon-spices.linuxmint.com/applets/view/108).

See issue #3313.

### DIFF
--- a/calendar@ccprog/files/calendar@ccprog/applet.js
+++ b/calendar@ccprog/files/calendar@ccprog/applet.js
@@ -66,7 +66,7 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
             let item = new PopupMenu.PopupMenuItem(_("Date and Time Settings"));
             item.connect("activate", this._onLaunchSettings.bind(this));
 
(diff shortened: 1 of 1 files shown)


brownsr:
ping @ccprog

ccprog:
This PR is obsolete, the change has already been implemented, see https://github.com/ccprog/cinnamon-spices-applets/commit/9b7325ae069b975a3622f019b62a85f4a9aeda82

Patience, this will be included together with a solution for #3317

sphh:
I was not aware of that commit. So I am glad, that you like my proposal from #3313 and that you have it already implemented.

I saw that you also got rid of the separator between the calendar widget and the world clocks. IMHO it looks better with the separator still in place (may be make it conditional and only display it, if there is a world clock configured?) Thanks.

May I suggest that you cross-reference your commits to the issues so that these situations don't occur again.

I would also suggest to reject my PR from your side. Thanks.

Issue #3105, opened 2020-06-30
Pomodoro timer by pomodoro@gregfreeman.org unable to load in Cinnamon 4.6.6


ozcavic:
```
I very much like your Pomodoro applet and have used it for quite some time now. Yesterday I installed the new Linux Mint 20
(Ulyana) with Cinnamon 4.6.6. Of course, I installed your applet (Updated on 2019-10-25 16:22:15) too. My first clue that a
problem existed was that the usual icon and timer indication did not appear in the panel. Today when I booted up, in addition to
no icon/timer appearing in the panel, I received an error notification from the system saying that Cinnamon could not load the
applet and to contact the developer. Hence, this note. I hope you find this helpful as I may not be the only one who will experience
this problem.

I am also attaching the following file in the hope it is helpful in arriving at a solution...
[termbin.com-.pdf](https://github.com/linuxmint/cinnamon-spices-applets/files/4849261/termbin.com-.pdf)

myh98:
Running "sudo apt install python-is-python3" fixes it. Apparently there's a call to "python" which doesn't point to any version of python. 
Found the solution under this.
https://cinnamon-spices.linuxmint.com/applets/view/131

ozcavic:
Works like a charm.

Thank you
Dean Whitney

Naushikha:
> Running "sudo apt install python-is-python3" fixes it. Apparently there's a call to "python" which doesn't point to any version of python.

This worked for me as well, much thanks! :smile:

claudiux:
`sudo apt install python-is-python3` does uninstall virtualbox-6.1.

arohde69:
> Running "sudo apt install python-is-python3" fixes it. Apparently there's a call to "python" which doesn't point to any version of python.
> Found the solution under this.
> https://cinnamon-spices.linuxmint.com/applets/view/131

thanks a lot. Works fine for me.

NikoKrause:
The latest applet version shouldn't need python-is-python3 anymore.

arohde69:
> The latest applet version shouldn't need python-is-python3 anymore.

ok. thanks for this information....i got the latest version of the applet and removed python-is-python3
now it works....

Issue #2407, opened 2019-04-08
pomodoro@gregfreeman.org: Sounds not playing


jkirk:
**A few facts**
* Applet version/Build date (to be found in "About.." or in **"metadata.json"**)

"About..." does not show a version or build date. (I think, this should be fixed, too).

`metadata.json`:
```json
{
    "dangerous": false, 
    "uuid": "pomodoro@gregfreeman.org", 
    "last-edited": 1541487177, 
    "name": "Pomodoro Timer", 
    "description": "Great tool for boosted productivity!"
}% 
```
* Cinnamon version (cinnamon --version)
Cinnamon 3.2.7

 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
Debian/stretch

**Issue**
No sounds are played.

**Steps to reproduce**
* Sounds -> Play Break Sound -> ON
* Timer Sound File:  `/usr/share/sounds/freedesktop/stereo/alarm-clock-elapsed.oga`

**Expected behaviour**
The sound should be played after the short and/or(?) long break.

**Thank you**
Thank you, @gfreeau! I really like this applet and appreciate your work! I hope you find some time to further improve this applet!

JanaFlauschata:
I had the same issue on Linux Mint 19.1 and managed to solve it with the tips from the [Cinnamon Spices site](https://cinnamon-spices.linuxmint.com/applets/view/131). So `sudo apt install sox` did the trick.

somayn:
does not have sound. linux mint 19.2 64b
the tip from jana did not help. where is help to be found?

omvsj:
> I had the same issue on Linux Mint 19.1 and managed to solve it with the tips from the [Cinnamon Spices site](https://cinnamon-spices.linuxmint.com/applets/view/131). So `sudo apt install sox` did the trick.

It worked for me.

PR #3326, opened 2020-11-21
pomodoro@gregfreeman.org: small bug fixes and changes

NikoKrause:
@gfreeau 

* Removed settings from left click menu since it's accessible through rightclick menu.
* Fix icon setting: Show/Hide icon instead of destroying the icon box
* Fix "What is this?" button: Use distro-agnostic "xdg-open" instead of "gnome-open"
* Remove emptyspace before "!"
* Let translator decide, which website to open whe you click on "What is this?"
* When showing remaining time, you need to consider the correct singular/plural for the time value: 1 minute, but 2 minutes
      Some languages have more than 1 Singular and 1 Plural, so we use dngettext now, which takes care of this.
* In Mint the 'sox' package, which is required for the sound to work, is not installed by default. Add instructions to README
* Add information label to install settings, if 'sox' is not installed
* Add yellow pomodoro icon, if the applet is not running
* Add option for symbolic icons
* Use python3 explicitly

Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3105
Closes https://github.com/linuxmint/cinnamon-spices-applets/issues/2407


gfreeau:
Hey @NikoKrause This looks pretty good. This looks to be all backwards compatible, I want all LTS versions supported back to mint 18 at least.

It would be great if you could commit a few more things in this merge. One is updating the README about python, the applet expects `python` to be on the path, but this is not always true for newer mint versions. The workaround is to install `python-is-python3` or make a symlink.

As described by users here. https://cinnamon-spices.linuxmint.com/applets/view/131

To make it easier, maybe extend the current message `Great tool for boosted productivity!` and specifically mention sox and python there to help users. I've been meaning to do this.

NikoKrause:
Hey,

I can add the instruction to the README, but I think, it would be better to solve it explicitly for everyone:
- Change the shebang in `kill-looping-sound.py` from `#!/usr/bin/python` to `#!/usr/bin/python3`
- Change the `python` command in the `applet.js` file to `python3`

This should work without having to install `python-is-python3`.

For the `sox` package I suggest to check for `/usr/bin/sox` and show a `label` in the settings, if `sox` is missing. (<- not sure, if there is a `label` type in Mint 18).
grafik

Btw are you aware of the multiversion feature, to keep applets backwards compatible?

P.S.: I have two things, which still bug me a little bit, and another feature idea.
1. If you hide the time in the panel, so you see only the tomato, you can't see if the applet is running or not. My suggestion. Show a yellow tomato, if the applet is not running. (If you could send me a yellow version of the icon.png, I could add it to this pull request.)
2. The applet doesn't remember it's state and the time. If I restart Cinnamon, the applet turns off and resets the timer. When working on applets etc. I need to restart Cinnamon a lot, which makes this applet unusable during those programming times. Unfortunately I don't have much time during winter to work on this. Maybe you have an idea and time for that. If not, it's ok. Maybe I'll find sometime next year for that.
3. Most applets in Cinnamon use symbolic icons now. So I'd like to add on option to use symbolic icons in the panel instead of the coloured tomatoes. My suggestion for symbolic icons:
screenshot-cinnamon-2020-11-22-141507 screenshot-cinnamon-2020-11-22-141540 screenshot-cinnamon-2020-11-22-141615
There's even an option to colourize the symbolic icons to green and red as the battery applet does:
screenshot-cinnamon-2020-11-22-142745

gfreeau:
Hey, here's an icon you can use to try out your idea when hiding time: https://www.gregfreeman.io/files/icon-running.png


Unfortunately I don't have time to do the things you requested at the moment, but I agree with all of them such as saving the state in a simple file in case or reload.

For now, happy to merge all of this as long as it doesn't break compatibility for versions earlier than 20. The multiversion feature will be helpful. My only concern is simply changing to python3, older LTS versions don't have it as default I believe?

NikoKrause:
Thanks for the icon. I'll try it next weekend probably. Also need to check compatibility with Mint 18 then.

NikoKrause:
Ok, I implemented the changes. 
Everything seems to work fine in Mint 18 and Mint 19.3.

gfreeau:
Hey @NikoKrause I tested it on 18 LTS versions and it works great, so happy to get this merged.

Suggestion though, instead of

`Please install 'sox' and restart Cinnamon to use the sound feature!`

I would recommend

`Please install the packages 'sox' and `python3` then restart Cinnamon to use the sound feature!`

I'm pretty sure python3 is not a default package in older mint versions.

Otherwise great! thanks a lot for your work.

gfreeau:
After another look, might be better to duplicate the message completely for `python3` since I see you are checking if the package is installed before showing the message.

NikoKrause:
Python3 is installed by default in Mint 18

gfreeau:
Awesome, then this is good to merge, thanks a lot for taking the time and effort.

PR #3348, opened 2020-12-04
radio@driglu4it: Added mpris control

jonath92:
The radio applet currently uses the `moc` player. However the `moc` player doesn't support MPRIS at the moment, which means that you can't control the radio with the the sound applet or kdeconnect. Even though this  feature has been requested already [11 years ago](http://moc.daper.net/node/496) and some attempts to implement it, I couldn't find any easy way to use MOC with MPRIS on Linux Mint. Therefore I changed the player to  `mpv` which supports MPRIS in combination with [this](https://github.com/hoyon/mpv-mpris) plugin. In order to keep also the existing functionality for the volume control in combination with mpv I also needed the `playerctl` package. 

So basically  I made a hugh change to the radio applet and I can understand when @Driglu4it is not willing to merge it. In this case I can change this pull request as a pull request for a new Applet but I think it is more user friendly when there are not multiple Radio applets. The disadvantage is that the applet won't work anymore out of the box for current users after updating to this version. They would need to install `mpv` and `playerctl` which is however in the default apt repository and I also added Installation Instructions to the Readme file. It also should be mentioned that it is not that easy to get the current version  working as there are no installation instructions and it is not sufficient to just install the player as also mentioned in the comment section of the [applet site](https://cinnamon-spices.linuxmint.com/applets/view/297). To be honest I even haven't manage to get the current version running on Linux Mint 20.  

As a summary I made this changes: 
- Replaced the moc player with the mpv player and playerctl which adds MRPIS Suppport. The plugin needed for this is automatically downloaded to the  applet directory at initialization
- Stops all other media with MPRIS Support when starting a radio channel
- Added Installation Instruction to Readme
- Changed example  Radio Urls to valid Urls


brownsr:
ping @Driglu4it

Driglu4it:
Good afternoon. I have not been developing Radio++ for a very long time. Please contact @NikoKrause

NikoKrause:
Hello I'll have a look at your PR later.
It sounds interesting, but I'll have to test it first.

jonath92:
@NikoKrause did you already find time to have a look :raised_eyebrow:?

NikoKrause:
Hi,

yes I had time to look at it. I'll get back to you tomorrow at the latest.

NikoKrause:
@jonath92 
I tried your pull request. See my comments below:

* I use Mint 20 and with mpv it seems to work a lot better than with moc. At least with the radio stations I tried. I haven't tried with other Cinnamon/Mint versions. If we see that it causes regressions with lower versions, we can go multiversion with the applet later.
* There's a big no go in your changes though. It downloads the `mpv-mpris` plugin without the user noticing it and without the users consent. I added a dialog, which informs the user and lets the user decide:
screenshot-window-2020-12-13-123315
* I made a few other changes:
1. The current `moc` version does work in Mint 20. The problem is, that the default radio stations seem to be down except for Chillout. And even with Chillout you need to wait a few seconds till it starts. So I removed the broken streams from the default selection.
2. The applet checks, if `mpv` and `playerctl` are installed and notifies the user to install those.
3. I updated the translation template and the German translation.
4. I added a little button in the settings, which in my opinion improves the usability of this applet a lot.

Here's the applet with my changes on top of yours:
[radio@driglu4it.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/5684525/radio%40driglu4it.zip)

If you're fine with my changes, could you please inlcude my changes in your PR? I changed all files except the icons.

There's another small bug, which is no deal breaker though. When the radio is playing and you restart Cinnamon the radio still plays. That's great. The minor issue is, that the applet doesn't remember, which radio station is playing, i.e. the little dot is gone. You don't have to fix that though. Just something I noticed.

NikoKrause:
There's also another little thing.
I just noticed, that if I scroll over the radio applet, the sound volume changes independently from the volume in the sound applet.
Wouldn't it be better to have those two volume settings in sync?
Right now you could have set the volume in the sound applet on max and the sound in the radio applet scrolled down to zero. (Could happen without the user knowing or noticing and you'll think it's broken, because you don't have a visual representation of the volume in the radio applet.)
Another option would be to remove scrolling of the volume in the radio applet since we control the radio with the sound applet now.

jonath92:
Thanks for the feedback and the addition. The radio stream search engine is awesome. I had no ideas about this engine. I always wasted a lot of time to find the right URLs. Many thanks. Some more comments: 

> I just noticed, that if I scroll over the radio applet, the sound volume changes independently from the volume in the sound applet. Wouldn't it be better to have those two volume settings in sync?

Just for clarity we need to differentiate between the volume setting in the sound applet which is shown when clicking on the applet with the left mouse button (global volume) and the volume shown in the sound applet under the applications section which is shown when clicking on the applet with the right mouse button (application volume). It is not (easily) possible to sync the global volume with the radio applet but the application volume and the radio volume should be synced. This is the case when using kdeconnect or playerctl but not with the sound applet and therefore I think this is a Bug of the sound applet which I have reported in the cinnamon repository. 

>There's another small bug, which is no deal breaker though. When the radio is playing and you restart Cinnamon the radio still plays. That's great. The minor issue is, that the applet doesn't remember, which radio station is playing, i.e. the little dot is gone.

Yes that's true. Also the tooltip is not shown. I will fix this in future. Another issue is that special characters (e.g German Umlaute) are not shown correctly on the sound applet (when the option "Show song information on the panel" is activated). This is a limitation of the used MPRIS Plugin. I also have some ideas how to improve the applet. For transparency I added a category "Known Issues" and  "Ideas for improvements" in the Readme. I haven't made any changes to the files you have provided.

NikoKrause:
For now it's good as it is.
Not sure about the volume issue. I'm not very familiar with sound setting or the sound applets code.

I just checked the log and it shows two error in the radio applet.

The first error:
```
(cinnamon:17136): Cjs-WARNING **: 20:02:40.708: JS ERROR: Exception in callback for signal: activate: TypeError: this.activeMenuItemChanged is not a function
anonymous/MyApplet.prototype.on_tree_changed/<@/home/niko/.local/share/cinnamon/applets/radio@driglu4it/applet.js:124:15
```
The second error:
```
(cinnamon:17136): Cjs-WARNING **: 20:04:18.899: JS ERROR: ReferenceError: currentMenuItem is not defined
anonymous/MyApplet.prototype.changeCurrentMenuItem@/home/niko/.local/share/cinnamon/applets/radio@driglu4it/applet.js:159:31
```
There's no hurry to fix those right now, though. Just an info for you to check the logs once in a while. :)

PR #3379, opened 2020-12-21
radio@driglu4it: Quick fix for Cinnamon 3.0+

NikoKrause:
Add multiversion to this applet:
* Cinnamon versions 4.6+ will use the new mpv radio applet
* Cinnamon versions 3.0 till 4.4 will use the old moc radio applet

Add second default radio station "Mint" ;)

The radio stream URL search engine sometimes seems to don't search at
all. Therefore I added a second button, which opens a website, which lists
URLs for radio stations. The website is in German, but should be self
explanatory without having to know German.


NikoKrause:
@jonath92 @Driglu4it fyi

PR #3386, opened 2020-12-23
tzclock@lhriley: new applet

lhriley:
This PR is for an applet for displaying a simple clock for a specified timezone.

I've found that while working, I often want a UTC clock in addition to my local timezone clock. Unfortunately, in the years that I've been using Cinnamon, I've yet to find what I'm looking for. This solves my personal use case, so I thought I'd share it back to the community.

Since this is my first applet contribution, I welcome any feedback / suggestions for improvements.

For attribution, the work is derived from a couple sources:
- https://github.com/linuxmint/cinnamon/blob/92b5c87c937212a96bd2e367df5f3c3d15b19e8b/files/usr/share/cinnamon/cinnamon-settings/modules/cs_calendar.py
- https://github.com/linuxmint/cinnamon-spices-applets/tree/d172c5008c324212e258952964f28e3607bc90f5/calendar%40simonwiles.net


brownsr:
I'm happy to pull this when it works.  It fails to load for me, with these errors:

Cjs-Message: 16:46:06.482: JS LOG: [LookingGlass/error] time.to_timezone(...).format(...) is null
Cjs-Message: 16:46:06.482: JS LOG: [LookingGlass/trace] 
<----------------
_get_world_time@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:194:37
_updateClockAndDate@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:158:32
on_settings_changed@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:111:14
_init@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:90:18
MyApplet@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:42:10
main@/home/simon/.local/share/cinnamon/applets/tzclock@lhriley/applet.js:213:20
createApplet@/usr/share/cinnamon/js/ui/appletManager.js:582:25
addAppletToPanels@/usr/share/cinnamon/js/ui/appletManager.js:357:34
finishExtensionLoad@/usr/share/cinnamon/js/ui/appletManager.js:95:14
_init/<@/usr/share/cinnamon/js/ui/extension.js:274:32

lhriley:
@brownsr thanks for taking the time to test and review.

The fact that you are seeing errors is very odd. I don't see these errors in LookingGlass, and I was using LookingGlass the entire time I was writing / testing.

Can you help me understand what might be different in our setups which might cause these errors?

I'm running Linux Mint 20 with Cinnamon with all of the latest packages installed. I don't think I've gone out of my way to do anything special with either.

Also, I'm only just passable at writing JavaScript, so maybe there's something obvious that I'm unaware of.

brownsr:
Thanks for your nice comments, but reviewing these is my commitment to the project.  I review everything to ensure there's no embedded malicious code, that the file structure is right etc.  And for new applets I want to ensure that they seem to work.

these errors are straight out of the .xsession-errors file in the home directory.  

As it happens I've been able to pin down the error for you.   Your applet wants to be able to cope with all the options the user might select in the date and time system settings.  If I choose to display seconds your applet works, if I turn the seconds display off then it fails with the error above.  

Regards

Simon

lhriley:
@brownsr thanks for setting me down the right path. I wasn't looking at the `.xsession-errors` log, so I just wasn't getting a full picture of the errors.

I believe there was some weird interaction with the system clock settings that I don't fully understand. I broke these settings out into the applet itself, which should be more customizable in the long run anyway.

Some weirdness that made this difficult to debug was that Cinnamon seems to have kept a copy of the applet in memory between using the reload applet. I was forced to restart Cinnamon to clear out the old copy of the code. It also seems that `LookingGlass` needs to be closed / opened after restarting Cinnamon. Not really sure what is happening with either of these quirks, but it was confusing.

In any case, I pushed up a new commit and I don't see any issues on my end when toggling through the various formatting function toggles.

brownsr:
Merging now.  Personally I'd recommend using the system date settings, as having two places where it can be set can cause confusion, but no biggie.

Regards

Simon

PR #3390, opened 2020-12-25
radio@driglu4it: Minor Bug fixes, readme addition and code clean up

jonath92:
I refactored the code for cinnamon 4.6 a little bit to make it simpler to add new features in future. In particular I made some functions shorter, added more comments and renamed some variables. So I hope the code is now a little bit easier to read. Also I fixed the minor bugs: 
- The current channel is now correctly indicated with a point, the icon changes and the channel is now shown in the tooltip when refreshing cinnamon
- The color of the symbolic icon is now changed immediately after changing the color in the applet settings. Previously cinnamon had to be restarted to apply the changes
- There are no errors in the log anymore

I also updated the readme. 

Merry Christmas @NikoKrause


NikoKrause:
Merry Christmas!

PR #3395, opened 2020-12-29
nightscout: Bug fix - Handled IOB not defined

inventor96:
When a user enables displaying the IOB, but one of the (sub-)properties isn't defined in `status.pump.iob.bolusiob`, the code silently stops execution and fails to update the applet label.

@ImmRanneft

### DIFF
--- a/nightscout@ranneft/files/nightscout@ranneft/applet.js
+++ b/nightscout@ranneft/files/nightscout@ranneft/applet.js
@@ -17,6 +17,7 @@ const log = function(message) {
 }
 
 const makeHttpRequest = function(method, uri, cb) {
+  uri = uri.replace(/([^:])\/{2,}/, '$1/');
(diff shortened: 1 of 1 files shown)


brownsr:
Will that log message be called a lot ?  I'd be happy to merge this if not for that - large log files can tend to cause their own problems.

Regards

Simon

inventor96:
Per line 13 in that file, logging is disabled by default. I just found it helpful when debugging this another another issue to follow, so I included it in hopes it will help another dev in the future.

https://github.com/linuxmint/cinnamon-spices-applets/blob/dbe3ba5f49c4f91958bd3121a872c680049072cd/nightscout%40ranneft/files/nightscout%40ranneft/applet.js#L13

inventor96:
@brownsr After I verified the solution to the second problem, I figured it should just be lumped into this PR since it's so simple.

PR #3399, opened 2020-12-31
nextslide@yaya-cout: New applet to change background ...

Yaya-Cout:


JosephMcc:
I'm curious why this instead of the default one that comes with Cinnamon?

Yaya-Cout:
I have created this because it's more convenient when I have the need to alter often wallpaper than going to the menu to change wallpaper

PR #3403, opened 2021-01-02
[radio@driglu4it] Add volume settings, change mpris plugin path, remove some notifications, add error message

jonath92:
I made some changes to the radio applet: 

- Previously the mpris plugin was saved to the applet folder (`~/.local/share/cinnamon/applets/radio@driglu4it`). This however was a bad idea as the plugin has been deleted on each update and after each update the user had to confirm a download dialog. Therefore the directory of the mpris plugin has been changed to the config directory (`/.cinnamon/configs/radio@driglu4it`). So after this download the user has to confirm the dialog only one last time
- In the comments of the applet somebody has been criticized that the radio is to quiet at the start. Therefore it has been added now an option for setting the initial volume. In this regard also an option for the max volume has been set which previously has been missing and could lead to scratching noises. 
- In the comments of the applet it has been also criticized from two users that the notifications of the current playing radio channel and the notification "radio stopping" when clicking on stop are annoying and useless. Therefore these notifications has been removed. 
- Previously the user hasn't been notified when a radio couldn't start. Therefore the user couldn't know if the stream is still loading or if an error occurs. Therefore it is shown now an notification when an error occurs. 
- Also updated the German Translation.


jonath92:
Forgot to ping @NikoKrause

PR #3430, opened 2021-01-12
[clean-show-desktop]Now is possible drag and drop files to copy to desktop, the css property width has been replaced by padding and the screenshot was also changed

FilipeTorresBR:


brownsr:
ping @FilipeTorresBR

PR #3437, opened 2021-01-13
[radio@driglu4it] Remove Playerctl

jonath92:
Currently we use playerctl for stopping the radio and increasing/decreasing the volume. This has several disadvantages - users have to install the package and it is very difficult to trigger actions when the volume is changed or the radio player paused from outside of the applet (e.g the sound applet). This pull request removes the playerctl dependency by using the D-Bus interface instead (similar to the sound applet). 

@NikoKrause

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/4.6/applet.js
+++ b/radio@driglu4it/files/radio@driglu4it/4.6/applet.js
@@ -1,15 +1,15 @@
 const { TextIconApplet, AllowedLayout, AppletPopupMenu } = imports.ui.applet;
 const { PopupMenuManager, PopupSeparatorMenuItem, PopupMenuItem } = imports.ui.popupMenu;
 const { AppletSettings } = imports.ui.settings;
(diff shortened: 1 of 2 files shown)


NikoKrause:
@Driglu4it Do you mind if I add @jonath92 as current maintainer of your radio applet? You don't have to, if you don't want to, though.

Driglu4it:
> @Driglu4it Do you mind if I add @jonath92 as current maintainer of your radio applet? You don't have to, if you don't want to, though.

I completely agree, because moved on to developing plasmoids for Plasma 5. What do I need to do to transfer rights?

NikoKrause:
I'll take care of it. Thx.

Driglu4it:
> I'll take care of it. Thx.

It's my pleasure. I am glad that development will continue. Thank you!

jonath92:
I am very thankful for this honour and for your contributions @Driglu4it  and @NikoKrause!

PR #3439, opened 2021-01-13
Add world clocks to tooltip.

### LINKED ISSUE #3315: calendar@ccprog: Display World Clocks in tooltip
```
 * Applet version/Build date: 1.1.3 (2020-11-16 20:54:10)
 * Cinnamon version: Cinnamon 4.6.7
 * Distribution: Linux Mint 20
 ```

**Notify author of applet**
@ccprog 

**Issue**
Currently the tooltip just displays the date.

Would it be possible to add the configured World Clocks to the tooltip? (May be make this configurable …)

Maybe some information about the holiday (if today is a holiday or the next due) can also be added?

**Steps to reproduce**
Move the mouse cursor over the applet. The date is displayed.

**Expected behaviour**
Display the date and the time of the configured world clocks.

sphh:
This adds the world clocks to the tooltip (closes #3315).

It also adds an option to display them in the tooltip or not.

It would be preferable to be able to decide on a per-clock basis, if the world clock should be displayed in the tooltip by adding a third column to the `World Clocks > Show World Times` table, but I could not figure out, how to do it.

There might be a more elegant way to achieve this …

### DIFF
--- a/calendar@ccprog/files/calendar@ccprog/applet.js
+++ b/calendar@ccprog/files/calendar@ccprog/applet.js
@@ -72,6 +72,7 @@ class CinnamonCalendarApplet extends Applet.TextApplet {
             this.settings.bind("use-custom-format", "use_custom_format", this._onSettingsChanged);
             this.settings.bind("custom-format", "custom_format", this._onSettingsChanged);
             this.settings.bind("keyOpen", "keyOpen", this._setKeybinding);
+            this.settings.bind("show-worldclocks-in-tooltip", "show_worldclocks_in_tooltip", this._onSettingsChanged)
             this._setKeybinding();
 
             /* FIXME: Add gobject properties to the WallClock class to allow easier access from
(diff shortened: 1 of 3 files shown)


sphh:
@ccprog

ccprog:
Sorry, I am not approving this PR at this time. My interest is in keeping the applet and its interface simple, and I don't see how this change would add more benefits than the disadvantage of making the main tooltip not easily readable at one glance.

As for the additional setting, my rule of thumb here is that configuration options should be kept sparse, and only be added if they change or add functionality. This one only changes grafical representation.

That said, I'll keep issue #3315 open for now as a chance to think more about this.

sphh:
As your very own definition it adds functionality: Adding the world clocks to the tooltip, so that you can quickly check the time in another place. (And for those people who likes the "simple" interface it can be switched off – a win-win situation.)

PR #3452, opened 2021-01-17
download-and-upload-speed@cardsurf: Fix toString for ByteArrays

NikoKrause:
@cardsurf Your applet throws a lot of warnings in Cinnamon 4.8:

```
(cinnamon:2012143): Cjs-WARNING **: 13:26:27.848: Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to explicitly call ByteArray.toString(array). (Note that array.toString() may have been called implicitly.)

ByteArray.toString(array) is not backward compatible with previous (< 4.8) versions of Cinnamon. You will have add versioning support to your applet if it does not already have it, so that compatibility can be maintained for users on older Cinnamon versions.
```

I've tried to fix them, but you probably want to test, if I did it right?

### DIFF
--- a/download-and-upload-speed@cardsurf/files/download-and-upload-speed@cardsurf/files.js
+++ b/download-and-upload-speed@cardsurf/files/download-and-upload-speed@cardsurf/files.js
@@ -2,6 +2,7 @@
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
 const Gettext = imports.gettext;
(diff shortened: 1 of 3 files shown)


NikoKrause:
@cardsurf Your other applets throw those errors as well. You might want to fix them as well?

PR #3465, opened 2021-01-24
[bing-wallpaper@starcross.dev] Fixed a bug where the symbolic icon wasn't respecting theme colors.

AndreiMiculita:


AndreiMiculita:
@Starcross You might want to take a quick look at this

Starcross (changes requested):
Thanks for this. I can see the icon is taking on the colour of the theme now.

The only problem is the icon appears a bit smaller now. I attempted to compensate for this by reducing the document size to the object (attached). I notice it must be square otherwise the icon is stretched! Perhaps this is the largest it can be now we are using a symbolic icon - just a little inconsistency in the way Cinnamon renders icons?



[bing-wallpaper-symbolic.svg.gz](https://github.com/linuxmint/cinnamon-spices-applets/files/5862097/bing-wallpaper-symbolic.svg.gz)

AndreiMiculita:
Hmm that's weird 🤔

So in the initial icon, there was an empty transparent square shape in the background, and 2 identical mountain icon paths overlaid on each other. When Cinnamon tried to color it, the square was also getting colored, and thus the whole icon was just a square.

In 4cacd2c, I removed the square, along with one of the mountain paths.

I now realized the mountain path I removed was much simpler while looking virtually identical; so I put it back in (in b57ed2bdf523ccc4b8da56325d2862d7f6c37368).

Could you check whether it's still too small? I couldn't reproduce the issue. 

If that's the case, we can crop it, as you suggested (or upscale the path, to keep the image size 24×24).

Starcross:
Ok the shrinking only seems to occur when set_applet_icon_symbolic_name is used. All versions of the svg have the same behaviour for me, including your latest

Could it be that Cinnamon is enforcing a square icon for symbolic, but not for the old set_applet_icon_name? 

In case it's significant, I'm using LM 20, Cinnamon 4.6.7

icon-change

AndreiMiculita:
I think I found a possible reason. With the old set_applet_icon_name, the size changes with "Colored icon size". With set_applet_icon_symbolic_name, it's controlled with "Symbolic icon size". If these are different, it might be the reason why the size wasn't consistent.
Screenshot from 2021-01-24 17-46-38

In any case, since it's still too small, I've upscaled the icon path to cover the entire document (e1d7d6b).

Starcross:
Yes! That's the reason

The defaults are 24px coloured icon and 16px symbolic - not sure why. I just checked this on a 20.1 livecd.

Starcross (approved):

Issue #3488, opened 2021-01-31
weather@mockturtl 2.7.0 not acknowledging Climacell API key


garuffalo:
weather@mockturtl 2.7.0 shows "incorrect API key" when switching to Climacell

```
 * Version 2.7.0 (2021-01-30 03:52:53)
 * Cinnamon Version 4.4.8
 * Distribution - Linux Mint 19.3 Cinnamon (Tricia)
 * Graphics hardware - Intel Mobile GM965/GL960 Integrated Graphics (driver: i915)
 * Snippet or attachment of ~/.cinnamon/glass.log
 * Snippet or attachment of ~/.xsession-errors

 ```

**Notify author of applet**
@mockturtl 

**Issue**
I signed up for a developer account at your link, https://developer.climacell.co/sign-up to obtain an API key. Sign-up appeared to go OK and took me to an account page with the "Secret Key". I have previously been using Dark Sky with no issues, but wanted to try another API as Dark Sky will cease to work at the end of this year. When selecting Climacell in the configure window and providing the API key I was given, the applet declared "incorrect API key".

**Steps to reproduce**
Switched back and forth between Dark Sky and Climacell several times with same failed result with Climacell every time. Restarting Cinnamon or rebooting the computer did nothing.

**Expected behaviour**
With correct API keys, one should be able to switch between different data providers.

**Other information**
I received a welcome email from Climacell and was able to log in to my account through a link in the email. However, when I attempt to log in at https://developer.climacell.co/sign-in, login fails. Even the "forgot password" link doesn't work.

If I go to to the homepage at https://www.climacell.co/ I can log in to Weather API v4 at https://app.climacell.co/signin - but not Weather API v3. API v3 has a white themed interface, and appears to be the one used at https://developer.climacell.co/sign-in, whereas API v4, at https://app.climacell.co/signin uses a dark theme.

Is it possible that the Weather applet is set to use API v3, but new signups are for API v4?

Gr3q:
You are right, they block registrations for v3 and redirect to v4. The applet uses v3, so i have to specify that, and create a new provider for v4 to make transition smooth, the remove v3 later this year, as do they

Gr3q:
I added the v4 version of Climacell so when 3.0.1 becomes available you can try it out

Issue #3486, opened 2021-01-30
weather@mockturtl V3.0.0


French77:
```
 * Versie 3.0.0 (2021-01-30 09:52:53)
 * Cinnamon version (4.8.6)
 * Distribution - (Mint 20.1)
 
 
 

 ```

**Notify author of applet**
@Gr3q

**Issue**
Wind direction icon does not show the correct current direction

**Screenshots**
#### Direction icon applet
![icon wind direction](https://user-images.githubusercontent.com/3099270/106361042-15265580-631c-11eb-992d-2082efabb80c.png)
#### Wind direction today
![wind richting](https://user-images.githubusercontent.com/3099270/106360740-88c76300-631a-11eb-96e9-3cb618730310.png)
#### Wind direction icon wttr.in
![wttr in](https://user-images.githubusercontent.com/3099270/106361062-2ff8ca00-631c-11eb-8626-722dcf643d94.png)

Gr3q:
I'll have a look, maybe I switched up the icons.

 Also, what theme are you using? The icon really shouldn't be black, I'll have to sort that out as well

French77:
> Also, what theme are you using? The icon really shouldn't be black, I'll have to sort that out as well

Cinnamox-Kashmir-Blue

Gr3q:
Oh yeah, obviously east wind is coming from the east, not going to east...

Gr3q:
@French77 If you switch to symbolic icons are they also black? (icons are not black for me btw)

French77:
@Gr3q When I switch to symbolic icons,direction icon is white after refreshment,but don't really like the symbolic icons.

Gr3q:
yeah sure, I just wanted to see if there is a problem with the logic

jorgenqv:
@Gr3q 
I think it's a good idea to keep the Wind direction arrow and the letters at the end to minimize confusion :-)
Like: ← 5.6 m/s E

PR #3490, opened 2021-01-31
weather@mockturtl Update 3.0.1 - hotfixes

Gr3q:
Fixes:

* Fix issue where Symbolic icons setting was not reflected in the app when changed
* Fix [#3486](https://github.com/linuxmint/cinnamon-spices-applets/issues/3486), now wind icons point to the right direction
* Fix wind icon color when non-symbolic icons are used in the applet popup menu
* Fix [#3488](https://github.com/linuxmint/cinnamon-spices-applets/issues/3488), add Climacell V4 as a new provider because Climacell v3 is deprecated and dosen't accept new signups.
* Make network and DNS related errors soft errors again


Gr3q:
@brownsr This is ready and tested on 3.8

PR #3521, opened 2021-02-06
spices-notifier@germanfr: rebuild dist, fix broken files

germanfr:
I regenerated the applet from source according to the new guidelines. Some files where broken.


NikoKrause:
Sorry, i must have picked the wrong icon.

The RADME. md and screenshot.png shouldn't be in UUIID/files/UUID, unless they are used in the applet. A normal user doesn't look in the applets directory and doesn't need to download those files with every update.

PR #3527, opened 2021-02-08
[radio@driglu4it] Add option to download current song from Youtube

jonath92:
Added a feature to download the current running song from youtube with youtube-dl. This feature doesn't work with the package in the default repository as you always get a "Unable to extract video data" error (see [stackoverflow](https://stackoverflow.com/a/65464596/11603006)) so I also added a script to install it including a confirmation dialog.

In this context I also refactored a little bit the javascript logic for the download prompts of the mpris plugin and the mpv package as it is much easier to handle multiple asynchronous actions with promises than with callbacks. Now users only need to click once on the applet to install both packages. But I haven't changed anything on the dialogs itself.


clefebvre:
You shouldn't use apt directly, this is supposed to work with any distribution. I'd recommend just checking the version of mpv installed (using mpv, not apt) and warning/asking the user about version or installation if appropriate.

jonath92:
thanks for the information @clefebvre. This was actually already the case before this PR but anyway I fixed it now by using the `find_program_in_path` from GLIB which I guess works on all distributions (at least this is also used in the Sensors applet).

Only when apt is present it is shown a download dialog. If not, it is shown a notification that the packages are required but the users have to install it on their own. The packages are also mentioned in the readme.

brownsr:
Waiting to see if use of apt is removed, as recommended.

Regards

Simon

jonath92:
Thanks for the answer @brownsr. But not sure what you mean. It is insofar removed that users without apt can now also use the Applet which wasn't the case before. However it is still used apt for Installation for apt users as it makes the Installation for them very easy and I don't see any benefits in removing it. Or do you mean a cofirmation of @clefebvre?

jonath92:
@brownsr Solved the conflict now (sorry for the delay). It would be really helpful if this could be merged. Not merging it because of apt doesn't make sense as I on the one hand have fixed it and on the other hand the problem with apt is also in the master branch (or to be more precise only in the master branch). Thanks in advance. Best regards

brownsr:
OK, I see the point about apturl having been there previously.  It's used elsewhere in other spices too.  We'll review, don't be too surprised if we decide to strip it out at some time across the board.

Regards

Simon

PR #3535, opened 2021-02-09
Update Cinnamenu Hungarian translation

KAMI911:


fredcw:
@KAMI911 Thanks for the translation for cinnamenu. Just in case you didn't know, the file you updated was based on a version of the .pot file from 2020-07-17. The way to get the latest strings to translate is to use a program like Poedit, load the hu.po file and then "Update from POT file" in the "Catalogue" menu. The .pot file is in the same folder and is much more up to date.

KAMI911:
Oh, I see :+1: Thank you for the heads up. I will commit the updated file soon.

PR #3542, opened 2021-02-10
add new fr.po file

InfoLibre:


NikoKrause:
No, that does not look right.

NikoKrause:
Do not delete the fr.po file. Just try to update the file as you do in other and replace the whole content of the fr.po file with the content of the fr.po, which you created with poedit.

Issue #3538, opened 2021-02-10
weather@mockturtl : No space before % and no uppercase at the beginning of days


InfoLibre:
```
 * Applet version 3.0.1
 * Cinnamon version 4.8.6
 * Mint 20.1
 * Graphics hardware *and* driver used
 * Snippet or attachment of ~/.cinnamon/glass.log
 * Snippet or attachment of ~/.xsession-errors

 ```

**Notify author of applet**
@mockturtl

**Issue**

1) In the French version, the days don't have a uppercase letter at the beginning (there are any in the .po file but not on the screen)

2) A space is missing between the value and %

![Capture d’écran du 2021-02-10 09-05-21](https://user-images.githubusercontent.com/3955499/107482445-fd529b00-6b7f-11eb-9a7f-9e387ff6f9c7.png)

Gr3q:
Hi, the humidity, I think it was always like this, but I can change it.

The days are coming from straight to Javascript that's why translations are not used. I'll add an extra check to capitalize them when french locale is used

InfoLibre:
I don't know the exact rule but I think that there's always a space between a value and a unit. Perhaps it's different with humidity ?

Gr3q:
No idea looking at weather websites (at least English ones) there is no space, but it looks ok, so I can add that as well

InfoLibre:
I found this :
http://bdl.oqlf.gouv.qc.ca/bdl/gabarit_bdl.asp?id=4870

Gr3q:
Javascript has a formatter based on language for percent, so I'll start using that. (UK is 5%, FR is 5 % for example)

Issue #3507, opened 2021-02-03
weather@mockturtl New 3.0 release - Configure improvments thoughts


jorgenqv:
```
 * Applet version/Build date 3.0.1 Feb 1, 2021
 * Cinnamon version 4.8.6
 * Distribution - Manjaro Cinnamon 20.2.1 Nibia
 * Graphics hardware *and* driver used - Intel Haswell-ULT Integrated Graphics Controller (i915)  

```

**Notify author of applet**
@Gr3q

Thank you and congrats to an excellent 3.0 release of the Weather applet! I really like it :) Here are some minor suggestions below to make it even better.

**Issue**

The Help tab in Configure contains lots of text and can be a bit hard to read and understand.
So here are some ideas for a better placed and clearer help texts by spreading it out.

1. Move text section 'How to choose a weather provider' from the Help tab to the bottom of the first Weather tab and change it to:
```
About Weather Providers:
You can use several different weather forecast providers by choosing from 
the dropdown menu Data service. Some providers require a API key to work, 
some have regional limits, differ in forecast length, and they all need a 
functional internet connection to work. Usually it takes a wile after 
registering before the API key works. Options are described in detail on 
the Cinnamon Spices Website, click the link button under the Help tab.
```

2. Move text section 'Allowed location formats' from the Help tab to the bottom of the second Location tab and change it to:
```
About Location formats:
Your location in most cases work automatically but you can enter it manually 
by activating Manual location and filling out the Location field. 
Use coordinates 'Latitude,Longitude' (e.g. 51.5085,-0.1257) or use 'City,
Country-code' (e.g. London,UK) or try to enter {House number} {Street} {City/
Town} {Postcode} {Country}. The search algorithm is very flexible with the 
format. After 3 seconds without typing, the address you entered is replaced 
automatically for validating your choice.
```

3. Change text sections 'View additional information' and 'Report an issue' under the Help tab to:
```
More about the Weather applet:
For more detailed information about the applet please go to the Spices Website 
by clicking the button below. You can also comment and ask a questions to the 
community at this page by logging in with your Github account. You can also find
information about known issues, troubleshooting, changelog etc. 

Report an issue:
If you find an issue with this applet please make a report by clicking on 
the button below. Please make sure that the issue is not already submitted.
Login in with your Github account. Don't forget to write weather@mockturtl 
in the top Title field to make it easier for everyone to know that the issue 
is about the weather applet (the form is for all Cinnamon applets).

```

[...]

Issue #3508, opened 2021-02-03
weather@mockturtl Add Weather Forecast Provider DMI possible?


jorgenqv:
@Gr3q

I have been using the Danish Weather Provider DMI on the web for many years and feel they often are very accurate where I live. They do not have a official API yet but plan for providing this for free next year.

But then I found this unofficial workaround, so I wonder if it's possible at all to add DMI in the Weather Applet as a provider now or is it not possible?

https://github.com/rugaard/dmi

Thanks!

Gr3q:
After having a look, i think I might be able to do this.

Their internal API url is this atm: `https://www.dmi.dk/NinJo2DmiDk/ninjo2dmidk?cmd=llj&lon=0.1084&lat=52.2231` with my full query.

Gr3q:
I'll integrate it if you define the payload for me (like each field's unit).

I'm most interested in `windDir` and `precipType` because they are not in English and their possible options, and the `symbol` property, all the possible options for it and what they mean (icons represent the symbol on the website).

You can cross-reference the symbol's in the API and the icons(conditions on the website).

If you provide all of this I'm more than happy to integrate it

Gr3q:
Here https://www.dmi.dk/lokation/show/US/4959902/Cambridge/#vejrgrafer I think the icons' filename in the website's html matches the symbol name (at least for hourly weather), so it makes life easier

jorgenqv:
Very cool that there is a chance that this could work :)  I will be happy to try to define the icons definition to English but I am unsure how to do it and to cross reference and match? If i look at the API feed `time "20210206220000"` compared to the website it looks like `precipType 'regn'` have an icon that looks like cloudy? https://www.dmi.dk/assets/img/3.svg
Also this looks strange at this time:  `windDir "Ã\u0098NÃ\u0098"`

Gr3q:
precipType is irrelevant (I know, it's weird) for the symbol, maybe the precip chance is under a certain % so it's not worth displaying. 

Focus on the symbols for hourly weather (`timeserie`) array and check if they match with the iconNames on the website. If they do, you just have to come up with an english title (or just give me the icons, whichever easier) for each code.

For windDir I guess is the browser can't decode the char `~` after every `Ã` char properly, you decide if the `~` char is important in your language

jorgenqv:
[...]

PR #3549, opened 2021-02-10
weather@mockturtl update 3.0.2

Gr3q:
Changes:
* Fix icons for Met Norway, some are missing
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3507 -  Reorganize settings, Add help text where it belongs, to the tooltips
* Adds https://github.com/linuxmint/cinnamon-spices-applets/issues/3508 -  Add Danish Meteorologist Institute as a Weather Provider
* Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3538 -  Make sure French days are capitalized and Use Locale formatting for percent (humidity)

Tested on 3.8


NikoKrause:
@Gr3q I just noticed that mockturtl is still listed as maintainer of the weather applet, but the maintainance seems to be shifted to you. Has there be an agreement? Either way, i think we should change that. Do you like to create a PR to reflect, that mockturtl is the original author, but you are the current maintainer.

See for example here: https://github.com/linuxmint/cinnamon-spices-applets/blob/4a9eb7d42ac9097ea19814eac6fe9b8f552d8d99/hwmonitor%40sylfurd/info.json

Gr3q:
@NikoKrause I'll include it in the next version's PR

PR #3557, opened 2021-02-13
weather@mockturtl : improve French translation

InfoLibre:


InfoLibre:
I will done another later

Issue #3533, opened 2021-02-09
cryptocoin@guantanamoe : applet is downloaded but doesn't appear in the installable applets


InfoLibre:
```
 * Applet version 1.7
 * Cinnamon version 4.8.6
 * Mint 20.1
 * Graphics hardware *and* driver used
 * Snippet or attachment of ~/.cinnamon/glass.log
No file.
 * Snippet or attachment of ~/.xsession-errors

 ```
[.xsession-errors.tar.gz](https://github.com/linuxmint/cinnamon-spices-applets/files/5950123/default.xsession-errors.tar.gz)

**Notify author of applet**
@imDeprecated

**Issue**

I can download Crypto applet :

![Capture d’écran du 2021-02-09 09-58-50](https://user-images.githubusercontent.com/3955499/107341044-ce282500-6abe-11eb-88cb-d468fc420f40.png)

But after, when I want to add it on my desktop bar, I can't because it doesn't appear in the list :

![Capture d’écran du 2021-02-09 09-59-17](https://user-images.githubusercontent.com/3955499/107341170-ee57e400-6abe-11eb-9228-8f14637e0dbb.png)

InfoLibre:
Is it since I updated the translation fr.po file ?

NikoKrause:
I can't reproduce it. Have you tried deleting and reinstalling the applet?

InfoLibre:
Yes, throught applets list and after, with Linux command. Same result.

InfoLibre:
How could I help ? By typing a command or sending you a file ?

NikoKrause:
Hey,

you were right. It is indeed a problem with the French translation.
Are you using `poedit` for translation or are you just using a normal text editor?

To create proper translation files you should get used to use poedit. Here are some instructions on how to get started with poedit: https://github.com/NikoKrause/CinnVIIStarkMenu/issues/19
You can ignore the first two sections and start with section "Create a translation for the first time with the application Poedit".

InfoLibre:
I used this Gitlab editor. I will try with poedit.

InfoLibre:
A new version or fr.po, edited with poedit. Sorry for the problem.

[fr.po.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/5957448/fr.po.zip)

NikoKrause:
Looks good.
Can you create a pull request?

InfoLibre:
[...]

PR #3575, opened 2021-02-14
validate-spice: check po files for errors

NikoKrause:
Use `msgfmt -c xx.po` command to verify po-files.

According to the man page (https://www.gnu.org/software/gettext/manual/html_node/msgfmt-Invocation.html), the command does the following:

`-c`, `--check`

Perform all the checks implied by `--check-format`, `--check-header`.

`--check-format`

Check language dependent format strings.

If the string represents a format string used in a `printf`-like function both strings should have the same number of `%` format specifiers, with matching types. If the flag c-format or possible-c-format appears in the special comment #, for this entry a check is performed. For example, the check will diagnose using `%.*s` against `%s`, or `%d` against `%s`, or `%d` against `%x`. It can even handle positional parameters.

Normally the xgettext program automatically decides whether a string is a format string or not. This algorithm is not perfect, though. It might regard a string as a format string though it is not used in a printf-like function and so msgfmt might report errors where there are none.

To solve this problem the programmer can dictate the decision to the xgettext program (see c-format). The translator should not consider removing the flag from the #, line. This "fix" would be reversed again as soon as msgmerge is called the next time.

`--check-header`

Verify presence and contents of the header entry. See Header Entry, for a description of the various fields in the header entry.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -29,5 +29,5 @@ jobs:
       - name: Validate spices
         run: |
           sudo apt-get update -y >/dev/null 2>&1
-          sudo apt-get install -y python3 python3-pil >/dev/null 2>&1
+          sudo apt-get install -y python3 python3-pil gettext >/dev/null 2>&1
           ./validate-spice --all
--- a/validate-spice
+++ b/validate-spice
@@ -2,6 +2,7 @@
 import json
 import os
 import sys
+import subprocess
 from PIL import Image
 
 if len(sys.argv) != 2:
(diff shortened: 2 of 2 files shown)


NikoKrause:
This check is due to the following critical issue: https://github.com/linuxmint/cinnamon-spices-applets/issues/3533

This PR will prevent such issues in the future. Furthermore it also detected this issue in other places, e.g.:
https://github.com/linuxmint/cinnamon-spices-applets/blob/master/brightness-and-gamma-applet%40cardsurf/files/brightness-and-gamma-applet%40cardsurf/po/pt-br.po#L16

This issue and 99% of all the other errors, which are shown by this check can easily be fixed with the following easy steps:
1. Open the po file with poedit. 
2. Press the save button. 
3. Done

Maybe we should also add those steps into `validate-spice `?

PR #3590, opened 2021-02-16
ratecurrency@magner: Fix pot file

NikoKrause:
The comment has to be above the msgid

@AleksiMagner fyi


NikoKrause:
@AleksiMagner According to your comment, you add two strings called "days", because some languages have multiple plurals, e.g. Russian.

The better way to handle this, is to use `ngettext`. See for example how it is done here: https://github.com/linuxmint/cinnamon-spices-applets/blob/master/CinnVIIStarkMenu%40NikoKrause/files/CinnVIIStarkMenu%40NikoKrause/4.2/applet.js#L3944

PR #3600, opened 2021-02-18
qredshift@quintao: Version 1.5.3

raphaelquintao:
* 1.5.3
    - Fixed some issues with Debian.
        - Automatically disable redshift default service.
    - Removed default keyboard shortcuts.
    - Added translations:
        - Hungarian by [KAMI911](https://github.com/KAMI911 "@KAMI911 on Github")
        - Danish by [Alan01](https://github.com/Alan01 "@Alan01 on Github")


NikoKrause:
IMO the description in the metadata is too long and contains information, which you already have in the README.

The description should not replace the README. It should just give a small information about what it does. If I see it correctly, there's already a mention of the conflict in the README and you inform the user, when he uses the applet. There's no need to add this info in the metadata description. 

Also keep in mind that translations do get lost everytime you change the description. I don't say you should never change the description again. I just think you should keep in mind, if a change and therefore new translations are really necessary.

PR #3620, opened 2021-02-22
sound150@claudiux: add "Always show input controls" option

dlenski:
I think the function is pretty straightforward based on the description.

I can try to add corresponding changes for other Cinnamon versions if there's interest in merging it.


brownsr:
ping @claudiux

claudiux:
I accept these changes. Thank you @dlenski.

dlenski:
> I accept these changes. Thank you @dlenski.

Great! Just note that I only modified them for `3.4`, since that's the only version I can directly test.

PR #3638, opened 2021-02-26
Add link to Translation Status Tables in README

NikoKrause:


AndreiMiculita:
Not sure how feasible this is, but it would be useful if the table also contained a column with the score on cinnamon-spices.linuxmint.com (i.e. the number of stars), so translators can prioritize the most popular spices.

NikoKrause:
This is a nice idea, but unfortunately this is not possible. The stars are only stored on the website and I can't grab that info from within GitHub. 
So translators have to open the website for now, if they want to translate by popularity. Sorry. 

I also wanted the tables to be sortable, but this is also not possible with GitHub Markdown.

PR #3657, opened 2021-03-01
Added function: automatically loop

kancolle386:
setting-schema.json
- add an item "auto-loop-enable" to page "General" to select loop on/off.

applet.js
- add variable "timerDurationLoop" to keep current timeDuration for loop .
- modify function "doStartTimer" to save current timeDuration value to new variable "timeDurationLoop".
- modify function "doTimerExpired" to recall the function "doStartTimer".


brownsr:
ping @jake1164

NikoKrause:
I haven't seen any activity or reaction from @jake1164 for quite some time now. Since this is a pretty small change, I'm gonna boost this process a little bit. If @jake1164 comes back and doesn't like this option, we can quite easily revert this commit.

PR #3667, opened 2021-03-04
tzclock@lhriley: fix performance / timing issues

lhriley:
This PR fixes an issue where Cinnamon would experience performance issues (high CPU use) and the clock itself would become desynced from real time. This would typically happen when the machine would be in a `suspend` state, but would also occur over time even when the machine had power saving functions disabled.

The primary cause was using a naive periodic timeout rather than relying on notifications from `notify::clock`.

This problem likely also occurs in [calendar@simonwiles.net](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/calendar@simonwiles.net/files/calendar@simonwiles.net/applet.js) as this was one of the original sources of code that I borrowed: https://github.com/linuxmint/cinnamon-spices-applets/blob/d172c5008c324212e258952964f28e3607bc90f5/calendar%40simonwiles.net/files/calendar%40simonwiles.net/applet.js#L256

This applet now works more similarly to the [calendar@cinnamon.org](https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/calendar@cinnamon.org/applet.js) applet. It shows none of the performance and timing issues so far, and multiple instances of the applet remain perfectly synced.

I did not include changes to the related applet `calendar@simonwiles.net` in this PR, but I could put something together if requested.

### DIFF
--- a/tzclock@lhriley/files/tzclock@lhriley/applet.js
+++ b/tzclock@lhriley/files/tzclock@lhriley/applet.js
@@ -22,6 +22,7 @@ const UPowerGlib = imports.gi.UPowerGlib;
 const Settings = imports.ui.settings;
 const GLib = imports.gi.GLib;
 const Gio = imports.gi.Gio;
+const CinnamonDesktop = imports.gi.CinnamonDesktop;
 
 let DEFAULT_FORMAT = _("%l:%M %p %Z");
 
@@ -55,6 +56,8 @@ MyApplet.prototype = {
 
             this._initRightClickMenu();
 
+            this.clock = new CinnamonDesktop.WallClock();
+
             // Track changes to clock settings
             this._dateFormat = DEFAULT_FORMAT;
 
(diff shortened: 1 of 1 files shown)


NikoKrause:
@simonwiles Would you like to take the offer from @lhriley

lhriley:
@NikoKrause I can make the changes for `calendar@simonwiles.net` in a separate PR if @simonwiles gives the :+1: . No reason to delay this PR over it.

lhriley:
Thanks @NikoKrause 👍

simonwiles:
Thank you both, and apologies for not getting to this sooner.

lhriley:
@simonwiles no problem, random comments on PRs don't demand real time responses ^_~

I can put together a PR to fix up the code when I have some free time (maybe this weekend). I'll tag you on the PR once I can get it posted.

I had no idea that your version of the applet had been forked that long ago, so it isn't that surprising that someone figured out a more efficient method since then. I had actually been using your version of the applet for several years until I recently decided that it was time to give in and make an applet that just displayed the clock I wanted. Funny how it all comes full circle back to the original `calendar@cinnamon.org` code showing us how to write code to display a clock.

PR #3677, opened 2021-03-07
Added 2 functions

kancolle386:
Add function: Automatically loop. (Restart the timer at same duration)
Add function: Specify sound volume for timer has elapsed.


brownsr:
this is just a version bump, but as I have just merged a change, I'm fine with this.

Regards

Simon

PR #3678, opened 2021-03-07
Add function: You can specify sound volume.

kancolle386:
applet.js
Add a parameter to spawnCommandLine.

settings-schema.json
Add scale for specify volume.


brownsr:
ping @jake1164

NikoKrause:
The last contributions from @jake1164 on his applets seem to be more than 2 years ago. Also I can't remember seeing any reactions to pull requests or issues regarding his applets.

So I suggest that we give it another 10 days and if we don't get any reaction we can call his applets abandoned.

@kancolle386 would you be interested to be the new maintainer in this case? If not, we just set the author to none.

kancolle386:
What happens to this applet when author is set to none?

brownsr:
We won't refer any changes to the author, as there is none. So changes in functionality can be made freely by anyone.

Regards

Simon

brownsr:
It's been a while now.  I'm going to assume applet is abandoned and merge this.

Regards

Simon

kancolle386:
Hello, I would try maintainer of this applet.

PR #3683, opened 2021-03-13
calendar@simonwiles.net: fix potential performance / timing issues

lhriley:
Quick PR to update `calendar@simonwiles.net` to use the notification interface for updating the clock rather than relying on a naive timer.

While I did not directly experience performance issues with this applet, I had originally used this code as a starting point for `tzclock@lhriley` where I did experience issues with high CPU usage and clocks becoming desynced. As a result, this patch is an attempt to proactively avoid similar issues by using the update methodology from `calendar@cinnamon.org`.

These changes are proposed with the blessing of the original author @simonwiles.

Related PR for additional context: https://github.com/linuxmint/cinnamon-spices-applets/pull/3667


lhriley:
Thanks @brownsr 👍

PR #3687, opened 2021-03-15
[Cinnamenu@json] Performance improvements

fredcw:
+ created new class for header text (.menu-applications-header-text)


fredcw:
Not sure why All checks failed?

NikoKrause:
This seems to be a temporary issue with GitHub Actions.
I saw this once. Maybe a server issue.

fredcw:
Ok thanks, I updated the pot file anyway, and there was one string that was different that I forgot about. Maybe it was that?

NikoKrause:
No, I had the same issue in the desklet repo. How did you update the potfile? I'm wondering, because all the comments are gone.

fredcw:
Yeah, I usually delete all the comments because I think it saves a lot of space, less than half the size. It's just a quick regex replace. I wonder though if the comments are useful to the translators and maybe should be kept?

NikoKrause:
Some comments can be useful.

PR #3691, opened 2021-03-18
azan@fahri.nurul.id: Add settings options to select juristic method

umarhussain15:
feat(azan@fahri.nurul.id): add settings options to select juristic method

This update will add the option to select the juristic method in settings menu.
Juristic method allows to adjust time calculation for Asr Prayer.
The library Prayer.js already has this option, now it is hooked to the settings menu allowing for users to adjust this setting.


brownsr:
ping @fahrinh

brownsr:
ping @fahrinh

brownsr:
ping @fahrinh 

May assume applet is abandoned without a response either way

Regards

Simon

PR #3692, opened 2021-03-19
devbar@ludvigbostrom: Add the devbar applet

ludvigbostrom:


brownsr:
Throws a couple of errors.  Take a look at .xsession-errors.  There is a tostring error showing there, and then there looks to be another error when msg is null, perhaps before configured.  Also can you please add more information about the applet, as there's not really enough there to be useful. If I wanted to use it, say, I'd want to know where the output gets put for example, but perhaps the errors are stopping me seeing output perhaps.  So some cleaning up to be done before it is usable by 3rd parties I think.

Regards

Simon

ludvigbostrom:
Great comments, I think the issues have been resolved.

Issue #3694, opened 2021-03-19
weather@mockturtl Temperature unit features


jorgenqv:
**Notify author and original request submitter**
@Gr3q
@spsf64
Summary and some more ideas for Temperature unit feature idea from the comment section on the Spices website.

**Feature request**
Add the option to display both temperature units (Fahrenheit and Celsius) at the same time in the weather applet.
Reason for the request is that it can to be useful for people moving between places that use different temperature units
with the purpose of learning and understanding temperature unit differences.

**List of ideas and requirement thoughts**
- The option could be placed in Configure under Other options / Units / Temperature unit.
- The name of the option could be "Show Fahrenheit and Celsius at the same time" and the default setting is false.
- If enabled both temperatures are shown like 25°C (77°F) in applet Panel and Popup current temperature on the top right. 
- If the option is enabled to true and Temperature unit is Fahrenheit the shown order should be 77°F (25°C).
- Another way to go is to use linked toggling as used in popup on Hourly, i.e. when clicking on any F och C unit marks in the applet the value to the left of the mark changes between Fahrenheit and Celsius on every click.

**Expected behavior**
When enabled you get the possibility to see Fahrenheit Celsius units more directly. 

**Other information**
- Unit change shortcut idea: Once upon a time many years ago the applet had a little Gear icon top right in the popup to access Configure (see screenshot on the Spices website). One idea is to reintroduce this icon again but this time make it a direct shortcut to the Configure "Other options" Tab (if possible?), this way you save 2 clicks and can be seen as a little bit more intuitive.
- Another minor thing related to temperature: Move "Show temperature Russian style" to Unit section under "Temperature unit", and add to description or tool-tip: (i.e. −2…+3 °C). The reason for this is that it's shown in both popup and panel, the parenthesis text explains what it does, and it's about temperature and therefor perhaps better placed under Units/Temperature).

Gr3q:
I have to choose a format, options:
Only display alternative unit's temperature
* 0 (32) °C 
* 32 (0) °F
* 0 (32) /  0 (32) °C 
* 32 (0) °F / 32 (0) °F

Show everything one after the other
* 0 °C (32 °F) 
* 32 °F (0 °C)
* 0 °C (32 °F) / 0 °C (32 °F)
* 32 °F (0 °C) / 32 °F (0 °C)

Show temperatures and units grouped
* 0 (32) °C (°F)
* 32 (0) °F (°C)
[...]

PR #3698, opened 2021-03-20
weather@mockturtl update 3.0.6

Gr3q:
Changes:
- Resolves https://github.com/linuxmint/cinnamon-spices-applets/issues/3694
- Move Typescript declarations out of the weather applet folder to prevent update triggers from other applets making declaration changes. **Please review this now** if it's ok, should be in a folder with a different name or have a more thorough README etc.
- Update de.po with contribution from kipuka@eclipso.eu
- Add deprecation notices for relevant providers and their EOL date
- Make provider names translatable
- Add stack trace to errors (even if they do not seem to be correct) 
- Add option to switch between textual and icon representation of wind direction
- Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3738
- Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3733
- Fix issue where locations in Africa would break OpenWeatherMap logic by not having minutely precipitation in their payload.

@brownsr @NikoKrause Please review the location of the Typescript declarations, I want to make sure everyone is happy with the naming etc. Ofc I will continue to maintain and extend them as before.

@KAMI911 @Pjotr123 @jorgenqv @sveinki @serkan-maker @FabienRCT @Alan01 @French77  Translations are now final for this PR and include minor changes.

@jonath92 This PR will change the location of the declaration files so you will have to update your `tsconfig.json` when this is merged.

 I will keep this PR a Draft until Sunday, than I will open it up for review.


jorgenqv:
@Gr3q  Here are some minor observations of stuff, may och may not be of interest. From my perspective not that important to be entered as issue, but if I'm wrong just tell me and I will be happy to report them as separate issues :) 

1. Mentioned last comment Feb 3 issue #3486 : I would appreciate if we could bring back the abbreviation of Wind directions last and still keep the arrow showing the direction first. Like "Wind: ← 5.6 m/s E" or "Wind: ↖ 5.6 m/s SE". Reason: There is sometimes confusion about wind direction and this makes it a bit more clear?

2. Powered by Weather Provider name in popup bottom right should be translated. Reason: Should be uniformly like it's shown in Configure Data Service drop down, some names are partly translated like Norway/Denmark, at least in my translations. Names can also be localized.

3. Hourly Forecast length (hours) in Configure cannot be set to more than 48 but some Weather providers do have hourly forecasts beyond 48 hrs. Maybe not possible, just curious about if Hourly forecasts can be longer if the provider have this?

Thank you!

Gr3q:
@jorgenqv number 3 and 2 can be done (although personally I don't think 2. is necessary).

The only reason I didn't implement it because:
* How would you display wind direction icon+ wind direction text + beaufort wind speed combination?
* Not keen on adding another option to the settings, especially for displaying the same data twice, but it can be done

jorgenqv:
> @jorgenqv number 3 and 2 can be done (although personally I don't think 2. is necessary).
> 
> The only reason I didn't implement it because:
> 
>     * How would you display wind direction icon+ wind direction text + beaufort wind speed combination?
> 
>     * Not keen on adding another option to the settings, especially for displaying the same data twice, but it can be done

Quite alright, if it's fully understandable. However Beafort is strength and not direction or am I missing something? 
Thanks! :)

Gr3q:
I just though it might be too much

jorgenqv:
> I just though it might be too much to display.
> 
> Do you want to display both, not one or the other?

If you think it's too much I fully respect your decision not to implement it. 

Example with different setting as it could look:
Wind: ← 9.3 m/s E
Wind: ← 18 Knots E
Wind: ← 33.5 kph E
Wind: ← 5 (Fresh breeze) E

The only difference is the last wind direction letter.

Gr3q:
@jorgenqv It still doesn't feel right, I think just changing between the 2 format is enough

PR #3705, opened 2021-03-24
[radio@driglu4it] Migrate to Typescript

jonath92:
I migrated the radio applet to typescript which makes it much easier to maintain. I am really sorry for so much change but most is only because the files are now twice (as with the weather applet).  Apart from typescript I also made some other fixes/changes: 

- In the past there have been several problems when using streams with .pls extension such as showing "stop" on the panel when adding a new station. This is now fixed
- When removing a station which is currently playing, the radio now stops
- it is not possible to pause/resume the radio by middle click on the icon


jonath92:
I found a not negligible bug in the pr. The stop icon is indicated with a dot when starting the radio the first time after start of the pc (not after restarting cinnamon). Luckily it hasn't been merged yet. I close the pr now and will reopen it after I have fixed it.

brownsr:
I was just reading it through ...  anyway, the tsconfig.json file has a link to the weather applet.  Is that right ?

Gr3q:
@brownsr He is using my type declarations (for GJS bindings and stuff in `/usr/share/cinnamon/js`), without them using Typescript is kinda pointless. (They aid static analysis in IDEs and at build time and provide documentation (and autocomplete) on how to use the libraries).

The declarations are incomplete ofc, because there is quite a lot of work to create them from 0, but since my last version they are usable separately from the weather applet.

I think now is the time to discuss if that folder still needs to live inside weather@mockturtl folder or is it time to move it out into it's own folder before @jonath92 creates another PR. Ofc I will still maintain and expand it, but other people can as well.

If moving declarations out to it's own folder is on the table, I'm prepared to write a new page in the wiki based on the current example (with any gotchas, etc.) on how to create an applet with Typescript.

jonath92:
@brownsr fixed it now - was just an one line fix (i.e one line in ts and one line in the responding js file). 

I agree with @Gr3q that it makes sense to put the typescript files in its own folder. Though it doesn't matter programmatically where the declaration are stored, it is more logical to store it outside of an applet. 

A tutorial would be definitely were useful :thumbsup:.  I would be willing to test it and provide feedback.

PR #3709, opened 2021-03-25
[Cinnamenu@json] Add some translations.

fredcw:
+ update some .po and pot files
 + bug fix


fredcw:
Oops, I didn't notice that @kelebek333 also has a PR updating tr.po. So I've updated this PR to include kelebeks updated version of tr.po. Both PR's can be merged in any order since they both now have the same version of tr.po.

PR #3711, opened 2021-03-27
[radio@driglu4it] Add volume slider

jonath92:
- Add a volume slider to the radio applet to make it more intuitive to use (later I also want to add media controls)
- Show the stop item only when radio is playing
- make station search case insensitive and remove trailing spaces 

@Gr3q I had to add `_slider` to `PopupSliderMenuItem`. I know that `_` usually is used in JS as a convention for private attributes but using this is the only way I know to add an icon in front of the slider. This is also used in the sound and the power applet for this purpose. I didn't make a separate PR for this as it is easier that way. So i think @Gr3q  needs to approve the declaration change and after that @brownsr can merge it. Hope that is ok..  Best regards


Gr3q:
@jonath92 I don't mind declaring private attributes in declaration files (actually, I included all I didn't miss, but I didn't do thorough work at all), so you are free to add them wherever you find missing ones.

The only requirement for me is that they should be marked with a private accessibility modifier. I know Typescript will complain, but you can still force access private variables with a `this["attributename"]` syntax.

Nevertheless in this specific case we could debate that those attributes are not even private, but _protected_, but to be honest the only ones who know is the developers who wrote the code. 

Anyway, I will open a PR into your PR and see what we come up with, personally I think it makes sense to make those attributes protected so when you extend those classes you can freely modify them.

TLDR:
The main takeaway here that those attributes **should exist** in the declaration files in some form (what we'll discuss in another PR in a minute), and they are missing only because I was not thorough.

jonath92:
> The only requirement for me is that they should be marked with a private accessibility modifier. I know Typescript will complain, but you can still force access private variables with a this["attributename"] syntax.

I didn't know that. Thanks for the info

>Nevertheless in this specific case we could debate that those attributes are not even private, but protected, but to be honest the only ones who know is the developers who wrote the code.

Yes I find *protected* most reasonable in that case as it is reasonable to use in child classes. So I changed it to protected (no idea why I haven't done it directly that way). 

>The main takeaway here that those attributes should exist in the declaration files in some form (what we'll discuss in another PR in a minute), and they are missing only because I was not thorough.

I think you have been pretty thorough considering how much declaration is needed. You really did a great job. I am not sure what you mean with another PR as this is just an one line change which I now did.

Gr3q:
I meant this https://github.com/jonath92/cinnamon-spices-applets/pull/1

jonath92:
ah sorry for the confusion @Gr3q . I  have overseen it. Merged it now anyway as you made some more declarations. @brownsr  you can now review as the changes in the weather folder are now from @Gr3q personally.

Gr3q (approved):

PR #3718, opened 2021-04-04
Hungarian translation and update for some previous translations for Cinnamon spices applets

bossbob88:
Hello everyone,

as described in the title these are translation file for the applets (starting with letter d all the way until x) and some of them are updates for existing translations.


brownsr:
Many Thanks !

Simon

PR #3722, opened 2021-04-06
[Cinnamenu@json] Fix tweaks for some themes.

fredcw:
+ remove duplicate search results.


brownsr:
A little puzzled by the settings override, but you must have your reasons.

Regards

Simon

fredcw:
>settings override

This is to fix this bug: https://github.com/linuxmint/cinnamon/issues/9920
It removes the file monitor from settings.js such that it relies solely on the dbus call to remoteUpdate() to be notified of an external change to the settings file. Although I had mitigated for this bug by finalising settings when starting an applet resize drag, it still occurs very occasionally. And there was still the potential for the bug to occur when changing other settings, so this fixes that as well.

PR #3741, opened 2021-04-13
[radio@driglu4it] Add Toolbar, Refactoring and experimental test support

jonath92:
First of all I am sorry to make so much code changes (again) but I am still learning why I didn't make everything right in the first place .. 

I made the following changes: 
- Add a toolbar for pausing/resume/stopy, copy song title and stop more easily
- The title of the song and the current channel is now additionally shown in an own section in the popup menu which makes it a more complete app
- A lot of refactoring - following a more functional approach and increase modularization 
- Add experimental support for unit testing. It is not optimal as currently you get an error when running the build script without running `npm install` before but it is a start
- Update readme and screenshot


Gr3q:
The baseurl in the `typescript.json` is very useful, I've stolen it already.

Based on the code I've read, you might be using Looking Glass to read logs (ignore if not).
Try to restart Cinnamon from the command line:
`export DISPLAY=:0; cinnamon --replace &` _(if it doesn't work, leave off the &)_

Everything you print with `global.log` will show up properly in your terminal (functions, objects properly printed, exceptions fully printed etc.). Looking Glass has trouble displaying anything except strings.

jonath92:
> The baseurl in the typescript.json is very useful, I've stolen it already.

Feel free to "steel" what ever you want. It's an honor for me when the best applet developer "steals" something from my code ;-). Also just fair as I have already stolen a lot from you :-)

> Based on the code I've read (out of curiosity), you might be using Looking Glass to read logs (ignore if not).

Yes that's true. Your hint is useful. So far I have overcome this limitation with `JSON.stringify` but I am still very unhappy with my current logging strategy. So far I usually have been read the log by either `tail -f ~/.xsession-errors' or by opening the `xsession-errors` file directly in a text editor. It is probably better to use the Looking Glass as I can filter there the logs by logging level (error, warning, info, trace) but I am just more used to reading text files for logging. But apart from that the main remaining problem for me is that you can't filter the log by applet (at least I don't know how). This is not so much a problem for stuff I am logging during runtime but it is a problem for stuff I am logging at cinnamon start as I every time waste my time searching the info for the radio-applet in the text-file (would be the same with Looking glas I think).  Also another limitation is that I can't toggle the logging for development/production (in production I obviously want to show less information than in development). 

Basically I want to do exactly what you do with your `logger.ts` file in the future - printing the  logs to an own file (I haven't tried your code but that's how I understand your Log class). The only thing I don't like about that, is the fact that I would have to copy your code for that which is obviously not good coding practice (I already have copied your south arrow file as I use this as download icon). This is on the one hand because you have hard coded the UUID (which obviously could be fixed easily) and on the other hand I am not sure if it is possible to use a file from another applet in typescript...  Alternatively it might be possible to make a node module for the Logger (just thinking not sure if this is possible)?  IMO it would be awesome if it were possible to create node modules for gjs/cjs.

Gr3q:
My logger class doesn't do much, it simply prepends logs with the applet name and instance id. If a file named `DEBUG` exists in the applet folder it prints debug messages as well.

Ofc I could also write them to a separate file too because it's centralized, but I don't think it's necessary for me.

fredcw:
Radio has stopped working.

```
(cinnamon:909): Cjs-WARNING **: 06:58:49.503: JS ERROR: Exception in callback for signal: NameOwnerChanged: TypeError: this.settingsData[key] is undefined
_setValue@/usr/share/cinnamon/js/ui/settings.js:480:13
bind/<@resource:///org/gnome/gjs/modules/script/lang.js:87:25
handleRadioStopped@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/RadioApplet.js:109:9
onStopped@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/RadioApplet.js:54:42
listenToDBus/<@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/MpvPlayerHandler.js:91:22
_emit@resource:///org/gnome/gjs/modules/core/_signals.js:133:47
_convertToNativeSignal@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:169:19

Cjs-Message: 06:58:49.526: JS LOG: Ignored exception from dbus method: Gio.DBusError: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.52298 was not provided by any .service files
Cjs-Message: 06:58:49.527: JS LOG: Ignored exception from dbus method: Gio.DBusError: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name :1.52298 was not provided by any .service files

```

```
Cjs-Message: 06:54:43.397: JS LOG: [LookingGlass/error] can't change playbackStatus to Playing as no channel is defined
Cjs-Message: 06:54:44.623: JS LOG: [LookingGlass/error] can't change playbackStatus to Paused as no channel is defined
```

jonath92:
Thanks for reporting @fredcw and sorry  for the inconvenience. Unfortunately I can't reproduce it and don't understand what happens.  It is obviously working on my system and I have also checked it on a fresh Linux Mint installation. In particular  I  find this line in your error very confusing: 

```
listenToDBus/<@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/MpvPlayerHandler.js:91:22
```

This is confusing because this file has been removed in the last update. 

It would be really helpful if you could provide some more information. In particular: 
- Did you restart cinnamon after the error? It might be possible that this is all we need to do to fix it. 
- Any change that you have two instances of the applet installed? The previous version and the current version? Just a guess but this might be the reason for the confusing error. 
- Can you describe in more detail what is not working? (e.g. is the icon shown in the panel? Is a popup menu shown when clicking on the applet?)

Best regards

fredcw:
@jonath92 Yes, I restarted cinnamon and now it's working. It seems for some reason the applet didn't restart automatically when I updated to the new version through the applets dialog. Maybe I messed up my installation at some time in the past which I may have done because I updated it directly from a PR once.

Love the new features btw. This applet is so much better than it used to be. It's much more convenient that you add stations directly rather than trying to find urls on a website. I was thinking that that was what was needed but I wasn't aware of the open source list of stations that you found. ("that that was what was" hehe)

jonath92:
I am glad it is working now. Your feedback is (like always) much appreciated.

PR #3762, opened 2021-04-26
[.typescript-declarations] add/update declarations

jonath92:
@Gr3q  I added/updated some declarations I need for the radio applet. Best regards


Gr3q (approved):
lgtm.

Is there a reason why you use this style for function type declarations for parameters:
`public connect(event: SliderEvents, callback: { (slider?: Slider): void }): void` ?

I'm usually using `public connect(event: SliderEvents, callback: (slider?: Slider) => void ): void`, Although I'm pretty sure it doesn't make a difference

jonath92:
No, it doesn't have a reason. Yours is better as it is shorter. I changed it now.

PR #3766, opened 2021-04-29
[typescript-declaration] add event signals to declarations

jonath92:
One of the few things missing at the moment in the declarations are the signals for the events so that you don't get auto-completion when connection to an event and bugs are not catched before runtime.  Unfortunately adding types for this is not that easy and my implementation therefore doesn't really look good but as far as I can see it is still the best opportunity for adding the signal types. But let me know if you have any other ideas for solving this problem @Gr3q (or to reject this PR if you think that not having the types is better than having a hard to read code ...). 

The arguments passed to the callback functions of the `connect` method depends on the value of the signal and of the respective actor e. g. 

**Clutter.Actor**
```
connect(signal: 'destroy', callback: (actor: this) => void): number;
connect(signal: 'button-press-event', callback: (actor: this, event: Event) => boolean | void): number;
```

**St.Widget** 
```
connect(signal: 'style-changed', callback: (actor: this) => boolean | void): number;
```

So initially I tried to add the `connect` methods to the existing class declarations but I received then the following error: 
```
Property 'connect' in type 'Widget' is not assignable to the same property in base type 'Actor'.
```

This is a common problem and a solution is to add all methods with the same name from the parent to the child class as described [here](https://stackoverflow.com/a/38026024/11603006) but this is not a good solution for this case as we would need to copy the method declarations for each subclass which are already methods only for the actor class.  
However it is possible to merge two interfaces two a type. Than again it is not possible to merge a type into a class but an Interface. Therefore the type has to be converted to an Interface again. This can be done by extending the type ... 

I have tested the declaration for both typescript applets and it is working except one Error in the weather applet but this is actually due to a Bug.  You have written: 
```
this.actor.connect("enter-event", (ev) => this.hovered(ev));
```
but the first arg is always the emitting object (see https://gjs.guide/guides/gobject/basics.html#contents)


.typescript-declarations/St.d.ts
```diff
@@ -1,9 +1,8 @@
 declare namespace imports.gi.St {
-	export class Widget extends Clutter.Actor {
-		constructor(options?: any);
+
+	interface WidgetWithoutParent {
```
Gr3q, line 3: Just call these `I{classname}` instead `{classname}WithoutParent`. They are declaring the properties going into it's class.

.typescript-declarations/St.d.ts
```diff
@@ -16,6 +15,14 @@ declare namespace imports.gi.St {
 		show(): void;
 		hide(): void;
 		style: string;
+		connect(signal: 'style-changed' | 'popup-menu', callback: (actor: this) => boolean | void): number;
+	}
+
+	type WidgetType = WidgetWithoutParent & Clutter.Actor;
```
Gr3q, line 21: Please add a comment here why this is necessary so everyone knows, like:

"this is the only way we can extend a class when its bases has different signatures` or something like this. Probably this is not the best explanation

Gr3q (changes requested):
I'm ok with this, much better than having no type declarations at all. Just a bit more code, but at least I can solve this issue at other places as well.

Apart from the minor issues mentioned in my comments, all good.

Gr3q:
And yeah, it's obviously useful because without it I would've never found that bug

.typescript-declarations/St.d.ts
jonath92: I think your comment is okay. I just added additional also link to this pr.

jonath92:
OK thanks @Gr3q. Solved your comments now. So I think @brownsr can merge it now.

Gr3q (approved):
Lgtm

PR #3776, opened 2021-05-06
New applet for Bose headphones

tuuxx:
Added new applet for Bose headphones, to manage those devices.


brownsr:
Hello @tuuxx I'm afraid I can't take what appears to be an unverifiable binary file "based-connect".  

Regards

Simon

tuuxx:
Hello Simon,
i thought it might be the easiest to deliver the user the complete application. But i got your point. 
I will made some changes, so the user can integrate it as easy as possible. 
And you're right with your assumption, it's https://github.com/Denton-L/based-connect. 

Regards 
Sammy

PR #3779, opened 2021-05-08
Clipboard eraser: erases the clipboard with one click

Techcrafter:
I added the clipboard eraser applet I created.
This applet erases the clipboard contents with one click.


brownsr:
Please take a look at the information about the file structure that is needed to put applets in, it's in the readme for the repository, and there's also information there about a validation routine you can use to check it.

I'll close for now, but do feel free to re-submit once you have got things into the right file structure.

Regards

Simon

Issue #3783, opened 2021-05-12
weather@mockturtl applet displays unit twice on vertical panel


prcr:
```
 * Applet version/Build date: 3.0.6 (2021-04-30 18:18:07)
 * Cinnamon version: Cinnamon 4.4.8
 * Distribution: Mint 19.3
 ```

**Notify author of applet**
@mockturtl 

**Issue**

The applet displays the unit twice on my vertical panel set to a width of 60 pixels:

![image](https://user-images.githubusercontent.com/60105800/117948491-15fe5880-b309-11eb-87dc-fb9d7c4916b4.png)

**Steps to reproduce**

Add the applet to a vertical panel customized to the maximum width.

**Expected behaviour**

The applet should display the unit only once.

**Other information**

I believe this is where the second unit is being erroneously added, only on vertical panels with a width of 35 pixels or more:

https://github.com/linuxmint/cinnamon-spices-applets/blob/90a130c3bfb50ba477cc02fccd2f6b09d5d8c498/weather%40mockturtl/src/3_8/main.ts#L198-L202

Gr3q:
Yeah due to the changes in temperature unit changes the logic is broken.

The idea was that the unit only shows up if you have enough space, but now it is added with `TempToUserConfig();` function by default, so it shows up twice

PR #3784, opened 2021-05-13
weather@mockturtl update 3.0.7

Gr3q:
* Resolves [#3783](https://github.com/linuxmint/cinnamon-spices-applets/issues/3783)
* Fixes wrong filename for declaration file (global.ts -> global.d.ts) causing all kinds of issues


prcr (approved):
The fix for #3783 looks good. :+1: :shipit:

PR #3790, opened 2021-05-20
Add more typescript declarations

Gr3q:
* St, Clutter and Meta fixtures are added
* Some Clutter, St, Meta declarations were added with comments.


brownsr:
what's going on around line 3556 in the first file ?  Looks like unintentional commented out code, perhaps 

Regards

Simon

Gr3q:
It is part of the markdown formatted docstring, but I was too lazy to indent it properly at that point.

PR #3795, opened 2021-05-29
[typescript-declarations] more types

jonath92:
I added more types again - nothing exciting. 

@Gr3q I added the signals to the PopupBaseMenuItem. I have declared it directly inside the class and not with the interface/type workaround as we did with the GJS declarations. The reason therefore is that with the interface/type workaround we lose the  access modifiers, i.e. all methods are always shown in the intellisense which makes it more difficult to find the public methods/attributes. The disadvantage is that we need to copy all connection methods to a subclass when we add another signal from a subclass as I have done with the PopupSliderMenuItem but I think this is not so problematic as the popupMenu module has much less hierarchy/signals as the GJS modules.


Gr3q:
I would say leave protected and private items out of interfaces and put them directly into the class if you want to use interfaces/types. This can lead to conflicts as well, but it's still pretty rare.

I say anything written in JS (like PopupMenu) just use classes because it usually has the same limitations. Redefining connect functions is fine.

PR #3803, opened 2021-06-08
[ color-picker@fmete ] : Removes redundant code which may ask the user to install a lib.

meerkut:
Removes redundant code "import numpy" in one file and the corresponding part in another that triggers checking if an error occurred while importing.


brownsr:
ping @icarter09

icarter09:
@brownsr you merge this PR.

NikoKrause:
@meerkut Why do you think numpy is redundant? I added the dependency a few years ago, because the applet didn't work properly without numpy.

PR #3804, opened 2021-06-08
[.typescript-declarations] more types

jonath92:
@Gr3q I added more types once more but this time it has some effect on our code. Most importantly I added more add Options and made them hierarchical.  As the options for the constructor are AFAICS  always a partial type of the writeable properties of the object, I summarized them in own interfaces which then can be used for the constructor and for the properties: 

Clutter.d.ts: 

``` ts
interface ActorOptions {
   ....
}

interface Actor extends ActorOptions, GObject.Object, .... {
   // readonly props and methods
}

class Actor {
   constructor(options?: Partial<ActorOptions>)
}
``` 

St.d.ts:

```ts
interface WidgetOptions extends Clutter.ActorOptions {
  ....
} 

interface IWidget {
   // needed to overwrite parent signatures
}

// as the props of Actor are in WidgetOptions and in Clutter.Actor some methods/attributes of Actor are merged twice but AFACIS this is no problem
type WidgetType = IWidget & WidgetOptions & Clutter.Actor & ...

Interface Widget extends WidgetType{}

class Widget {
   constructor(options?:Partial<WidgetOptions>)
}

```

Both typescript Applets get some errors after that. 

1.  Both applets gets an error that Type `Align` is not assignable to type `ActorAlign` but this is actually a Bug. According to the [docs](https://gjs-docs.gnome.org/clutter4~4_api/clutter.actor#property-y_align) the type of `y_align` is Clutter.ActorAlign and not St.Align. For the radio applet I could simply fix it by just removing it as it was actually not needed but I haven't tried for the weather applet. 
2. The weather applet gets an error that some properties are missing from the ButtonOptions. This is just because I made the Options now Partial instead of adding a question mark for each prop but this can easily be fixed by specifying the type to Partial which I did (but feel free to change it if you don't agree)
3. The weather applet gets an error `Property 'icon_name' does not exist on type 'Actor`. This is because in the past the child prop of button was any and now it is Actor. So it need to be typecast. 

And I have one question. Your buttons have the `url` attribute. But I don't understand where this comes from. Does this really exist in GJS?


jonath92:
> Both applets gets an error that Type Align is not assignable to type ActorAlign but this is actually a Bug. According to the docs the type of y_align is Clutter.ActorAlign and not St.Align. 

This is unfortunately not fully true. I have now tried to assign a Clutter.ActorAlign Value to a `St.Bin` instead of a `St.Align` and I received the error: `3 is not a valid value for enumeration StAlign`.  This is really annoying. It seems that the docs are not right at that point. According to [this docs](https://developer.gnome.org/st/stable/StBin.html) `St.Bin` overwrites the `align` properties from `Clutter.Actor`. So the code was correct but this makes it more difficult to implement as all classes which implements `St.Bin` and `Clutter.Actor`  must be able to override the `Align` props. So I close this PR for now and reopen it when I fixed it.

jonath92:
Fixed the Align Problem now. According to the GJS docs, the Clutter Actor properties `x-align` and `y-align` have the type `Clutter.ActorAlign` but in typescript the type must be defined as `Clutter.ActorAlign | St.Align` because otherwise  objects of type `St.Bin` can't also be of type `Actor.Align` which however is the case.  This means that all child classes of `Clutter.Actor` must specify whether the `x-align` and `y-align` props are of type `Clutter.ActorAlign` or `St.Align` separately. Annoying but otherwise not possible. Also added a detailed comment in the code.

brownsr:
are you ready to merge or are you still in discussion with @Gr3q about this ?

Regards

Simon

jonath92:
It is not urgent so I prefer to wait  for an approval from @Gr3q.  I have no converted to Draft and will make it ready for review after @Gr3q has approved.

Gr3q (approved):
I think the changes you made are reasonable, I do not see the better way to solve this.

The url property in button should not be there and can be removed. I think it's a remnant from the first version of declarations, and the property itself is just set for the button as if it were a regular object (came from a version of the weather applet before me). I really should just extend the button in the applet code.

jonath92:
> The url property in button should not be there and can be removed. 

Ok I removed the property

> Edit: Sorry for the late reply I was moving.

No problem. Take your time

@brownsr You can merge it now :-). Thank and best regards.

PR #3820, opened 2021-06-22
[radio@driglu4it] radio update

jonath92:
- add a seeker
- add own popup menu implementation as the official implementation is sometimes not that easy to extend
- improve types for mpris
- improve testing for mpv

The commit also includes some very few changes in the typescript declaration. As the changes have no impact for the weather applet, I assume that @Gr3q  has no problem with that so that Simon can merge without confirmation (if there is an unexpected problem it still can be reverted).


fredcw:
@jonath92 It's leaving a grey line on the screen. Play a station and close applet menu. Grey line appears on screen where top of menu was. Also a line appears in top left corner of screen sometimes. (tested on manjaro and mint 20.1 (daily PPA).

Also, after messing around for a bit it won't play any station and cinnamon restart doesn't fix it:

```
(cinnamon:870): Cjs-CRITICAL **: 22:40:23.928: JS ERROR: Error: initial Volume must not be undefined or null
setUrl@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/mpv/MpvHandler.js:178:23
handleChannelClicked@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/applet.js:149:20
onActivated@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/ui/ChannelList/ChannelMenuItem.js:19:39
createActivWidget/<@/home/fred/.local/share/cinnamon/applets/radio@driglu4it/4.6/lib/ActivWidget.js:12:108

```

jonath92:
@fredcw Thanks for the info and sorry for the inconvenience. 

Seems the grey line indeed occurs on Cinnamon Theme and maybe other but not on all (not Mint-Y and not Adapta Nokto). I will try to fix it asap (hopefully not to difficult 😨). 

The volume problem is really confusing. I have no idea how this can happen. You can fix this by changing the "last-volume" value to a value (e.g. 80) in the /~/.cinnamon/configs/radio@driglu4it/radio@driglu4it.json file. This occurs sometimes during development. When you stop the radio, the last volume is saved to the configs file and reloaded when cinnamon is restarted. When however due to a bug the last volume accidentally is set to null or undefined, mpv can't be started. But I don't see how this can happen in the code. Can you give me the steps to reproduce this? Anyway I can add a workaround that when the volume is null or undefined, it is applied a fallback volume so that the radio not stops working.

jonath92:
@fredcw I made now a PR which should fix the line problem and adds the workaround described in the previous comment for the volume problem. But as this is just a workaround, I am still interested in how to reproduce the issues you described. It will log a warning in the future when the fallback volume is applied. So feel free to comment if you find a way to reproduce it. But if not I guess it shouldn't be a to big problem. I will definitely have an eye on the logs for the fallback warning in the future.

fredcw:
Ok, I updated to the new version and the line bug is fixed. The other bug is still there though and it sometimes crashes cinnamon.

Here's how to reproduce it but admittedly it's a bit obscure and only seems to happen with one radio station so probably very rare.

1. play any station
2. set panel edit mode and move radio applet to different part of panel (e.g right side to middle) while still playing
3. radio stops: `Cjs-Message: 01:48:02.345: JS LOG: [LookingGlass/info] Reloading applet: radio@driglu4it/387`
4. turn off panel edit mode.
4. try to play this station (one of my presets): http://yp.shoutcast.com/sbin/tunein-station.pls?id=499023
5. Sometimes it works, sometimes the menu closes and nothing happens, sometimes cinnamon crashes, one time: `Cjs-Message: 01:43:34.863: JS LOG: [LookingGlass/warning] initial Volume was null or undefined. Applying 50 as a fallback solution to prevent radio stop working`

jonath92:
I must admit that I also have recognized unexpected behavior when moving the applet. No idea why the radio is stopping in that situation. In general streams with a `.pls` ending are different than other streams as the url loaded in mpv is actually not the `.pls` url but an url received from a server when opening the `.pls` url which had lead to problems  when restarting cinnamon in the past but I don't see how it is related to that issue. I will try to fix it on the weekend. Out of curiosity what is the name of the station from the url you provided? I want to add more default stations which are  language independent (i.e. no or almost no conversation) , free (or almost free) of advertising, reliable and with correct metadata (not easy to find stations meeting that requirments :thinking:). Your url seems to meet this requirements ...

fredcw:
I got the station from the applet's list - "Deep House"

I don't know if there's a way to prevent the applet being reloaded. It seems to be done in usr/share/cinnamon/js/ui/appletManager.js
Also, when an applet is moved to say the right side of the panel, all the applets on that side get restarted and this also seems to cause the bug. i.e. I moved a different applet to the part of the panel with the radio applet and it caused it to be restarted.

It seems that applet.on_applet_reloaded() is called before reloading (line 312 in appletManager.js) Maybe you could implement this method in the applet to cleanly prepare the applet for removal before it's reloaded.

Gr3q:
Form my experience, on_applet_removed is always called when an applet is removed, and moving something in the panel removes and readds everything in the affected sections, so that's your best bet.

The applet after moving is actually a new instance, but it might've changed since I had to deal with the consequences of moving (in the weather applet my loops persisted from the removed applets exhausting people's API quotas)

jonath92:
Thanks for your comments. This really helped a lot. 

> The applet after moving is actually a new instance, but it might've changed since I had to deal with the consequences of moving (in the weather applet my loops persisted from the removed applets exhausting people's API quotas)

This hasn't changed. It seems that all widgets (e.g the popup menu and the applet) are getting automatically destroyed but not listeners to dbus events. In my case this lead to the problem that each time new instances of listeners to mpv events have been created but the old listeners were still active and most problematically the listeners triggered functions which tried to mutate the destroyed widgets. This lead to a lot of errors and could even lead to crashing cinnamon when dragging the applet multiple times. I am not 100% sure but I guess this lead also to the volume = null problem. 

> Edit: It seems on_applet_reloaded(), on_applet_removed_from_panel() and on_applet_added_to_panel() are all called in that order.

Yes that's helpful because I need to differentiate when the applet is moved and the applet is permanent removed from the panel (e.g. by clicking on 'Remove Radio++' in the context menu) and is in the latter case the radio shall be stopped but not in the first case. As the `on_applet_reloaded` function is only called when the applet gets moved, I can easily handle the both use cases in the code. 
 
Unfortunately there are still some errors in the log when dragging or removing the applet, which I don't know where it comes from but I will investigate further.

jonath92:
> Unfortunately there are still some errors in the log when dragging or removing the applet, which I don't know where it comes from but I will investigate further. 

I now tested in a virtual machine once more and it seems that the error messages are mostly gone. So it seems to related to my system which is not a big problem.

PR #3827, opened 2021-06-27
[radio@driglu4it] Fix Problems when moving applet

jonath92:
- In the past the mpv listeners haven't been deactivated when the applet has been moved or permanently removed from the panel which lead to the problem that the applet has been tried to access destroyed widgets which even could lead to cinnamon crashing. This has been fixed  
- add another default station
- prevent opening multiple installation windows when clicking on the applet while an installation is already in progress


fredcw:
@jonath92  I don't actually listen to that deep house station much, I think this one is a lot better: Deep House Radio http://62.210.105.16:7000/stream . It's in the applet list. It's also the most popular deep house station on internet-radio.com https://www.internet-radio.com/stations/deep%20house/?sortby=listeners

jonath92:
ok I will change it net time I update the radio

PR #3843, opened 2021-07-13
weather@mockturtl 3.1.1 Hotfix

Gr3q:
~~Some esnext syntax was mixed into the bundle, temporarily fix bundle manually until I have time to
fix bundling.~~

Some more details:
[Object spread](https://caniuse.com/mdn-javascript_operators_spread_spread_in_object_literals) is an EsNext feature like `{...otherObject}`, which is present in the bundled output from the luxon library. ~~I removed it manually for now.~~

This is now a full fix, bundling is fixed.

@brownsr I also would like to add `package-lock.json` to the gitignore list, I don't see any benefit of keeping it, it's also polluting the PRs quite a bit.


Gr3q:
@brownsr this is a P1 because the applet is not working on half of the released Linux Mint versions.

PR #3844, opened 2021-07-13
[Cinnamenu@json] Allow selecting default skin tone

markkrj:
The emoji picker on search is an awesome feature, however, I was a bit annoyed that it didn't keep the skin tone, like chatting apps do. So I did this mod to at least allow selecting a default skin tone.

ping @fredcw 

Edit: adding some screenshots
image
image

PR #3845, opened 2021-07-14
[Cinnamenu] Add more theme tweaks.

fredcw:
+ some code refactoring

### DIFF
--- a/Cinnamenu@json/files/Cinnamenu@json/4.0/applet.js
+++ b/Cinnamenu@json/files/Cinnamenu@json/4.0/applet.js
@@ -19,7 +19,6 @@ const {getAppFavorites} = imports.ui.appFavorites;
 const {TextIconApplet, AllowedLayout, AppletPopupMenu} = imports.ui.applet;
 const {PopupResizeHandler} = require('./resizer');
 const {AppletSettings} = require('./settings');
-const {addTween} = imports.ui.tweener;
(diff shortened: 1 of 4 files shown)


markkrj:
@fredcw @brownsr this PR reverted the changes I made in #3844. @fredcw do want me to create a PR on your repository, as it seems that is upstream to this? Otherwise you should merge changes from here before creating pull requests...

fredcw:
@markkrj Sorry about that but I changed my mind. Having given it some thought, I really want to keep the number of user options as low as possible because for each addition setting, it's just that little bit harder to find the setting you're looking for and may give the impression that the applet is too complicated for people who don't have time to read all the options. And since no one else has requested this feature, I can only assume that few people would need it. I have however thought of an idea that may be better - put recently/frequently used emoji in the recents category. This may make it even quicker to find the emojis you frequently use and wouldn't require an additional option in settings.

markkrj:
Well, it didn't change any default and added no overhead to users at all. The only thing it did was present an option to change skin color and the option disappear if user disable emoji search. Did you try the applet with the patch? Like, every emoji picker I knew keep last used skin color, but my knowledge is limited to implement such feature, thats why I added that option. Anyway, you're the author, so I'll respect your opinion. I'll keep my fork and will merge your changes occasionally. Thank you very much for coding this applet, it's wonderful.

PR #3847, opened 2021-07-14
Cinnamenu@json: Update sv.po

eson57:
Update Swedish translation from POT


eson57:
Sorry for the mess! The heat is terrible right now. ;)

PR #3856, opened 2021-07-19
Update RU translation for weather

dmiwell:
Now it's 100% based on template. Outdated strings are removed. But there are still some issues in the code like:
```
_("As of {lastUpdatedTime}", { "lastUpdatedTime": lastUpdatedTime })
```
Template generator does not handle it correctly


Gr3q:
Please still include the Russian translation in this PR, but don't include the .pot file, because it is missing lines (which is not actually your fault).

There are 2 issues with the .pot file:
* Poedit won't be able to generate a complete .pot file, you should use the script `cinnamon-spices-makepot` script in this repo for it which generates correct translations.
* `cinnamon-spices-makepot` won't generate a complete .pot file at the moment - because I made changes how the code is built/bundled in a way that makes the script skip lines, which will be fixed in https://github.com/linuxmint/cinnamon-spices-applets/pull/3857.

The lines in question:
https://github.com/linuxmint/cinnamon-spices-applets/pull/3856/files?file-filters%5B%5D=.pot#diff-4d9caa08127e2eba576ff5c1da06703712ddf32a4a5ea6387c5d334b243651f3L1390-L1644 are actually used by the applet.

dmiwell:
@Gr3q 
Ok. I used `cinnamon-spices-makepot` script to generate the pot file. I thought it works fine and generates latest actual template.
I reverted the pot file back.

dmiwell:
@Gr3q give me some time then and I'll translate the strings added by the pot version which is in master now

dmiwell:
@Gr3q I've added translations from reverted pot file, so it's ready to be merged, if you don't see issues with the po file itself

Gr3q (approved):
all good

PR #3890, opened 2021-08-16
radio@driglu4it: Update pot file

eson57:
Fix some missing translation strings and errors


eson57:
There is also two strings from `4.6/RadioSearchWidget.py` that I couldn't get to work.


```

#: 4.6/RadioSearchWidget.py:48
msgid "Find Radio Station"
msgstr ""

#: 4.6/RadioSearchWidget.py:53
msgid "Add to My Stations"
msgstr ""
```

PR #3893, opened 2021-08-18
[radio@driglu4it] Update readme

fredcw:
@jonath92 Just thought I'd fix a few grammaroos for ya.


jonath92:
Thanks, that's great.  I am always struggling with english grammar :D.

PR #3904, opened 2021-08-24
Update translate pl

szedar14:
I translated into Polish at https://cinnamon-spices.linuxmint.com/applets/view/17
I checked and everything is fine.

I do not know if I added it in the right place, if not, please write where I should send the correction of the translation into Polish.


Gr3q (changes requested):
It's all good, except for some reason you readded the file to ` weather@mockturtl /files /weather@mockturtl /po /pl.po`, with extra spaces in the foldernames.

It should be ` weather@mockturtl/files/weather@mockturtl/po/pl.po`

szedar14:
@Gr3q  I don't know why it happened, I did edit the pl.po code and I pasted my translation and it should be ok.
What should I do in this case to make it work properly?

PR #3910, opened 2021-08-29
Remove author name

fredcw:
Sorry but I can no longer be associated with cinnamon or mint if you're going to tolerate the bullying and abuse of children by members of the mint development team. I shall have no further association with you.

### DIFF
--- a/Cinnamenu@json/info.json
+++ b/Cinnamenu@json/info.json
@@ -1,5 +1,5 @@
 {
-    "author": "fredcw",
+    "author": "",
     "original_author": "jaszhix",
     "license": "GPL-3.0"
 }


JosephMcc:
You are obviously well within your rights to remove your name for any reason. But I'm curious as to what bullying and abuse you are talking about?

fredcw:
The specific comment I'm referring to is this one: https://bit.ly/3mQ9D6F

Obviously I'm not going to presume to tell @clefebvre what to do as he is a far more reasonable and level headed person than me but I can only be honest and say that I no longer feel like I want to have anything to do with linux mint if this kind of behaviour is regarded as acceptable. This poor kid will probably never want to have anything to do with linux ever again if the first time he tries to contribute, he is spoken to in this manner. I'm disgusted and sickened quite frankly.

For me, this comment is beyond the pale and it's just one of a stream of comments from the same person which seem unnecessarily rude and abusive.

https://bit.ly/3kGfRmV
https://bit.ly/3zxBcF2
https://bit.ly/3mLZqYS
https://bit.ly/2Wv0Igc
https://bit.ly/2Y4DJsx
https://bit.ly/2Wxowzu
https://bit.ly/3kA0bS6
https://bit.ly/3ysP5mS
https://bit.ly/3DvsXMb
https://bit.ly/3zpK7sa
https://bit.ly/3DxEvhV
https://bit.ly/38m5cYN

clefebvre:
Thanks for clearing this up. Before shining in on this I have to say I'm relieved. After reading "child" and "abuse" I had imagined much worse.

A little clarification which is somewhat important. Though @leigh123linux is part of the Cinnamon team, he is a Fedora maintainer, not a Mint developer. I am myself in charge of both the Cinnamon and the Mint projects, so I do nonetheless have a little bit of authority in the matter.

@leigh123linux is an important and valuable member of the Cinnamon team. He is appreciated by the other members. He is known for sometimes lacking tact and being a bit abrupt. Some people in the team want to please everybody and don't know how to say NO, for Leigh it's sometimes the opposite. Nobody's perfect and we need to address that, but that doesn't change the fact that he's welcome in our team and that we're glad he's here.

If I look at the comments @fredcw linked above, I'd say many lack tact, some are a bit rude, but only a few are actually harmful. Ideally none of them should be, ever. I'll talk to Leigh about it to see if we can address this. I'll also try to clarify the fact that outside of Cinnamon, Leigh is a Fedora maintainer, not a Mint developer.

All in all, I think this is blown out of proportion though. If I look at the triggering incident: https://github.com/linuxmint/cinnamon/pull/10292#issuecomment-906957457

I mean, the comment is silly, but so is this reaction. I don't know the age of @Varao-Br, I don't see abuse and I don't see bullying. I don't see any indication that Leigh was rude because he was a child (assuming he is). It would never have had occurred to me that this could be interpreted as child abuse. I sure would like everybody to always be welcomed with open arms and with a smile, but more importantly if we want to continue to work well we need to calm down on both sides here and be pragmatic.

PR #3919, opened 2021-09-05
[radio@driglu4it] update tests

jonath92:
The tests have been updated so that there are no falling tests anymore and it is used now an own test-runtime environment which is hosted in an own repository (https://github.com/jonath92/cjs-jest-test-runtime).


brownsr:
Just to explain the delay.  The applet file has too many differences to review in github, so I will have to do it manually on my machine.  Github appears to think it is one massive one line file.  I've noticed this before, but let it pass.  This time I think I need to review it, as I suspect there is quite a bit of change there, cumulatively certainly.

Regards

Simon

jonath92:
Thanks for the Update. Actually there souldn't be much change in the applet.js file as the changes were almost all only related to the Test files.

brownsr:
I see that the applet is tiny, and it looks like all the code is in radio-applet.  That right ?  Anyway radio-applet appears to be a 1 line file, generated perhaps.  Unfortunately that makes it unreviewable as it stands,  I imagine it is missing all the normal line terminations, or something of that sort.  Can you provide something that can be reviewed ?

Regards

Simon

jonath92:
> I see that the applet is tiny, and it looks like all the code is in radio-applet. That right?
Yes that's right

I changed now the webpack configs and changed lodash with lodash-es as in that way not the hole lodash package is included in the output file (which decreases the file by 4000 lines :D).

PR #3922, opened 2021-09-07
ipindicator@matus.benko: Use ip command to retrive list of interfaces.

cquike:
The legacy tool ifconfig is not installed by default in many distributions and has been replaced by "ip".
Use that tool to get the list of network interfaces.


brownsr:
ping @PrimaMateria

PrimaMateria:
@cquike thank you for PR, I was not aware of the new "ip" tool.
@brownsr all good, please merge

PR #3969, opened 2021-10-02
calendar@ccprog: update pt_BR.po

marceloaof:


marceloaof:
Hallo Herr Colloseus
Please don't you worry regarding quality, your applet is going to be well represented, since I tested the translations and the applet side by side and I have experience in translating software. Here I have fixed typos in the original translation, I added missing translations, and indeed few decisions regarding being more fluent or more precise. I understand your review, because I tend to prioritize precision actually, in order to respect the original term, the innovation. When necessary, I offer different words to keep the translation of the idea unchanged. Both verbs "consider" and "mark" are written similarly in portuguese, as "considerar" and "marcar". However the following phrase "Marcar como dias de fim de semana" is not similar to "Markiere als Wochenendtage". Actually it is similar to saying "Marcar em algum dia do fim de semana", which means "Planen Sie das Treffen am Wochenende"/"arrange the meeting on the weekend". I do not remember reading or hearing the current translation, so I suggested "considerar como dias de fim de semana", which retains the original meaning.   But 'yes', the current translation is understandable (after experiencing it). I would rate it 6 out of 10 the need to change the current translation (in brazilian portuguese).
I understood your explanation regarding saturday and sunday being exceptions, therefore I undid my suggestion regarding these strings. Please accept and disregard the portions you consider necessary. I am going to reflect on other suggestions of mine offered here.

PR #3981, opened 2021-10-09
rswitcher@nixdev.com: Adding rswitcher to repo

mindbeat:
Hello friends, 

I wrote up a resolution switcher to make it more easy to go from 4K HiDPI 150% Fractional scaling to a 1080p resolution. 

This applet is the result.

Right now it only uses the primary display, and won't work with a dual setup. But that's my next feature to implement. 

So this applet rocks the boat here. Now I can play OpenRA for example without open "Display" and fiddeling with display settings everytime. Just set it to 1080p from the resolution switcher applet and start the application. 

The applet uses the bash-script rswitcher that is in the same folder. Its possible to run this separate in shell as a command. Can be useful if you want to connect a resolution to a shortcut key. 

Hopefully someone find this useful.


mindbeat:
Now it seems to work.
 ./validate-spice rswitcher@nixdev.com
No errors found. Everything looks good.

PR #3995, opened 2021-10-23
googledrive@pbojan: Add a new Google Drive applet

pbojan:


brownsr:
There looks to be a spelling mistake 'configuraion' instead of 'configuration' in a few places in the code.

PR #4002, opened 2021-10-25
Improve typescript declarations

Gr3q:
* Fix Enums, they are generated uppercase
* Now docstrings are also generated with links and deprecation support - the majority of new lines from this PR
* Add new generated declaration files (Cinnamon, CinnamonDesktop, Clutter, St, Cvc, HarfBuzz, Caribou, GModule)
* Fixed generated formatting
* Null support in generated files
* Support for named array return types
* Support for extending declaration files without touching generated text
* And much more

Generated with https://github.com/Gr3q/GIR2TS

@jonath92  I updated the Type npm packages as well, could you try them to see if there are any problems with it? 

~They are not quite final, because I got a hold of a gir file for St too https://github.com/Gr3q/GIR2TS/tree/st-typings, and now I'm looking for a non-1.0 version of the Clutter gir file. Once that's done i will have to rebuild the constructor options/etc for St, because it is a mess.~


jonath92:
@Gr3q This is awesome. I have some minor problems which are mostly related to connecting to events, e.g. I get an error when trying to connect a St.Widget to the `key-press-event` .
image

In this regard there is also some other weird thing. I have written this is in my code:
```
   widget.connect('key-press-event', (actor, event) => {
        const symbol = event.get_key_symbol();
```

This works without problem. But it gives an error with the new types. However when looking in the docs, this indeed shouldn't work as the event is a `Clutter.KeyEvent` and not a `Clutter.Event`. Not sure how to deal with that. 

One other thing I noticed in the new types is, that in the GTK defintions, the constructor usually (always?) doesn't have any arguments but that's not the case in real. It should be the same as for `St` (passing the props in an object).

Gr3q:
I think the signal issue is either that I didn't update the inheritance chain properly in St or I have to study a bit how 'union' nodes work in .gir files, maybe I'm missing some key field during generation from it for inheritance (Event is an union node).

As for the constructor, it is not documented in the .gir files, so I'm generating them myself, empty.

If we can find the patterns with trial and error on which namespaces has constructors like this, exactly which properties are included as options and why I can include them to try to automatically generate the options as well

Gr3q:
@jonath92 I'm going to update the npm packages again soon, but all seems to work now. Might be not correct that I add init options for everything, but it works quite well.

Gr3q:
@jonath92 published, can you validate everything works as expected?

Events can be used properly, initializers etc.

The only downsides:
* There is no rule yet to work out where the owner is added to the callbacks in connect functions (if not all, I haven't tested it actually)
* Everything has constructor options and probably that's not true

jonath92:
> There is no rule yet to work out where the owner is added to the callbacks in connect functions (if not all, I haven't tested it actually)

I am pretty sure that the first argument in the callback function is always the owner because this is mentioned here: https://gjs.guide/guides/gobject/basics.html#connecting-signals. There it is said: 

> Signals often have multiple callback arguments, but the first is always the emitting object

I still get some errors but I think all of them are related to two problems: 
- The new types often forces to provide a return value for the passed callbacks for events, e.g. this gives an error:
-   
```
    widget.connect('button-release-event', () => {
        global.log('button clicked')
    })
```

This can be fixed by adding `return true`.  I know that in the doc it is said this this is actually required but I usually have omitted this in my code and never had any issues with it. I am not sure but I guess it would be okay to add `| void` to the return types of the callback functions ? But if you disagree, I also can live with always providing the return value ;-). 

The `Imports.gi.GObject.IObject.disconnect` method is AFAICS not right. Because the argument is specified as string but the return value of the `connect` functions is always a number and the disconnect method is used to disconnect a signal by the id returned from the connect function, e.g. this should not give an error: 

```
    const signalId = widget.connect('button-release-event', (actor, event) => {
        return false
    })

    widget.disconnect(signalId)
```
See: https://gjs.guide/guides/gobject/basics.html#connecting-signals. 

In this regard I also find the JSDoc a little bit confusing. 

image

Gr3q:
> This can be fixed by adding return true. I know that in the doc it is said this this is actually required but I usually have omitted this in my code and never had any issues with it. I am not sure but I guess it would be okay to add | void to the return types of the callback functions ? But if you disagree, I also can live with always providing the return value ;-).

If it requires a return value by the docs I'd like to keep it that way. I'll generate better docstrings to describe the callbacks better so it is obvious what it does.

as for `disconnect`, I can just simply fix it with modifiers during generation. It would be worthwhile to check if it actually uses `...ids: number[]` or `id: number`

Gr3q:
Should be better now, packages are updated as well.

Gr3q:
@brownsr This is not ready for review. @jonath92 please approve it if you agree, else request changes.

The declarations are dramatically improved, generated for critical libraries like St and Clutter with the version Cinnamon actually uses.

There are still some issues stemming from the .gir files themselves, like:
* some static constructors does not return the correct type, but the base type, this needs manual review from my side because the .gir files themselves are incorrectly generated.
* I have to review how do I generate declarations for callback functions
* I have to review do I need to change how do I generate declarations virtual functions
* write-only properties are not included for now because there is no equivalent in typescript
* Will have to review how I handle `{Classname}Class` and `{Classname}Private` nodes during generation. the former mostly contains protected elements, and the latter private elements for `{Classname}`.
   This only causes issues (meaning: no access to virtual functions) if you'd like to extend the class. Disclaimer: I don't see an example that someone does this because one must register the new class with GObject before initializing it.

jonath92:
LGTM

Gr3q:
@brownsr now this can be reliably used by basically anyone who wants to write stuff for cjs, except for:
* missing stuff from global or missing globally declared stuff
* missing declarations from injections to types (which I don't know of)
* Incomplete declarations for JS side of cinnamon.

It's getting to the point where it's also too big, and there is the problem that I can't use it in the extensions repo unless it's directly included there as well, but that doesn't really make sense.

jonath92:
> It's getting to the point where it's also too big, and there is the problem that I can't use it in the extensions repo unless it's directly included there as well, but that doesn't really make sense.

I don't understand why you are not using your node packages :thinking: . They have so many advantages (e.g. version control, usage for different kind of applications/platforms, ..). The only disadvantage I can think off, is that is more complicated to set up but this could be simplified by providing templates (I actually already have done something for applets: https://github.com/jonath92/cinnamon-template-applet. It is not perfect as it is still requires some manual steps but I find it helpful). I really don't understand why webpackage (or other bundling tools) are so rarely used for Cinnamon spices and Gnome Shell Extensions as this makes code sharing so much easier :disappointed:

brownsr:
There's one file for clutter that is too big to review in github.  And there's another few reworkings of the interfaces to standard libraries that I am taking as read.

It's definitely too big.  Do remember that I need to scan all submitted code to ensure there's nothing malicious there.  As these are declarations the risk is minimal, but this is getting very unwieldy for me.  I am tempted to reject this because of the over-large clutter declarations file, but I'm giving this a pass for now.  I'm sure others in the team are looking on a bit bemused at all this work you are doing to be able to use typescript.  I'd point out that there is a maintenance burden whenever the interfaces to the functions you are using change.  I see that these appear to be script generated, so that will make that easier.  But keeping code up to date as the world changes around it is a non-trivial task.

Anyway, as I can't see anything harmful here, I am merging this.  I presume that any further changes from conversations between you will come through as another PR.

Regards

Simon

Gr3q:
@jonath92 I'm not using it here because if I remove the declarations from this repo I have a feeling I might not be allowed to use it. But it is obviously not scalable, so we need a solution for this soon.

@brownsr 
On the declarations, why I'm doing all the work and why they shouldn't be bemused there are a multitude of reasons:
* To write applets for cinnamon we have to use Javascript which is a terrible language, hence I'm using Typescript which is bearable.
* To know how use the libraries in cjs I have to refer the **Gnome** documentation because Cinnamon's documentation is non-existent (on this, definitely), posing the biggest challenge for most people to write **anything** useful.
* It didn't take long to fix up the tool (about 8 hours), and updating the declarations now takes less than 10 seconds (that could be automated), any time beyond that goes to me finding the correct machine-readable documentation files - _.gir files, documentation and other things are often generated from them_ - which are either not included with Cinnamon or they are dumped into random non-standard locations during your packaging. Figures that's the reason why you can't find any useful generated docs for cinnamon. The remaining part of my time goes to fix errors in the generated declarations which are coming from the .gir files and those errors are most likely coming straight from C source code.
* On top of these declarations providing autocomplete support and documentation in you IDE for typescript projects, they can also do the same for regular JS projects, you only need to dump a `tsconfig.json` into your folder including the declarations and the `allowJs` flag.
* I'm not doing this work just for myself ofc, I want to provide this to anyone who want's to write anything for Cinnamon in JS so they won't have that absolute horrible experience starting out as I had.
* Now that the tools is working I can also install Gnome, get all the .gir files for that (for Gnome they are included and placed in the correct location) and generate declarations for that as well in its entirety in one go.

That is all, really.

brownsr:
OK, good to know.

Regards

Simon

PR #4045, opened 2021-11-16
ddcci-multi-monitor@tim-we - Make sure code is es2017 compliant

Gr3q:
App can't be run on older versions on CJS.

This fix makes sure it is compatible from ES2017+ (cjs version 52 and up)

@tim-we


tim-we:
Sorry about that. Thanks for the fix :+1:

PR #4054, opened 2021-11-21
[radio@driglu4it] update types

jonath92:
just a small update to apply the new types

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/4.6/radio-applet.js
+++ b/radio@driglu4it/files/radio@driglu4it/4.6/radio-applet.js
@@ -287,18 +287,24 @@ function createActivWidget(args) {
     widget.can_focus = true;
     widget.reactive = true;
     widget.track_hover = true;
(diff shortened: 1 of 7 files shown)


Gr3q (changes requested):
Let me update the declarations for you, `notify` callback expecting a number return value is a mistake in the generated code.

jonath92:
> Let me update the declarations for you, `notify` callback expecting a number return value is a mistake in the generated code.

Ok thank you :-)

jonath92:
There is also another problem in this pr. The popup menu doesn't open anymore when clicking on stop.  I need to investigate. So right now this definitely shouldn't be merged

Gr3q:
If you return `true` in callbacks which are expecting it I think that means the event has been handled, no need to pass it up the chain.

Previously you didn't return anything (well, undefined), which was interpreted as `false`. If you return false I think you will be fine.

jonath92:
> Previously you didn't return anything (well, undefined), which was interpreted as false. If you return false I think you will be fine.

yes, thanks. That fixed it

jonath92:
@brownsr from my point of view, it can now be merged :-).

PR #4082, opened 2021-12-05
[octopussy@centurix] Use URL from settings instead of default

oerkel47:
When opening the webcam stream or the octoprint page it will use the hardcoded url "octopi.local" even if a different url is set. This PR fixes it.
@Centurix


Centurix:
Thanks!

Centurix (approved):

PR #4097, opened 2021-12-11
IP-Indicator: updating IP services 

cinnamintdev:
Updating IP services

https://bot.whatismyipaddress.com/ has been discontinued.
Replaced it, plus added two other good services.

@PrimaMateria


PrimaMateria:
Approved and @cinnamintdev thanks for these updates

PR #4105, opened 2021-12-12
Mini-Calc applet (calc@ptandler)

ptandler:
I converted the Mini-Calc desklet to an applet; seems a bit more handy:

Very simple and minimalistic but yet powerful calculator that evaluates JavaScript expressions.
I developed it because I was missing a very quick and easy way to do some simple calculations. And JS turned out to be quite handy ;-)

Now you can define a keyboard shortcut to open / close the Mini-Calc (defaults to Calculator key and Ctrl-Super-Alt-C)
Also: added a wee bit of documentation in the applet


ptandler:
@brownsr thanks for merging! :smile:  How long does it take until it shows up at https://cinnamon-spices.linuxmint.com/applets/latest?

ptandler:
@brownsr I'm just wondering as Mini-Calc still does not show up on https://cinnamon-spices.linuxmint.com/applets/latest if I made an error (e.g. in the meta info)?

Might it cause an issue if I use the same ID for the Mini-Calc desklet and Mini-Calc applet?

As far as I remember, the desklet did show up in the Spices site immediatle when it was merged.

Issue #4113, opened 2021-12-13
batterymonitor@pdcurtis Feature Request: Display power draw from the battery


joka42:
**Notify author of applet**
@pdcurtis

**Request**
Add a mode which shows the current power consumption of the laptop instead of the current battery percentage.

I was about to write a new applet for that, but it might be a good idea to integrate it into the battery monitoring applet that already exists. 

This is what it would look like:
![image](https://user-images.githubusercontent.com/68659955/145842672-e262d7a4-63f0-47bd-a7c8-6d40ffc203cf.png)

Why is that useful/interesting?
You can check how much power is drawn from your battery for different activities (such as idling, browsing, compiling, watching videos, etc.) in real time. I find this quite interesting, but also helpful when there are processes running in the background and there is more power drawn from the battery than you would assume.

If you are interested in integrating it, but don't want to do the work, I would contribute to the applet. Otherwise I'd write a new one. 

Thanks for considering.

joka42:
new applet created.

PR #4131, opened 2021-12-17
batterypower@joka42 Adds an applet that shows power consumption of the system drawn from the battery

joka42:
Adds a applet that shows power consumption of the system drawn from the battery.

image

Feature is missing from batterymonitor@pdcurtis applet (Issue https://github.com/linuxmint/cinnamon-spices-applets/issues/4113), but request was ignored.


ptandler:
I like the idea of being able to see current power consumption. If this helps to reduce consumed energy a little bit, that would be at least a little step towards our CO2 goals WRT the climate crisis...

PR #4158, opened 2021-12-22
[radio@driglu4it] update

jonath92:
- The tooltip includes now besides the volume also the current channel, the current song title and the currently downloading songs
- [yt-dlp](https://github.com/yt-dlp/yt-dlp) has been added as second option for the youtube download (thanks @bamper45 for bringing this to my attention)
- It is now shown a second icon while downloading from youtube
- The readme and the station list have been updated

@brownsr I am sorry for that many changes but I don't know how else to do this. Most changes are in the radio-applet.js file and a very hugh part of these changes are caused because git is not smart enough to see that the `deepEqual` function loaded from lodash has moved in the `radio-applet.js` file. TBH I think you don't need to read this in order to ensure that I haven't put any malicious code in it. It is sufficient to read the typescript files in which I have made much less changes and the cinnamonPopup which I am referencing which is here: https://github.com/jonath92/CinnamonPopup/blob/master/index.ts. When you run `npm install` and `npm run build` (right now this will fail due to some wrong types in the ci-types/cjs node package but i will make a pr later this day to fix this) you will get exactly the same radio-applet.js file.


jonath92:
Now building without an error

fredcw:
Not really a problem but something maybe worth fixing at some point: tooltip pops up behind applet
Screenshot from 2021-12-23 19-18-22
.

jonath92:
@fredcw Yes that's odd. Should now be fixed

Issue #4195, opened 2021-12-31
batterypower@joka42 no such device


wennaspeedy:
```
error t=2021-12-31T15:23:01Z Failed to read from file “/sys/class/power_supply/BAT0/current_now”: No such device
error t=2021-12-31T15:23:01Z 
[batterypower@joka42]: Applet batterypower@joka42: Could not create applet object.
[batterypower@joka42]: Error importing applet.js from batterypower@joka42
```

After reboot.

fredcw:
Ping @joka42

joka42:
@wennaspeedy thanks for reporting the issue!

Has it worked after installing the applet? Does the error also occur when you restart cinnamon when the system is already up and running (`Alt+F2` then `r` and enter)?

wennaspeedy:
> 

Yes, it worked after installing and works after restart cinnamon.

joka42:
@wennaspeedy the merge request should fix it, can you confirm?

wennaspeedy:
> @wennaspeedy the merge request should fix it, can you confirm?

It seems ok. Thanks.

PR #4198, opened 2022-01-02
batterypower@joka42 added Upower as source, if battery files are not found

joka42:
added Upower as source, if battery files are not found.

Updating upower is very expensive and is not the best solution. using the files as first option is much better, if they are available. The applet seems to draw 1W if upower is used.


joka42:
I removed the update during initialization of the applet which should fix #4195 

However, I can not test it, because I could not reproduce the error.

PR #4218, opened 2022-01-10
[radio@driglu4it] version 2.0.2

jonath92:
- Integrate cinnamon popupmenu directly in the code again. I still think that the official popupmenu implementation is a mess and that we need something better but right now it is easier to maintain it that way for me
- Override `_meta` with more globally useful stuff which doesn't change during the lifetime of the applet (e.g the `monitor` is used several times in the code)
- Show better notification message when radio doesn't start (thanks @Andriamanitra)
- Update readme
- Update allStations
- Fix: radio not working when adding a station with a space at the start or end
- Update default stations: replace "chillout" with "Smooth Chill" as the url is not working anymore and add "Radio KIKS Rock" and "Radio BOB"


jonath92:
I think there is a mistake. Closing it for now

jonath92:
Ok now working. I didn't consider that the applet is not reloaded when users change the monitor which lead to the problem that the popup menu appeared on a wrong monitor.

PR #4219, opened 2022-01-11
calc-js@ptandler: Changed UUID from `calc@ptandler` to `calc-js@ptandler` to be really unique

### LINKED ISSUE #4177: calc@ptandler: Mini-Calc Applet does not show up Spices (website and settings app)
```
 * Applet version/Build date 1.0
 * Cinnamon version (5.0.7)
 * Distribution - Mint 20.2
 ```

**Issue**

@brownsr was so kind to merge the PR #4105 with the initial version of my Mini-Calc applet. However, it does not show up on https://cinnamon-spices.linuxmint.com/applets/latest and the "Applets" settings on the download panel.

I wrote the previous version as desklet and there it [showed up immediately on Spices](https://cinnamon-spices.linuxmint.com/desklets/view/60) once it was merged. So I wonder, what is different here?

Could it cause problems that both the Mini-Calc desklet **and** applet share the same UUID `calc@ptandler`? From what I saw and experienced so far, I had the impression that the UUID is scoped to desklets / applets / extensions. Is this assumption correct?

Or what else could cause this?

ptandler:
Maybe this fixes #4177 (applet not visible in online listing) ...?


ptandler:
added another feature: add option to convert locale number format (e.g. `1,23`) to JS notation (i.e. `1.23`)

ptandler:
Actually, I'm a bit surprised that this PR lists all 22 commits. Most of the should have been merged by PR #4105 ... why do they appear again?

PR #4238, opened 2022-01-24
[radio@driglu4it] HOT FIX - Fix config not working on fresh install

jonath92:
In the last pr I have accidentally used `ur` instead of `url`. This lead to a type Error each time the applet makes some changes in the settings for new users (which is very often the case and makes the applet almost useless for new users). 

@brownsr when possible please accept this soon. There are just very few changes.

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/4.6/radio-applet.js
+++ b/radio@driglu4it/files/radio@driglu4it/4.6/radio-applet.js
@@ -2564,7 +2564,9 @@ const createConfig = () => {
     appletSettings.bind('initial-volume', 'customInitVolume');
     appletSettings.bind('last-volume', 'lastVolume');
     appletSettings.bind('tree', 'userStations', (newVal) => {
-        const trimmedStations = newVal.map(val => { return Object.assign(Object.assign({}, val), { url: val.url.trim() }); });
+        // temporariy solution to fix typo in settings-schema
+        // @ts-ignore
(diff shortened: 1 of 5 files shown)


fredcw:
The applet description in the applets dialog is still "A simple radio applet for cinnamon". Maybe this undersells it a little bit now? How about "A radio applet with 1000's of searchable stations" or something similar?

Also, I just tried removing and reinstalling the applet but the download dialog didn't appear, only a notification "Couldn't start the applet. Make sure mpv..." I don't remember doing this last time. Is this intended? I'm using manjaro.

jonath92:
@fredcw yes that's a good idea. I will Update the description in the next Version.

Currently mpv only gets installed automatically on Systems with apt but I agree that it would be nice to also handle pacman

fredcw:
That's strange, I sure it installed it automatically last time. And how would apt install .mpris.so into ~/.cinnamon/configs anyway? And pacman wouldn't help with that either. I would have thought that you would download it directly from https://github.com/hoyon/mpv-mpris/releases/download/0.6/mpris.so and that that would work on any distro?

jonath92:
Ah ok I see what you mean. There are two Dialogs - one for installing mpv and one for the mpris plugin. And the mpris plugin indeed should be downloaded on any distro. I am currently just on the Smartphone so I can't test but I don't see any reason in the Code why it should not work. I will look into it in more detail tomorrow.

jonath92:
@fredcw it definitely works on Linux mint. No idea why it doesn't work on manjaro. I will test it on manjaro tomorrow.

PR #4247, opened 2022-01-31
[Cinnamenu@json] Show slightly fewer recents

fredcw:
\+ show web search option before web bookmarks in search results


fredcw:
Going to close this PR and open another due to conflict with another PR. (file tr.po)

PR #4260, opened 2022-02-13
 airstatus@vadgus new Applet: AirPods Battery Status

vadgus:
AirPods Linux Battery Status (Ubuntu, Cinnamon)

You can check the battery percentage of your AirPods and their charging case.

When the battery charge of your AirPods is low, you get a notification on your desktop. You get notifications when the charge is less than 10 percent.

You can change check frequency and disable notifications.


vadgus:
Tested, not ready yet.

PR #4263, opened 2022-02-16
clean-show-desktop@filipetorresbr: Made the applet look more like Windows 10

smezzy:
Made the border to be left-only, and added an option for left-margin
it looks better if it's on the right side, but I'm sure that's where most people who come from windows are going to put it anyway


FilipeTorresBR:
I agree, you can merge @brownsr

PR #4269, opened 2022-02-21
added JDatPNW/CinnamonSleetimerApplet to the repo

JDatPNW:
This applet displays the time left until a scheduled shutdown will be performed. It shows the time until then in hours:minutes format and warns the user 15 minutes ahead of time as well. It also allows user to start shutdown or sleeptimer's through the popup menu of the applet.
Based on NIck Durante's tutorial.


JDatPNW:
I initially added the applet as a submodule instead of it being a directory. That has now been updated. Sorry about that.

PR #4273, opened 2022-03-04
radio@driglu4it: Add Croatian translation

muzena:
@jonath92 
From which source Radio applet obtains the radio stations playlist?

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/po/hr.po
+++ b/radio@driglu4it/files/radio@driglu4it/po/hr.po
@@ -1,0 +1,322 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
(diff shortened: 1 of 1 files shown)


jonath92:
From radio-browser.info with this Script:
https://gist.github.com/jonath92/0fd6419a2cfecbbe2f6b6d790dc84298

muzena:
@jonath92 
I tried to download the json playlist directly by category Country (Croatia), but this playlist doesn't work.
> curl -0 --http2 -o ./allStations.json http://all.api.radio-browser.info/json/stations/search?countrycode=HR

It would be good to implement if possible in applet options for downloading playlist by category (Country, Language, etc.) instead all playlist.

jonath92:
@muzena I think you need to use the script I mentioned and change the URL in the Script. The Script does not only download the file but also removes all attributes except the title and the URL. This is necessary because currently the complete file is loaded when opening the settings and this process gets really slow when the attributes are not filtered. That's also one of reason why is not that easy to implement filtering Features in the ui. It is of course possible but it doesn't have a high priority for me but pull requests are welcome.

PR #4282, opened 2022-03-23
Updating italian transaltion

lorypaduli:


brownsr:
the it.po under the cinnamenu files directory looks to be in the right place.  However the other two do not.  Can you check and correct, please ?

fredcw:
@lorypaduli Your updated it.po file seems to be based on a very old version of cinnamenu.pot. If you're using Poedit program, you'll first need to update the it.po file from the cinnamenu.pot file that's in the same directory. In poedit, open it.po and then go to Translation->Update from POT file... This will add all the new strings that need translating. Thanks.

brownsr:
Closing for now.  Free free to reopen or to resubmit a new PR when you've addressed @fredcw 's comments.

PR #4311, opened 2022-04-17
pomodoro@gregfreeman.org: add time left in tooltip

goctav:
This change adds the time left (noticeable as the `(minutes:seconds)` part) at the
end of the tooltip message, thus making it easy for
anyone who does not have the pomodoro timer enabled
in the applet panel to see the time left for the current cycle.
The tooltip message will update as often as the ticker changes.

The code changes aim to preserve the current applet behavior when displaying a tooltip. All the messages that were already present in code were reused, so translations will not require an update. 

How will the new message look like:
- during a pomodori run: pomodori run
- during a short break: short break tooltip
- after a short break, when a pause was requested: pause after short break
- during a long break: long break tooltip


goctav:
@gfreeau I would appreciate if you could have a look :)

gfreeau:
@goctav Looks good to me, thanks for your work.

PR #4314, opened 2022-04-19
Add workflow to generate releases.json

Gr3q:
## Description
These scripts will create/update the `releases.json` inside `.releases` folder on every commit to master.

It also appends a `-{revision}` to a released version if the version number was not changed to reliably track versions.

## Reasoning

TODO: write reasoning why is this useful

Part of the output:
```
{
    "AutostartPrograms@spacy01": {
        "id": "AutostartPrograms@spacy01",
        "path": "AutostartPrograms@spacy01/files/AutostartPrograms@spacy01",
        "versions": [
            {
                "version": "0.0.0-1",
                "commit": "ba0627d3",
                "message": "Initial commit",
                "date": "2016-12-31T14:25:05+00:00",
                "applet_checksum": "740473f8cf878c44481cf4f1fe1d3caed7f6bcc8",
                "url": "https://github.com/linuxmint/cinnamon-spices-applets/archive/ba0627d3.zip"
            },
            {
                "version": "0.0.0-2",
                "commit": "fd105009",
                "message": "Create bg.po",
                "date": "2017-04-24T20:11:06+03:00",
                "applet_checksum": "afb893fc6691e479d041b9ccf73c3c703cd4f410",
                "url": "https://github.com/linuxmint/cinnamon-spices-applets/archive/fd105009.zip"
            },
```


.github/workflows/releases.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      # Runs a set of commands using the runners shell
+      - name: Generate Translation Status Tables
+        run: |
+          sudo apt-get update -y >/dev/null 2>&1
+          sudo apt-get install -y python3 gettext >/dev/null 2>&1
+          cd .applet-versions
+          python3 release_generator.py
+          git config user.name "NikoKrause"
+          git config user.email "8415124+NikoKrause@users.noreply.github.com"
```
rcalixte, line 34: Should this have one user's credentials hard-coded?

Gr3q: It's been a long time, but I think I copied it from the workflow to generate the translation tables.

I'm not sure what I should be using, if it were on me I would use a generic CI account

rcalixte: I'm going to take a shot at adding some workflows in a few weeks that will work on a similar basis. Let's see what we can come up with that is a cleaner design.

PR #4317, opened 2022-04-24
ptt@thbemme: Push-to-talk applet and microfone indicator, Initial release

thbemme:
Initial release of ptt@thbemme


fredcw:
F11 key used for PTT is already used for fullscreen toggle and so doesn't work.

fredcw:
Also, the icons are a bit confusing. If you turn the mic off by clicking on the applet, an icon with a mic and a red dot appears, indicating that the mic is off but if you turn the mic off using the sound settings dialog, the applet icon is a mic with a red line through it instead (the one with the red line through it is much better IMO). Also, if you switch the mic off and use the push to talk feature, afterwards you can't turn the mic back on using the sound settings dialog as it keeps resetting back to off.

thbemme:
> Also, the icons are a bit confusing. If you turn the mic off by clicking on the applet, an icon with a mic and a red dot appears, indicating that the mic is off but if you turn the mic off using the sound settings dialog, the applet icon is a mic with a red line through it instead (the one with the red line through it is much better IMO). Also, if you switch the mic off and use the push to talk feature, afterwards you can't turn the mic back on using the sound settings dialog as it keeps resetting back to off.

The Applet has for stages
PTT mode off:
1. Mic on -> normal mic icon
2. Mic off -> mic icon with red line
PTT mode on:
3. Mic off, waiting for PTT to be activated -> mic icon with activity dot
4. Mic on, PTT active -> mic icon with green sound wave

And during PTT mode on but not active, the microphone is of course disabled continuously - even if you manually activate the mic - thats the expected behavior of a push to talk mode.

> F11 key used for PTT is already used for fullscreen toggle and so doesn't work.
The PTT key will temporarily assigned during PTT mode being active - so there should be not issue with double assignments, the F11 fuction will be restored once PTT mode is off. To be safe I assigned default to F9 (On my german layout I use the ^ key right below ESC).

fredcw:
Ah yes, I see now. For some reason I assumed that clicking on the applet would toggle the mic on/off but it instead turns the PTT mode on/off. And when PTT mode is on, F11 does work. I guess I was being a bit slow, sorry.

thbemme:
Is there anything missing or wrong with this PR or any other reason why this won't be merged?

PR #4321, opened 2022-05-02
[Cinnamenu@json] Add Persian translation

mskf1383:
Please merge it.
Thanks!


mskf1383:
I change lang name from `fa_IR` to `fa`, because there is no other persian translation available for this applet.

Issue #4331, opened 2022-05-09
color-picker@fmete - Stuck when trying to pick colors 


Siphalor:
```
 * Applet version/Build date: 2021-12-11 11:22:30 (last-edited: 1639218150)
 * Cinnamon version: 5.2.6
 * Distribution: Linux Mint 20.3 Cinnamon
 * Graphics hardware *and* driver used: Intel Corporation UHD Graphics 620 [8086:5917] with Mesa Intel(R) UHD Graphics 620 (KBL GT2)
 * Snippet or attachment of ~/.cinnamon/glass.log: N/A
 * Snippet or attachment of ~/.xsession-errors: No logs related to this applet
 ```

**Notify author of applet:** @icarter09 

**Issue**

When triggering the color picking functionality, through either a click or the keybinding, the cursor will transform into a crosshair, but neither clicking nor pressing escape will confirm or exit the process.

The cursor is stuck like this until Cinnamon is restarted.

All clicks over applications are captured by the applet and not handed to the application.
Clicks on panels work like normal, this makes navigation and restarting Cinnamon easily possible.

**Steps to reproduce**

1. Click on the icon / Use your keybinding
2. Try to click anywhere on the screen to pick the color (nothing happens)
3. Try to press escape to leave the picking mode (nothing happens)

**Expected behaviour**

Clicking anywhere on the screen should pick the color and (per the config) open a popup.

Alternatively, pressing escape should leave the picking mode and should return the cursor back to normal.

**Other information**

<details><summary>Configuration</summary>

```json
{
    "head": {
        "type": "header",
        "description": "Color Pick Settings"
    },
    "combo-selection": {
        "type": "combobox",
        "default": "1",
        "description": "Color code format",
        "options": {
            "HEX-Color: #xxxxxx": "1",
            "HEX-Color: #XXXXXX": "2",
            "RGB-Color: 250,250,250": "3",
            "RGB-Color: (250,250,250)": "4"
        },
        "tooltip": "Select the format of the color code, which will be copied to the clipboard by the color picker.",
        "value": "1"
    },
    "div1": {
        "type": "separator"
    },
    "head1": {
        "type": "header",
        "description": "Applet Settings"
    },
    "pick-notification": {
        "type": "checkbox",
        "default": "true",
        "description": "Show picked color code in a popup notification",
        "tooltip": "Show the picked color code, which has been copied to the clipboard, in a popup notification.",
        "value": "true"
    },
    "icon-name": {
[...]

PR #4350, opened 2022-05-23
Change calls to python to /usr/bin/python

Siphalor:
As discussed in #4331, it probably makes more sense to call to `/usr/bin/python` or `/usr/bin/python3` directly, instead of letting the `PATH` decide the python binary to use.

The main argument is that applets usually recommend/require certain apt `python-something` packages to be installed, which always target the python system installation at `/usr/bin/python`, while the user may choose to prefer another installation for most non-system use cases.

*I have only tested the color-picker applet, as this change should be pretty straight-forward.*
Someone should still look over this and check if I accidentally broke anything.

As mentioned in #4331, I would completely understand if this is deemed to be only a minor edge case.

---

Not sure if I should mention every applet author that is concerned :thinking: 

---

It should probably be considered to look for other binaries that are invoked similarly in this repository.
I chose to focus on python because of simplicity and because the `color-picker` was my main concern.

### DIFF
--- a/app-launcher@mchilli/files/app-launcher@mchilli/applet.js
+++ b/app-launcher@mchilli/files/app-launcher@mchilli/applet.js
@@ -788,7 +788,7 @@ class MyPopupSubMenuItem extends PopupMenu.PopupSubMenuMenuItem {
     _onButtonDelete() {
         this.applet.menu.close();
         Util.spawn_async(
-            ['python3', `${this.applet.metadata.path}/dialogs.py`, 'confirm', this.icon, this.name],
+            ['/usr/bin/python3', `${this.applet.metadata.path}/dialogs.py`, 'confirm', this.icon, this.name],
             (response) => {
                 response = JSON.parse(response);
                 if (response === Gtk.ResponseType.YES) {
@@ -803,7 +803,7 @@ class MyPopupSubMenuItem extends PopupMenu.PopupSubMenuMenuItem {
         this.applet.menu.close();
         Util.spawn_async(
             [
-                'python3',
+                '/usr/bin/python3',
                 `${this.applet.metadata.path}/dialogs.py`,
                 'edit',
                 JSON.stringify(this.applet.listGroups),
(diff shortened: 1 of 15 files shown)


fredcw:
I'll ping them just in case. @mchilli @joka42 @icarter09 @jake1164 @lauritsriple @gfreeau @jonath92 @jerrywham @cardsurf @Gr3q

Gr3q (approved):
> Just an idea: maybe it would be better to declare a global variable containing the python path

That just seems to me the same as using `python3` from a technical perspective, unless only cinnamon can change it. In the end even if something like that is implemented it's not going to fix any of the previous versions. 

Anyway, I'm ok with this change, although I'm sure [it will break someone's setup](https://imgs.xkcd.com/comics/workflow.png)

Drugwash2:
> That just seems to me the same as using `python3` from a technical perspective

From a certain point of view it's true, but as it is now any change of path sometimes requires multiple changes in multiple files; having only one declaration would just simplify any further changes.
Not sure what to say regarding previous versions.
As for setups and workflows... well, the current situation already broke someone's workflow as seen in #4331, reason why this change has been suggested.

mchilli:
Seems ok for me, anyway the path `/usr/bin/python` doesn't seems to exist anymore since Mint 20, but 
 I think that only applies for `softyubikey@yubiserver.include.gr`, which seems not to be under active development.

Siphalor:
> Seems ok for me, anyway the path `/usr/bin/python` doesn't seems to exist anymore since Mint 20, but I think that only applies for `softyubikey@yubiserver.include.gr`, which seems not to be under active development.

Ah, nice spot.
Not sure if I should change that to `/usr/bin/python2`, since that might actually result in a change of behavior then, since there is `python-is-python2` and `python-is-python3` nowadays ^^
They probably need to address that anyway if it's still under development.

Siphalor:
Since this has pr has been lying around for a few months now, I'd like to give this a little bump and ask if there's anything that still needs to be done to get this merged.

Sorry for notifying everyone in this thread, but I want to prevent this going completely stale :)

brownsr:
I was never convinced, so I just let it run to gather comments.  Still in that position.

Regards

Simon

claudiux:
In any case, this PR is too old to be merged and concerns too many different applets. Let's leave it to their respective authors to make the changes they deem necessary. Ping @mchilli @joka42 @icarter09 @jake1164 @lauritsriple @gfreeau @jonath92 @jerrywham @cardsurf @Gr3q.

PR #4355, opened 2022-06-01
[Cinnamenu@json] Add some translations.

fredcw:
Add translations using google translate.


cdrw-nutyx:
Updating the Russian translation
[ru.po.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/8839637/ru.po.zip)

KAMI911:
Is the quality of Google Translate good enough for those languages? In case of Hungarian language the GT quality is not good at all. Probably deepl.

cdrw-nutyx:
> 

The https://www.deepl.com/translator translator has a much better translation quality than google

fredcw:
@cdrw-nutyx Thanks for the translation, I'll include it in the next update. Yes, GT is obviously not going to be perfect but I guess that some translation is better than none and I only used it for previously untranslated strings. Also, I was thinking that maybe if someone sees a bad translation, it may encourage them to fix it :p. Thanks for the deepl tip, last time I looked at that site it only had a few languages.

fredcw:
@KAMI911 Actually, hungarian was one of the few languages I didn't update as someone translated it recently. Yes, GT is obviously not going to be perfect but I guess that some translation is better than none and I only used it for previously untranslated strings. Also, I was thinking that maybe if someone sees a bad translation, it may encourage them to fix it :p.

cdrw-nutyx:
> @cdrw-nutyx Thanks for the translation, I'll include it in the next update. Thanks for the deepl tip, last time I looked at that site it only had a few languages.

I'm sorry, you made a mistake when you updated the translation.
It should be like this
msgid "Top"
msgstr "Вверху"

msgid "Bottom"
msgstr "Внизу"

After the update it looks like this

msgid "Top"
msgstr "внизу"

msgid "Bottom"
msgstr "внизу"

You shouldn't have merged the translations. There are a lot of mistakes. Sorry, I don't know how to edit translations.

fredcw:
@cdrw-nutyx Actually, the above translation of "Top" is nearly 2 years old and is probably an editing mistake rather than a mistranslation and I've only used google to translate strings that would otherwise have no translation. Although GT is not perfect, I think that even when it's wrong it's usually close enough that people can guess what it means.

You say that you don't know how to edit translations but you seem to have linked above to an updated ru.po? Do you mean that you don't know how to merge your changes into the repository? If so, that's not a problem, you can just link to the updated .po file here like you have above or you can email me a copy, either way is fine. Thanks.

PR #4377, opened 2022-06-28
[Cinnamenu@json] Bugfix

fredcw:
Fix regex doesn't work in Mint 19.3 bug - prevents applet from starting.

@brownsr @mtwebster @clefebvre @Cobinja This is an urgent bugfix, can someone merge ASAP, thanks.


fredcw:
@collinss @xenopeek @icarter09 @NikoKrause

PR #4392, opened 2022-07-08
[CassiaPanelLaunchers@klangman] Initial commit

klangman:
Signed-off-by: Kevin Langman <klangman@gmail.com>


Drugwash2:
Oh this is bad. This is beyond bad! Can't even begin to explain how bad it is!
ALL launchers are gone. ALL of them! Disabling the built-in Panel Launchers was THE disaster because whoever created this POS called Cinnamon didn't think things through. Or didn't think, period. There is no backup for the individual applet settings unless the user would expect everything to fail and would search realy-really hard for the obscure backup method of "three-horizontal-lines-in-an-icon-like-button-that-looks-like-a-hamburger". I didn't. Stupid me, I guess. So when enabling back the built-in Panel Launchers the panel got only the three useless launchers: Firefox, Terminal and Nemo. But I had a lot of them, many custom-made using commands that I can't remember anymore.

Moreover: I had selected another panel to add your applet, not the main one, precisely in an attempt to preserve the existing launchers. When adding back the built-in one it added it to the same secondary panel. Then I added it to the main panel too. Then I wanted to remove the one from the secondary panel. Guess what? It removed ALL instances of it!!! How stupid is that?!?

I need to find another OS, this one is unusable.

Oh and you might want to warn all possible users a.s.a.p. about this issue so they would not lose all their customized launchers. And if possible find a way to copy the current items from the built-in applet to your applet for a seamless transition.

Drugwash2:
**P.S.** Your applet doesn't even launch in Mint **19.2** (Cinnamon **4.2.4**).

```
error t=2022-07-08T15:50:15Z 
[CassiaPanelLaunchers@klangman]: (intermediate value)._launcher.launchersBox is null
[CassiaPanelLaunchers@klangman]: Failed to evaluate 'main' function on applet: CassiaPanelLaunchers@klangman/36

```

klangman:
> There is no backup for the individual applet settings

That's true. There is no backup or remembering what launchers were added to the applet after you remove the applet. When editing or creating a launcher the default panel launcher creates custom desktop files in a special location (which you could find I suspect). When editing, my applet just creates a copy of the desktop files in the standard ".local/share/applications/" which has it's advantages and disadvantages.

Drugwash2:
In a recent, untweaked setup it's relatively easy to find things but in an old and heavily tweaked setup everything is a maze. Incidentally I had found the custom launchers a few days ago while looking for something else, otherwise it would've taken me a looong time to find them.

Even so, their names are machine-friendly not human-friendly (`cinnamon-custom-launcher-1.desktop` ... `cinnamon-custom-launcher-N.desktop`) so, unless one uses a real file manager like Double Commander which knows how to display the associated icons finding which is what would be an additional waste of time.

On top of that, the **original order** of the icons/launchers is lost, the user would have to rely on their memory - and mine is so very bad I sometimes forget the names of my favorite applications that I've used for maybe decades. Luckily I take fullscreen screenshots from time to time when working on some applet, so I get the original sorting order from there. Anyway you look at it, it still takes some time to get them all back. Time that could've been used more fruitfully if only they knew what they were doing with this Cinnamon - it has too many flaws and shortcomings, and the more I [attempt to] use it the more of them I discover.
<hr/>
All this aside, the idea is for your applet to get the current value (a list of *.desktop filenames) from the built-in Panel Launchers applet's settings and use it as is - or process it first if necessary - so that the user wouldn't have to recreate the old appearance. No idea if it does that or not since I couldn't even start it on my 19.2 setup.


And more than that, it would be really helpful for the user if your applet could create backups for both your applet's settings and built-in's, so that going back to the built-in applet for whatever reason would be much easier. A popup notification or any other way of telling the user where the backup is and how to restore the settings would also be of great help.

Now, besides the above suggestions I'd be interested to know if you could work on making this applet multiline-aware, that is display launchers on as many lines as possible according to panel height and icon size for its current panel zone, thus maximizing free space for the window list. Granted the same thing should be done for the applets/system-tray zone as well. I already modified two applets to take advantage of a higher panel size but the entire area should become multiline-aware.

Oh and maybe you could please look into the non-launching issue on 4.x Cinnamon.

klangman:
Noticed some issues on low resolution displays, but no time to debug it at the moment. I'll close for now and submit it again after I find time to address the issues.

PR #4394, opened 2022-07-11
[weather@mockturtl] Hebrew update for v3.2.6

avma:
Hebrew he.po update


brownsr:
sorry @avma something isn't right here with the weather translation file.  The other two are not showing issues so should be OK.  I'll close this for now while you look at what you can do to resolve it.  In general if you can submit the translations for things one at a time it will avoid this sort of thing where a fault in one file holds up changes in others.

Regards

Simon

PR #4397, opened 2022-07-12
[darkMode@linuxedo.com] Fix configuration dialog not opening

### LINKED ISSUE #4382: Can't configure the applet
```
 * Applet version/Build date Updated on 2022-06-07 17:26:49
 * Cinnamon version 5.2.7
 * Distribution - Linux Mint 20.3 Cinnamon
 * Graphics hardware *and* driver used

 ```

**Notify author of applet**
@slgobinath 
**Issue**
Dark mode applet is not letting me configure it. Whenever I'm clicking the configure option, it's not showing the usual window. 

**Steps to reproduce**
Right click the applet, click configure.

**Expected behaviour**

**Other information**

fredcw:
Fix configuration dialog crash in pre cinnamon 5.4 versions due to no icon theme being found when ~/.local/share/icons folder doesn't exist and thus no icon theme options were available for the combobox causing xlet-settings.py to crash.

Fixes #4382

@slgobinath

### DIFF
--- a/darkMode@linuxedo.com/README.md
+++ b/darkMode@linuxedo.com/README.md
@@ -8,6 +8,8 @@ For more details and feature requests: [Dark Mode in Cinnamon Desktop Environmen
 License: [GPL-v3](https://github.com/linuxedo/cinnamon-dark-mode-applet/blob/main/LICENSE)
 
 ## Changelog
+* Jul 12, 2022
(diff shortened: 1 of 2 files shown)


slgobinath:
Thanks. Didn't notice this bug I introduced in the recent commit.
Makes sense to me. Thanks @fredcw  for taking your time and effort to fix it.

PR #4407, opened 2022-08-01
qredshift@quintao - Fix applet on 5.4

### LINKED ISSUE #4406: qredshift@quintao sliders not working on arch derivates
```
 * Applet version (1.7.1)
 * Cinnamon version (5.48)
 * Distribution - (Arch)
 * Graphics hardware *and* driver used: intel uhd
 ```

**Notify author of applet**
@raphaelquintao 

**Issue**
The applet works fine expect the sliders have no function when using an arch derivate. Tested with several LTS kernels. No issues when using mint cinnamon.

**Steps to reproduce**
Use something else then mint cinnamon, e.g., manjaro, install QRedshift, put it in the panel  try to use the panels popovers sliders.

**Expected behaviour**
Sliders work

---
edit: I didn't realized that mint 20.3 uses an earlier version of cinnamon then `5.4`. I assumed that mint beeing the primary devoper of the cinnamon desktop, it's handled like with KDE Neon and Plasma, that it always has the latest version of the DE. My mistake. This issue probably isn't related to the distro but to the cinnamon version and willl probably be experienced when using mint 21.

### PR DESCRIPTION
**Pull request by Gr3q**
@raphaelquintao This is a complete fix, using device from event to grab the pointer instead of removed function.

* Fixes #4406

This PR is now ready

### DIFF
--- a/qredshift@quintao/files/qredshift@quintao/js/QUtils.js
+++ b/qredshift@quintao/files/qredshift@quintao/js/QUtils.js
@@ -486,13 +486,18 @@ class QPopupSlider extends QPopupItem {
         this.emit('drag-begin');
         this._dragging = true;
 
-
-        // FIXME: we should only grab the specific device that originated
-        // the event, but for some weird reason events are still delivered
-        // outside the slider if using clutter_grab_pointer_for_device
-        Clutter.grab_pointer(this.slider);
-        this._signals.connect(this.slider, 'button-release-event', this._endDragging.bind(this));
-        this._signals.connect(this.slider, 'motion-event', this._motionEvent.bind(this));
+        if (imports.misc.config.PACKAGE_VERSION < "5.4") {
+            // FIXME: we should only grab the specific device that originated
+            // the event, but for some weird reason events are still delivered
+            // outside the slider if using clutter_grab_pointer_for_device
+            Clutter.grab_pointer(this.slider);
+            this._signals.connect(this.slider, 'button-release-event', this._endDragging.bind(this));
+            this._signals.connect(this.slider, 'motion-event', this._motionEvent.bind(this));
+        }
+        else {
+            this._signals.connect(this.actor, 'button-release-event', this._endDragging.bind(this));
+            this._signals.connect(this.actor, 'motion-event', this._motionEvent.bind(this));
+        }
 
         let absX, absY;
         [absX, absY] = event.get_coords();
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Drugwash2** [comment]
This is not quite right. If you look at `usr/share/cinnamon/js/ui/popupMenu.js` you'll see that now the **grab/ungrab** is performed through the `event` object. Done correctly there will be no side-effects for either Cinnamon 5.4 or the earlier versions.

For my **Applet PYE** version of the customized popup menu slider I opted for  try/catch pairs so that the code itself would decide which methods to call based on their existance and not on mere abstract version numbers. This is one of the mistakes many developers have made in years: assuming a version number would tell everything thy needed to know, forgetting that customized libraries/whatever could exist. If right now someone would ship a distro with a, say, v2.5.4 Cinnamon that would actually be a newer, improved version of 5.4 a lot of "v5.4-updated" software would go belly up.

So, back to the issue at hand, my suggestion is to use

```
try { event.get_device().grab(this._slider); }
catch(e) { Clutter.grab_pointer(this._slider); }
```

in the `_startDragging(actor, event)` method, and

```
try { event.get_device().ungrab(); }
catch(e) { Clutter.ungrab_pointer(); }
```

in the `_endDragging(actor, event)` method. _(note it now needs the `actor` and `event` parameters in the definition)_
This would make all Cinnamon versions happy regardless of their version.

Gr3q:
> This is one of the mistakes many developers have made in years: assuming a version number would tell everything they needed to know, 

Usually that's the case, but because Cinnamon officially does not support ( at least Cinnamon itself would break in various ways) other versions of Clutter for a given Cinnamon version, this should not be a problem.

Also the version is always semantic in that file, if some distro changes it than that is a problem.

Not really my first choice in general to rely on it, but still.

As for the fix, I will do that, again I'm happy if I have time to make them not crash

Drugwash2:
<blockquote>Cinnamon officially does not support [...] other versions of Clutter for a given Cinnamon version [...]
Also the version is always semantic in that file, if some distro changes it than that is a problem.</blockquote>

This seems to be the main reason for the utter lack of backward compatibility in Linux. For me coming from 20+ years of Windows this is a big dissapointment and drawback.

<blockquote>As for the fix, I will do that, again I'm happy if I have time to make them not crash</blockquote>

It should be safe. Also you may want to check and exclude `Clutter.ScrollDirection.SMOOTH` from all scrollable operations where scroll direction is checked for, in order to avoid unnecessary processing since this seems to be sent along with and before any other scrolling directions.

Gr3q:
Ok, tested it on 3.8, it doesn't need any checks, `event.get_device().grab` and `event.get_device().ungrab` is the recommended way to do this and they were supported on all previous versions (at least up to 3.8 what this applet officially supports).

Drugwash2:
> `event.get_device().grab` and `event.get_device().ungrab` is the recommended way to do this and they were supported on all previous versions

That's good news, less checks means faster code.
Still kinda wonder when and why was Clutter brought into the mix. Maybe copied from some old code or something? Oh well, nevermind.

Gr3q:
Cinnamon depends on `Clutter`, `St` and other GTK libraries since the time it was forked from Gnome. Can be used from `imports.gi.*` namespace.

`Clutter` as a dependency was not updated for a long time (stuck on 2.6.x) until it was completely removed by the GTK team (I suppose). So the Cinnamon team had to update a more recent version, and they chose version 7. 

Essentially jumping 5 versions, so the implementation->deprecation->removal flow of items in the library got squashed into implementation->removal, causing issues.

I know all of this because I generate the Typescript declarations of all of these libraries for myself, like [this is before 5.4](https://github.com/Gr3q/types-cjs/tree/5.2.1), and [this is after](https://github.com/Gr3q/types-cjs/tree/5.4.2)

Drugwash2:
Oh I kinda know all that, more or less. Was just wondering only about `Clutter.grab/ungrab_pointer()`, how they got there in the context of those methods when the `event` object would/should have been the most logical and immediate choice.

PR #4433, opened 2022-08-18
[Cinnamenu@json] Fix Soup version

cheeeeer:


fredcw:
@cheeeeer Thanks, the first part of your fix is perfect `imports.gi.versions.Soup = "2.4"`, however the comparison line `if (Soup.MAJOR_VERSION === 2.4) {` will prevent suggestions and wikipedia search from working as `Soup.MAJOR_VERSION` only returns 2, not 2.4

I submitted another PR to fix this so this PR will be unnecessary. Thanks for your help.

Fantu:
`imports.gi.versions.Soup = "2"` is wrong, probably only change in `imports.gi.versions.Soup = '2.4';` is needed

brownsr:
Thanks for this.  As far as I can see the author @fredcw has built on your suggestion and put a change through, so this doesn't appear necessary now.

Regards

Simon

PR #4440, opened 2022-08-24
[Cinnamenu@json] Fix key navigation for mint 21

### LINKED ISSUE #4437: Cinnamenu@json no longer accepts numpad enter as enter after updating to mint 21
```
 * Version 5.0.1 (2022-08-20 07:14:57)
 * Cinnamon 5.4.10
 * Mint 21
 * rx580, mesa 22
 * file doesn't exist
* Snippet or attachment of https://github.com/linuxmint/cinnamon-spices-applets/files/9394058/xsession-errors.txt

 ```

**Notify author of applet**
@fredcw

**Issue**
Cinnamenu should accept both numpad enter and normal enter to open an application

**Steps to reproduce**
search for something then use numpad enter, it will do nothing

**Expected behaviour**
it should open the searched-for app

**Other information**
issue appeared after upgrading to mint 21

### LINKED ISSUE #4438: Cinnamenu@json keyboard navigation problem
* Version 5.0.1 (2022-08-20 07:14:57)
 * Cinnamon 5.4.10
 * Mint 21
 ```

Notify author of applet
@fredcw

Issue
in grid view , is not possible to retourn to category view from app view

**Steps to reproduce**
open menu, use the arrows to choose a category, then use the right arrow to go to the applications view. now it is impossible to retour with keyboard to category view

**Expected behaviour**

retour to category view from app view with keyboard

**Other information**

Work as expect in list mode.

fredcw:
and add code in case libsoup3 is loaded.

Fixes #4437
Fixes #4438

### DIFF
--- a/Cinnamenu@json/files/Cinnamenu@json/4.0/applet.js
+++ b/Cinnamenu@json/files/Cinnamenu@json/4.0/applet.js
@@ -810,8 +810,9 @@ class CinnamenuApplet extends TextIconApplet {
         };
 
         switch (true) {
-        case (symbol === Clutter.KP_Enter || symbol === Clutter.KEY_Return) && ctrlKey:
-        case symbol === Clutter.KEY_Menu && noModifier:
+        case (symbol === Clutter.KEY_KP_Enter || symbol === Clutter.KP_Enter ||
+                                            symbol === Clutter.KEY_Return) && ctrlKey:
+        case (symbol === Clutter.KEY_Menu) && noModifiers:
             if (this.contextMenu.isOpen) {
                 this.contextMenu.close();
             } else if (focusedAppItemExists) {
@@ -822,7 +823,8 @@ class CinnamenuApplet extends TextIconApplet {
                 categoryButtons[focusedCategoryIndex].openContextMenu();
             }
             return Clutter.EVENT_STOP;
-        case (symbol === Clutter.KP_Enter || symbol === Clutter.KEY_Return) && noModifiers:
+        case (symbol === Clutter.KP_Enter || symbol === Clutter.KEY_KP_Enter ||
+                                        symbol === Clutter.KEY_Return) && noModifiers:
             if (focusedContextMenuItemExists) {
                 contextMenuChildren[focusedContextMenuItemIndex].activate();
             } else if (focusedAppItemExists) {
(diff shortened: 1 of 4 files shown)


Drugwash2:
Just a curiosity. I saw you disabled the HTTP Proxy in the sessions. Doesn't the new way work?
`_httpSession.add_feature(new Gio.ProxyResolver.get_default());`
Of course it would require Gio as an additional import if it's not already imported but... Is there any catch here?
Admittedly I'm a novice with all this so apologies if it sounds too noob.

fredcw:
@Drugwash2 Tbh, I've no idea what that line even does. I just copied it from https://github.com/linuxmint/cinnamon/blob/master/docs/search-providers-examples/wikipedia%40cinnamon.org/search_provider.js

I just commented out the line and everything seems to work exactly the same so I guess it isn't needed.

Drugwash2:
Frankyl I have no idea either but I've seen it in many xlet scripts and I can only assume it may be taking into account any proxy that may have been set for the currently active Internet connection. If that is true then by eliminating that call we may inadvertently disobey user's choice, and the connection may not even work - properly or at all.
Hopefully someone more knowledgeable would enlighten us both on this matter.

fredcw:
@Drugwash2 It looks like you're pretty much spot on. It seems in libsoup 2.4 you have to set a proxy resolver when using Soup.SessionAsync().

From [here](manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupProxyResolverDefault.php)

>[SoupProxyResolverDefault](https://www.manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupProxyResolverDefault.php) is a SoupProxyURIResolver implementation that uses the default gio GProxyResolver to resolve proxies.

>In libsoup 2.44 and later, you can set the session's [“proxy-resolver”](https://www.manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupSession.php#SoupSession--proxy-resolver) property to the resolver returned by g_proxy_resolver_get_default() to get the same effect. Note that for "plain" [SoupSessions](https://www.manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupSession.php#SoupSession-struct) (ie, not [SoupSessionAsync](https://www.manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupSessionAsync.php) or [SoupSessionSync](https://www.manpagez.com/html/libsoup-2.4/libsoup-2.4-2.62.0/SoupSessionSync.php)), this is done for you automatically.

And in libsoup 3, it's apparently added by default. from [here](https://gitlab.gnome.org/GNOME/libsoup/-/blob/master/docs/reference/client-advanced.md)

>By default libsoup tries to respect the default proxy (as best as
[[func@Gio.ProxyResolver.get_default](mailto:func@Gio.ProxyResolver.get_default)] knows), however you can set a custom one or
disable it outright using the [property@Session:proxy-resolver] property.

I've added the lines back in. Thanks.

PR #4457, opened 2022-09-04
[better-backgrounds@simonmicro] Add support for libsoup3

fredcw:
Add support for libsoup3. For more info see https://github.com/linuxmint/cinnamon/commit/76224fe409d074f8a44c70e4fd5e1289f92800b9

@simonmicro

### DIFF
--- a/better-backgrounds@simonmicro/files/better-backgrounds@simonmicro/applet.js
+++ b/better-backgrounds@simonmicro/files/better-backgrounds@simonmicro/applet.js
@@ -1,6 +1,7 @@
 const Applet = imports.ui.applet;
 const AppletManager = imports.ui.appletManager;
 const Soup = imports.gi.Soup;
+const ByteArray = imports.byteArray;
 const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const Tweener = imports.ui.tweener;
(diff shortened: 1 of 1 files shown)


simonmicro:
Thanks for your work - I'll merge it also into my branch / upstream, if it works :grin:

DRR0B0T:
URI does't work check it in http://source.unsplash.com/ , i see it in your source code. Change your uri to the uri wich works pls.

fredcw:
Just to be clear, the issue mentioned by @DRR0B0T is not related to this PR as this PR makes no changes to any of the urls used by the applet. I've tested this PR with both soup 2.4 and 3.0 and it all seems to work fine.

Having said that, @DRR0B0T , the unsplash url you mentions seems to work for me in the applet as it is although it does seem quite slow to respond (about 4-5 seconds), I'm in the UK if that makes a difference? Perhaps open a new issue for this, as it's not related to this PR?

simonmicro:
@DRR0B0T This does not seem right. Could you share the URI you have tried and does not work for you? It will be logged on the console. If it fails for multiple people, it may seem worth investigating...

@fredcw Thanks for your work!

DRR0B0T:
@simonmicro This http://source.unsplash.com/.

DRR0B0T:
@fredcw At now it working normally, but downloading works with biggest timeout. Linux Mint Cinnamon 21

simonmicro:
@DRR0B0T Moved to here: https://github.com/simonmicro/better-backgrounds/issues/22

PR #4470, opened 2022-09-12
ddcci-multi-monitor@tim-we: Prevent the applet from crashing when invalid displays are present

Naushikha:
### Issue

The applet has issues when running on laptops where displays do not support DDC/CI,

image

The parsing code for the command 'ddcutil detect' incorrectly identifies 'i2c-7' as the bus for my external monitor (which should be 'i2c-4') because there is no guard condition that prevents execution when it meets an 'invalid display'. 

This results in the applet crashing,

image

### Fix

The proposed fix makes sure that the code does not continue execution and update values (currentDisplay.bus and currentDisplay.name) that have already been assigned. This effectively gets rid of 'invalid displays' with minimal changes to the codebase. Also, it preserves the existing display naming scheme (with an added fallbackName property).

**TL;DR - Applet can now be used with laptop displays and external monitors**

@tim-we

### DIFF
--- a/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
+++ b/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
@@ -224,23 +224,26 @@ async function getDisplays() {
             const index = parseInt(displayMR[1], 10);
 
             if (currentDisplay) {
+                if (currentDisplay.name === undefined) {
+                    currentDisplay.name = currentDisplay.fallbackName;
+                }
                 displays.push(currentDisplay);
             }
 
             currentDisplay = {
                 index,
-                name: displayMR[0],
+                fallbackName: displayMR[0],
             };
         } else {
             if (currentDisplay) {
                 const busMR = line.match(/^\s+I2C bus:\s+\/dev\/i2c-(\d+)$/);
 
-                if (busMR && busMR.length === 2) {
+                if (busMR && busMR.length === 2 && currentDisplay.bus === undefined) {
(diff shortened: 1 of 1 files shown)


Drugwash2:
@Naushikha : Just out of curiosity: what's your version of `ddcutil` and which OS was that?
I just tested here on a **Samsung R580** notebook under **Mint 19.2** and the `ddcutil` returned, among others, the option **Supports DDC**. Maybe testing/processing that info would be a valid alternative.

Screenshot from 2022-09-12 23-57-25

Naushikha:
Hi @Drugwash2 
I am running Linux Mint 20 with ddcutil v0.9.8 (HP ProBook 445 G7).
Here are some screenshots,
 
image

image

I am a bit confused though, your screenshot shows that the display clearly doesn't support DDC? (it's listed as an 'invalid display')

Let me know if you need any more info :D

Drugwash2:
That info is enough, thank you. No idea why mine shows the DDC suport clearly as false. Maybe an old option that has been stripped from never versions of `ddcutil`; that's why I asked for your version.

Dunno if it's important in the context but that display is not the original that came with the notebook; that one got smashed by accident and had to replace it with this one that was laying around for quite a while (it comes from an ASUS notebook).
Here's my neofetch report for comparison:
Screenshot from 2022-09-13 09-54-04

Drugwash2:
Haven't looked through the code but I believe one - or more, if needed - well crafted regular expression patterns could guard against any changes in the output format.

Dunno if I'll ever get to test the applet here in order to maybe find some solution. Using an external monitor in my setup creates certain issues that I'm not willing to deal with right now.

PR #4500, opened 2022-10-15
nordvpn-indicator@nickdurante : Fix connection status detection

ShikharArvind:
## Issue
The new nordvpn app on linux prints about its meshnet feature any time a command is called. The breaks the connection status detection as is not as structured as before.

## Fix
Instead of assuming a structured output for `nordvpn status` , we just a regex match for the word after `Status: ` and match it with Connected or Disconnected.


ShikharArvind:
Pinging @nickdurante just in case .

PR #4522, opened 2022-10-30
rswitcher@nixdev.com: create da.po

Alan01:


Alan01:
There was an error in the translation

PR #4553, opened 2022-11-06
Cinnamenu@json: improve german translation

nekomajin:


erwn16:
**This new translation contains some deteriorations**

402: The commas are wrong because "statt" is not used in conjunction with an infintive group.
The new lines 572, 575 are less precise than the old translations. In these cases word by word translations related to the original texts are not an improvement because the original texts are not precise too.
686: "Themes" in a German sentence = "DEnglisch".

brownsr:
will leave open for now after the adverse comments.  would you like to consider these ?

Regards

Simon

nekomajin:
I thought there would be annotations to my comments.  Because that doesn't seem to be the case, I'm closing this pull request.

PR #4558, opened 2022-11-13
Radio3.0 v1.0.0 20221113

claudiux:
New applet Radio3.0@claudiux.
Fully functional. Allows the user to create its own list of web radio stations from free databases, from directories such as ShoutCast or importing files usually used by music-readers. Also, the user can record a music or radio show (directly or programming it). This applet also can extract soundtracks from a YT video or playlist.


jonath92:
Hi claudiux, that looks pretty interesting. But I don't understand why it requires a new applet. Wouldn't it be more user friendly to collaborate on one applet?

jonath92:
Hi @claudiux,  
There are of course many good reasons to develop multiple applications with similar functionality, e.g. technical reasons or irreconcilable differences of opinion. To a certain extent, competition is even a good thing, because you can benefit from each other's ideas. In this case, however, I personally think that both the users and ourselves would have benefited more from a collaboration. Of course my implementation is not perfect and in fact more complex than I like, but still I personally think that working together to reduce complexity would add more value than creating two similar applications, but of course it's absolutely your choice. Contributions to the existing radio applet from your side would definitely be appreciated. 

But other than that, I find the name a bit misleading. From my point of view, Radio3.0 suggests that it is an continued development of the existing radio applet, which is not the case. Therefore, I would find another name, e.g. simply without 3.0 much better, but that is of course also your decision.

fredcw:
When radio++ is installed and then Radio3.0 is installed, radio++'s window goes from working normally to briefly appearing in the top right hand corner and then disappears each time you click on the applet. The other way around - installing radio3.0 and then installing radio++ causes cinnamon to crash. After cinnamon is restarted, both applets work normally.

jonath92:
@fredcw I can't reproduce this. Are you sure you have the newest version (2.2.1) installed (there was a fix for a similar issue)? If so, anything in the log?

fredcw:
@jonath92 You're right, I was using the previous version. Sorry.

claudiux:
Hi @brownsr 
Can you merge this branch with the master one, please?
Regards.
Claudiux

Drugwash2:
Tested this just for kicks earlier and there are a few things to ponder upon:

- at first start it required a few missing packages on this Mint 19.2 Cinnamon; that may be alright but it would be quicker and more professional if it checked for all packages at once, built a list of all the missing ones, and installed them in one pass instead of multiple separate passes (I had three missing packages and required three subsequent passes).

- after installation of packages there was still an error about a missing default value in the settings schema:
`error t=2022-11-17T23:49:43Z Settings key search-list-page-label is missing property default`
That one was easy for me to fix manually but regular users would have a hard time debugging it.

- finally after all was fixed and the applet started there was an error about lack of free space for recording. I hadn't even started to check the settings or anything, no recording was ever set up, so the error is at least misleading.

The real problem is I didn't see any place in the Settings panel where the user could set up the destination path for the recording. As it happens I have very little free space on the root/home partition, and if I ever wanted to record anything I would have to choose one of the mounted network drives as that's where free space exists. The applet should allow the user to specify the destination path for the recording and also always check if the path is valid/available before attempting to save (or load) anything in that path.

- one other thing would be that some of the pages in the Settings panel are way too tall for a regular/old laptop screen (1366x768). Maybe they need some rethinking, splitting - dunno.

That's about all I can say for now. Again, this was just a short test run out of pure curiosity. Hope the above helps improving this applet. Good luck! 👍

PR #4568, opened 2022-11-18
 [nordvpn-indicator@nickdurante] Make the update function asynchronous and fix regex exec null error

ShikharArvind:
### Issue(s)

1. Previously, the update function (`_get_status()` ) called `nordvpn status` from `_run_cmd() `which called `spawn_command_line_sync ()` (i.e. synchronously). This was called periodically for a given `update_interval`. This is not desireable as it blocks the main thread until `nordvpn status` returns something causing applet load time during boot over 20s in my usecase. 
2.  The connection status was detected using `regex.exec()`  which can return a null  if there are no matches. 

### Fixes

1. The `_run_cmd() `function is  rewritten with asynchronous Gio subprocess. It returns a `Promise` object. This is then used in `_get_status()` to update the applet  GUI. This does not block the main thread allowing the applet to update the status whenever nordvpn loads during boot and returns `nordvpn status`. 
2. The `regex.exec() `is now wrapped in try catch block to catch any null errors thrown. 

### Additional

- I'm also interested in maintaing this applet incase @nickdurante has moved on from this project. It is a fun and didactic project for me. 

Cheers!
Shikhar


brownsr:
ping @nickdurante 

@ShikharArvind Note that there are conflicts which need resolving.

Regards

Simon

nickdurante:
Hi @ShikharArvind thanks for your work.
I apologize for the delay, other things in life made me forget to maintain this applet. You can take over the project if you want.

Best,
Nick

ShikharArvind:
> ping @nickdurante
> 
> @ShikharArvind Note that there are conflicts which need resolving.
> 
> Regards
> 
> Simon

@brownsr The conflicts should be resolved now. 

Regards,
Shikhar

PR #4570, opened 2022-11-22
A new Finnish translation for the program temperature@fevimu

mahtiankka:
This program has not been translated into Finnish before. I'm not sure if it's enough to just add the po file to the folder.


mahtiankka:
New Finnish translations for applets

PR #4574, opened 2022-11-27
Cleaning up the BAMS Applet

rcalixte:
### 1.5.0
 * Switched from `upower` to the power status module provided by the kernel at /sys/class/power_supply/BAT0
   (`upower` updates on a two minute interval while the kernel module updates live and the applet supports smaller increments, increasing the overall accuracy of the applet.)
 * Added a setting for manual configuration of the system battery path
 * Consolidated changelogs into CHANGELOG.md
 * Formatted .json files to standard JSON schema
 * Cleaned up unneeded legacy files and organized into a consistent versioned structure
 * Code comments improved and other comments removed
 * Updated batterymonitor.pot and translation files
 * Renamed ua.po to uk.po (Ukrainian)
 * Fixed file modes
 * Cleaned up whitespace, semicolons, and formatting
 * Cinnamon versions up to 5.6

@pdcurtis as well.


pdcurtis:
@rcalixte Thank you for the work you have put in tidying up this applet in general and your work on the battery check script to reduce the number of calls to `upower`.

The existing battery check script has remained unchanged since the applet was originally written many years ago and I have never had any comments that it fails on any of the 'flavours' it has been deployed on. Your modification looks sensible and safe but I can not deploy it and test on anything other than a standard Mint system. 

I am happy to agree your PR on the basis that we could easily revert that battery check commit in the very unlikely event that any users do have problems.

rcalixte:
@pdcurtis Sorry for the confusion about the script changes! It's not that the script fails (I believe most distros ship/install `upower` by default), but the change in logic.

The `/sys/class/power_supply` path is [provided by the kernel](https://www.kernel.org/doc/Documentation/power/power_supply_class.txt) and it is updated live whereas `upower` only refreshes every 2 minutes. Given that the applet polls more frequently than two minutes by default, it either makes sense to update/restrict the polling interval in the applet or keep the polling in place but use a live source. I opted for the latter as I think this made for a better user experience. (Both strategies are also dependent on the kernel module being installed and active for the power information to even be available to poll so no changes there.)

Would you be more comfortable if I tested this wider? I've tested so far on two Debian-based systems and one RedHat-based. [Arch](https://wiki.archlinux.org/title/laptop#Hibernate_on_low_battery_level) and [Gentoo](https://wiki.gentoo.org/wiki/Sysfs) documentation also include that they use the same kernel path. (I actually don't run any Mint systems as of today but I believe this covers the family of distributions where Cinnamon is supported/installable.) [Another desklet](https://cinnamon-spices.linuxmint.com/desklets/view/31) I use also makes use of [the same path for inputs](https://github.com/linuxmint/cinnamon-spices-desklets/blob/master/battery%40schorschii/files/battery%40schorschii/desklet.js#L100-L109) as well. (This is how I originally noticed the two being out of sync.)

pdcurtis:
@rcalixte I started to look at the change log to see if it that also needed to be changed and realised that changes have already been made by @sphh and the script is no longer in use because of the risks of wear on SSDs due to the large numbers of writes. Have a look at #2659 where I made several comments. It looks as if the script file was left in case the change needed to reverted or by accident. 

The change made still uses upower but changing to use sysfs calls would be a more complex change than in the script file. I tend to be a "If it ain't broke don't fix" person but if you want to have a look please go ahead if not please proceed with the other 4 commits in the PR. You have put my mind at rest that you have been testing thoroughly so far and I have to admit I had a look at your profile and you are far more active than I am these days! Currently most of my efforts have been going into OpenEmbedded systems for Solar Power using Node-RED.

sphh:
@rcalixte: I am the author of PR #2659 which got rid of writing to temporary files, but I did not get rid of using `upower` at that time, mainly because I was not aware of `/sys/class/power_supply/BAT?/uevent`. Personally I like the approach of reading directly form `/sys/class/power_supply/` system! May I suggest to read directly from `/sys/class/power_supply/BAT?/uevent` from inside the applet without using `upower`? That would be great, if you do not have to start a script every 2 s …

rcalixte:
@sphh Thank you and agree! You've pointed out on change along with a few others I was hoping to have time to implement. Another one is versioning ([example here](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/workspace-grid%40hernejj/files/workspace-grid%40hernejj)) so that the applet can better support versioning. This would allow us to have more flexibility for pinning changes to maintain backwards compatibility while also being able to be forward-looking with new features (such as precisely what you encountered in your PR).

This was a larger change and I was hoping to keep things small but it seems like it might be more prudent to append with an updated directory structure and remove the need for scripts for the latest version. (I don't have an older client to test with and I would prefer not to break anything.)

@pdcurtis If this path makes sense to you, I'll take some time and apply the changes to the paths and re-implement the applet logic. I can also do that work in a new PR as well. I am fine with either.

rcalixte:
> Currently most of my efforts have been going into OpenEmbedded systems for Solar Power using Node-RED.

Also, very cool. 😎

sphh:
@rcalixte: The change from `upower` to `/sys/…` should quite easy:
1. Instead of calling `upower` with a call-back function at https://github.com/linuxmint/cinnamon-spices-applets/blob/7c5ce4c03148c9d0aae05785c7624383313dc589/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/3.2/applet.js#L532 call `this.updateUI()` without any parameter.
2. `updateUI()` https://github.com/linuxmint/cinnamon-spices-applets/blob/7c5ce4c03148c9d0aae05785c7624383313dc589/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/3.2/applet.js#L374 will now not take any argument and the first thing it has to do is to read `/sys/…` into a string (e.g. like here https://github.com/linuxmint/cinnamon-spices-applets/blob/7c5ce4c03148c9d0aae05785c7624383313dc589/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/applet.js#L305). The parsing of the read string obviously has to be changed, but I believe the new parsing will be much simpler.

(Sorry for stating the obvious!)

pdcurtis:
@rcalixte and @sphh . I am happy for you to proceed along the lines suggested. I am just very pleased that others not only find the applet useful but are prepared to put time into improving it. 

I see no good reason to make it a additional PR but can I suggest you do update the change log and version number when you are finished  as it will be a significant improvement. Also make sure you take due credit for the changes between you!

rcalixte:
> @rcalixte: The change from `upower` to `/sys/…` should quite easy:

@sphh I'm leaning in this direction but also extending the functionality to support an option for users to manually enter a different path. (Some users have machines with more than one battery or may want to use a different kernel path/device.) This would be similar to the [similar desklet](https://cinnamon-spices.linuxmint.com/desklets/view/31) I've mentioned previously.


> @rcalixte and @sphh . I am happy for you to proceed along the lines suggested. I am just very pleased that others not only find the applet useful but are prepared to put time into improving it.
> 
> I see no good reason to make it a additional PR but can I suggest you do update the change log and version number when you are finished as it will be a significant improvement. Also make sure you take due credit for the changes between you!

🙏🏽 @pdcurtis Thank you for the fantastic foundation! I've converted the PR to a draft for now and I'll switch it back once it's ready for review.

sphh:
I'm already looking forward to the new version!

rcalixte:
@pdcurtis @sphh I just finished pushing up the changes we discussed previously. I understand if this will take some time to review. If you are interested in testing the applet itself, you can [download the zip file](https://github.com/rcalixte/cinnamon-spices-applets/archive/refs/heads/bams_fixup.zip) and just extract/replace the directory for this applet.

Be careful if you test with the Advanced option for the device path! I opted to have it fail silently (not my usual preference) if the pre-checks fail, rather than filling the logs or annoying the user. (There is a reason it's labelled 'Advanced' after all. 😉) In the worst case, it's easy enough to reset the applet to the defaults in the menu and then reload it. That is the only way to reset to the default path cleanly. (You could always use an invalid path and reload the applet which would use the default kernel directory but it would not be a clean appearance in the settings, a minor inconvenience at best.) I thought about including a stern "do not touch unless you're really sure!" kind of thing but again, I wanted to keep it light.

@pdcurtis I estimate around Cinnamon 6.2 (if versioning stays the way it is going currently) and Linux Mint 22.0 that we can remove the Household Tasks section as it shouldn't be needed anymore. That gives legacy users a chance to cleanup by then and it should be safe to remove nearly two years from now. That's probably the only change I can see in the future, pending feedback.

Happy testing!

pdcurtis:
@rcalixte Something seems to have gone astray as I get a lot of messages like this when I try to look at the files:
`This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.`

I am going to be pretty much out of circulation for several months from tomorrow,  with minimal access to my Linux machines or even internet so I am going to have leave this you and @sphh . I am happy with the principles and approach as discussed earlier so please proceed, however if you break it you have to fix it ! 😉

rcalixte:
> @rcalixte Something seems to have gone astray as I get a lot of messages like this when I try to look at the files: `This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository.`

Right, the link is to the zip file of the branch on the forked repo where this pull request is opened from. It's a download for this entire PR.

> I am going to be pretty much out of circulation for several months from tomorrow, with minimal access to my Linux machines or even internet so I am going to have leave this you and @sphh . I am happy with the principles and approach as discussed earlier so please proceed, however if you break it you have to fix it ! 😉

Safe travels! I'm fine to wait if needed but we'll consider it your blessing if @sphh finds no issues. (If something somehow breaks, we'll fix it before you come back. 😉)

sphh:
I don't believe, that you need my blessing! :innocent:

BTW at that time I was using BAMS, but now moved away from it, because the built-in power manager is good enough at the moment …

I have an heretic thought. As much as I liked BAMS, I notice that there are now two programs/applets doing the same: The built-in power manager and BAMS. It would make much sense to me to get BAMS's superior features into the built-on power manager instead of maintaining this applet. Is this a stupid idea? (When I checked, the power manager is written in C++ and I would not even know, how to add new configuration settings to it.)

rcalixte:
> The built-in power manager

I see some of the overlap but it comes down to preferences.

> It would make much sense to me to get BAMS's superior features into the built-on power manager instead of maintaining this applet.

This would be up to the team to decide? It would need a lot of options if you were to consolidate the two but I like the idea so you've got my vote.

> (When I checked, the power manager is written in C++ and I would not even know, how to add new configuration settings to it.)

Javascript [here](https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/cinnamon/applets/power%40cinnamon.org). It wouldn't be that bad... just someone deciding to do the work. We could technically merge the features in this applet and if the team wanted, they could just replace the Power Manager applet with minimal work. I wouldn't mind porting some of the features on another branch/pull request after this is merged.

sphh:
> I see some of the overlap but it comes down to preferences.

Yes. I like the (old?) Unix philosophy: Do one thing and do it right. Then you only need one program and not a myriad of programs which do it a bit right. Having configuration options is always a good thing (but they must have sensible defaults and probably be ‘hidden’ in an expert's section).

> This would be up to the team to decide? 

Yes, indeed. But if you do not overload the options (see above) and argue your point, they probably accept it. I got some options into the official sound applet – and had to change some and discard one. I believe the developers of Mint are really sensible people!

> Javascript [here](https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/cinnamon/applets/power%40cinnamon.org)

Ah, I see. I believe (but might be wrong) they have split the functionality of BAMS into a [Power Manager applet](https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/cinnamon/applets/power%40cinnamon.org) to show the battery status and set the brightness and a [Power Manager daemon](https://github.com/linuxmint/cinnamon-settings-daemon/tree/master/plugins/power) to handle shutdown and that stuff. The first is written in JavaScript, the other in C.

Please do not let me keep you from updating BAMS!!

rcalixte:
> I believe the developers of Mint are really sensible people!

Strong agreement here. It's the reason I'm thinking so ambitiously about some of the things that can be done.

> Please do not let me keep you from updating BAMS!!

This PR is ready so I'll keep waiting until the merge happens. The team is busy with the new release so it might be a few weeks is my guess.

pdcurtis:
It's well over 30 years since I touched C or C++ but I still have Kernihan and Ritchie on the shelf! 
Off to the port shortly.

sphh:
Bon voyage, @pdcurtis!

PR #4621, opened 2022-12-28
[batterymonitor@pdcurtis] - BAMS 1.5.1 - Improving default battery path logic over hard-coding it

rcalixte:
### 1.5.1
 * Improved the logic for determining the default kernel path (some systems default to BAT1 instead of BAT0)
   - Fixes "-- Warning" permanent message and removes the need for manual configuration in the single battery condition to address it

@pdcurtis (who is on vacation but gives his blessing)

### DIFF
--- a/batterymonitor@pdcurtis/files/batterymonitor@pdcurtis/5.4/applet.js
+++ b/batterymonitor@pdcurtis/files/batterymonitor@pdcurtis/5.4/applet.js
@@ -14,6 +14,7 @@ const St = imports.gi.St;
 const PopupMenu = imports.ui.popupMenu; // Needed for menus
 const Lang = imports.lang; // Needed for menus
 const GLib = imports.gi.GLib; // Needed for starting programs and translations
+const Gio = imports.gi.Gio; // Needed for path/file checks
 const Mainloop = imports.mainloop; // Needed for timer update loop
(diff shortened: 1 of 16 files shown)


rcalixte:
As a note here, I intentionally did not update the other language files right away because there is at least one PR already open and there are no string changes made by this, just line changes. I can make one quick fix PR after everything is merged in so that there are no conflicts with the things currently open but also so that open PRs can be merged as they are fit. Hopefully this doesn't disrupt any process for the duration the line changes are out of sync. 🙏🏽

haggen88:
@rcalixte if it helps, I can close the PR and wait for everything to be updated to redo the translation.

rcalixte:
@haggen88 No, it's fine. Your translation adds value. Worst case, you can update with the translation template here or we can do it after. No strings changed, just the lines changed due to the new block of code so the updates are trivial. Please keep the translations coming!

haggen88:
es.po file according to the .pot template of this PR  
[es.po.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/10320605/es.po.zip)

Regards

rcalixte:
Oh, we could've kept the PR. I'll include it here and update the translation files.

rcalixte:
@haggen88 I updated this PR with your work. (You should be credited for your contributions!)

This is ready for review/merge!

PR #4623, opened 2022-12-29
Two new applets

Severga:


ssm@Severga/files/ssm@Severga/applet.js
```diff
@@ -0,0 +1,452 @@
...
+				let flags = GLib.SpawnFlags.SEARCH_PATH | GLib.SpawnFlags.DO_NOT_REAP_CHILD;
+				let [result, pid] = GLib.spawn_async(null, argv, null, flags, null);
+			}));
+			
+			this.waitingForSensors = false;
+			this.sensorsFailed = false;
+			this.currentTemp = -283;
+			this._netBlackList = [];
+			
+			//this.sensorsPath = "/usr/bin/sensors";
```
mtwebster, line 59: With this commented out do you really need to have lm-sensors as a requirement?

ssm@Severga/files/ssm@Severga/applet.js
```diff
@@ -0,0 +1,452 @@
...
+			this.waitingForSensors = true;
+			Util.spawn_async([this.sensorsPath], Lang.bind(this, this._fetch_cpu_temp));
+			this.timeoutR = Mainloop.timeout_add_seconds(60, () => {
+				this._remove_timeoutR();
+				this.sensorsFailed = true;
+				this.waitingForSensors = false;
+			});
+		}
+		this._remove_timeout2();
+		this.timeout2 = Mainloop.timeout_add_seconds(this.tempRefresh, () => this._update2());
```
mtwebster, line 172: A suggestion on when you have recurring timers:

Instead of starting a new timer at the end of the callback each time, you can simply return `true` (or, for more readability, `GLib.SOURCE_CONTINUE`). When you no longer want the timer to continue, returning `false` (or `GLib.SOURCE_REMOVE`) will remove it (or what you do in `remove_timeout2()` is fine also).

This is fine how you have it, I just wanted to point it out.

timer@Severga/README.md
```diff
@@ -0,0 +1,11 @@
+This is a timer Cinnamon applet.
+
+## Requirements
+This applet requires (optional):
+
+1) sox package (to play sounds);
+2) zenity package (to show system notifications).
```
mtwebster, line 7: You can do perform both playing sounds as well as sending notifications with Cinnamon, and avoid spawning commands - 

for sounds, check out:
https://github.com/linuxmint/Cinnamon/blob/master/js/ui/soundManager.js

For notifications, have a look at this example:
https://github.com/linuxmint/Cinnamon/blob/master/files/usr/share/cinnamon/applets/menu@cinnamon.org/applet.js#L838-L845

mtwebster:
Hi, I added a few comments in the code, mainly just suggestions.

But also, could you squash these into two commits? One for each applet you're adding.

Thanks

ssm@Severga/files/ssm@Severga/applet.js
Severga: I'd been using these lines before I add the settings support. Such commented out lines in the code may be deleted without any harm, but I left them just for myself to remember the names of my variables and what they contain. This property (sensorsPath)'s value is now taken from the applet settings. Maybe I was wrong including this setting. But I wasn't sure my code would be able to find "sensors" executable without user's help. :) My knowledge of Mint's libraries is very bad.

Anyway, you are right. The applet can work without "lm-sensors" at all. And it will do it without errors. Just there will be no color changes according to the CPU's temperature. So I should find other term instead of "requirement".

ssm@Severga/files/ssm@Severga/applet.js
Severga: Thanks for the advice! I really appreciate it!

timer@Severga/README.md
Severga: Thanks again! I will check these out and consider. To tell the truth, I tried to use a part of other applet's code to send notifications but it wasn't working for me, and I decided to use "zenity".

Severga:
Thanks, Michael.
I will edit the code considering your suggestions and squash into two commits as soon as I have free time.

timer@Severga/README.md
Severga: Unfortunately, I cannot make my script play sounds with the internal means of Cinnamon. The `global.display` object in my system doesn't contain `get_sound_player` function. There is `global.play_sound_file` function. But I also can not make it play anything. :(

Severga:
Finished:
1 Simplified System Monitor Applet -> https://github.com/linuxmint/cinnamon-spices-applets/pull/4654
2 Timer -> https://github.com/linuxmint/cinnamon-spices-applets/pull/4655

Issue #4628, opened 2023-01-03
weather@mockturtl Conditions not matching Weather Underground website (plus "!" on icon)


garuffalo:
```
 * Version 3.2.11 (2023-01-01 15:22:22)
 * Cinnamon Version 4.4.8
 * Linux Mint 19.3
 * Intel Mobile GM965/GL960 Integrated Graphics *and* i915 driver

 ```

**Notify author of applet**
@

**Issue**
Applet was reporting "Scattered Showers" with "cloud" icon and exclamation point. Observed that applets current conditions were not what the Weather Underground website was producing (it was actually "Clear" at my location). Same exact lat/long coordinates as website station are used for weather applet. When conditions changed from "Scattered Showers" to something else, the exclamation point icon went away.

Other weather data, such as temperature, were matched to the WU website info, but conditions continue to be a mismatch, although not as radical as the mismatch observed with the exclamation point icon, i.e., "Scattered Showers" vs. "Clear".

So there are actually two issues here: The icon with the exclamation point. -- And, the mismatch of current conditions to the WU weather station displayed on the WU website.

**Steps to reproduce**
Unable to reproduce, as it was observed as a one time occurrence. I am aware of the issue of certain conditions not having a related icon, so perhaps the next time "Scattered Showers" appears, the exclamation point will reappear. Won't know until that happens.

**Expected behaviour**
Would expect the weather applets current conditions to match those of Weather Underground. Also would be good if exclamation point did not appear on icon.

**Other information**
![Screenshot_applet+website](https://user-images.githubusercontent.com/28717740/210401238-d775ba78-352f-4f1c-a54f-98aa624808db.png)

Gr3q:
Unfortunately the data access through the API is very limited, If you registered a weather station you can only query other Personal weather stations and on 5 days of forecast, so sometimes it doesn't have enough data even if it tries to merge observations from multiple stations around your area

Gr3q:
Actually, conditions are only available in forecast, I pull them from there because the applet doesn't support not having them. Even then sometimes they are missing.

garuffalo:
Thanks for the explanation. Too bad the API for Weather Underground is so limited. They have a high rating for accuracy, and I like the phone app and website. I've used Dark Sky since I started using the applet, and that worked really well. But that API is going dark in March, so I am looking for a good replacement.

garuffalo:
[...]

Issue #4625, opened 2022-12-30
weather@mockturtl: A few fixes, one question and a bug


Nightwing0815:
Hey there!
I'm using latest Mint 21.1 and latest version of your applet. Which is great, btw :+1: 
In the last few days, i changed some piece of code, and wanna share and ask:

The question (maybe another one):
I just had the wish, that the applet tooltip shows location, temperature and condition short. My problem is the temperature: how can i get it there? 
where can i increase the size from the icon in the panel?

The bug:
scrolling in hourly forecast shows more than a medic emergency, than smooth scrolling: you can't get back, it's weird...

The fixes:
Your bottombar icon is easily to bring up in the middle, and i corrected the wrong behaviour / wrong switch.

stylesheet.css

```
@@ -1,5 +1,8 @@ - define the box fix

 .popup-menu-content.weather-menu {
+    min-width: 100%;
+    width: 100%;
+    margin: auto;
     padding-bottom: 0em;
     padding-top: 0em;
 }

@@ -86,7 +89,7 @@ - just cosmetics

 .bottombar {
+    padding: 7px 10px 7px 10px;
-    padding: 10px;
 }
```

weather-applet.js

```
@@ -8528,7 +8528,7 @@ - Change text from tooltip
 function GenerateLocationText(weather, config) {
     let location = "";
     if (weather.location.city != null && weather.location.country != null) {
-        location = weather.location.city + ", " + weather.location.country;
+        location = weather.location.city;
     }
     else {
         location = Math.round(weather.coord.lat * 10000) / 10000 + ", " + Math.round(weather.coord.lon * 10000) / 10000;

@@ -8752,7 +8752,7 @@ - ellipsis on russian only? serious? wtf!
     let result = "";
     if (first_temperature != null)
         result = first_temperature;
-    result += ((config._tempRussianStyle) ? ELLIPSIS : ` ${FORWARD_SLASH} `);
+    result += ((config._tempRussianStyle) ? ` ${FORWARD_SLASH} ` : ` ${FORWARD_SLASH} `);
     if (second_temperature != null)
         result += `${second_temperature} `;
     result += `${UnitToUnicode(config.TemperatureUnit)}`;

@@ -15460,13 +15460,13 @@ - just cosmetics
         const sunsetIcon = new Icon({
             icon_name: "sunset-symbolic",
             icon_type: uiSunTimes_IconType.SYMBOLIC,
-            icon_size: 25,
+            icon_size: 24,
[...]

PR #4642, opened 2023-01-10
weather@mockturtl Update 3.2.12

Gr3q:
Changes:

* Some changes from https://github.com/linuxmint/cinnamon-spices-applets/issues/4625
   * Make sure show hourly button icon is pointing to the right direction based on panel orientation.
   * Add option to override tooltip label for applet
* Make "Location Label Override" behavior identical to Panel label override.
* Try to make hourly weather elements slimmer by putting precipitation change and volume in different rows (and only display them if they are supported)
* Only show 1 decimal for precipitation
* Adjust some incorrect condition->icon mappings for MET Norway
* Update luxon and suncalc
* Update webpack because of breaking Node changes with newer OpenSSL versions
* Update README to indicate https://github.com/linuxmint/cinnamon-spices-applets/issues/4628
* Fix https://github.com/linuxmint/cinnamon-spices-applets/issues/4628 properly, only scattered showers was missing
* Fix https://github.com/linuxmint/cinnamon-spices-applets/issues/4644
* Fix MET UK crash when observation stations return a single object instead of an array for it's observation 
* Added editorconfig and formatted files - I recommend hiding whitespace changes
* Add support for saving config from new path and respect ENV variable for xsession-errors

- [x] Test on Cinnamon 3.8 because of external library updates


brownsr:
Just flicking through that 16000 line .js is pretty time-consuming.  Would appreciate smaller updates to review !

Regards
Simon

Gr3q:
I don't recommend reading through the `.js` file, especially when I update npm libs

PR #4652, opened 2023-01-21
[c-eyes@anaximeno] new spice applet Cinnamon Eyes

anaximeno:
Added a new applet named **Cinnamon Eyes** with UUID ***c-eyes@anaximeno*** which displays a small eye on the panel that follows the movement of your mouse, based on the gnome extension *eye-extended-shell-extension*.
screenshot


Drugwash2:
Is that halo around the cursor built by your applet? If so it might be interesting to check and develop into something a bit more... refined.

I got a similar applet but it's based on a different Gnome extension: Glasa. Don't wanna mess with the official applets though, I keep it "[at home](https://github.com/Drugwash2/cinn-r-eyes)". 😎

anaximeno:
Ready for MR.

mtwebster:
There goes my productivity.

Ok to squash?

anaximeno:
> There goes my productivity.
> 
> Ok to squash?

Yes.

Drugwash2:
In Mint 19.2 (Cinnamon 4.2.4) it throws an error related to `try/catch` and doesn't load the applet. Turns out this (older) version of JS requires the format `catch(error) {}` instead of the simpler `catch {}` used in this applet.

Moreover, the cursor circle doesn't appear under any circumstances in the above-mentioned system after fixing the try/catch issue. The applet doesn't even mention in which conditions it should, so a short 'readme' might be in order.

mtwebster:
This isn't going to work at all on pre-5.4 cinnamon. Maybe a 'cinnamon-version' needs to go in the metadata.json file.

Drugwash2:
The eye does work (after fixing the `catch` issue), it follows cursor movement and changes color to yellow when clicked. Only the circle doesn't.
I may look into it at some point.

For the moment it can be released as is with a mention in a readme about non-working circle in older Cinnamon. A minimal 5.4 requirement would be kinda drastic IMHO, but then again it's your call.

anaximeno:
> This isn't going to work at all on pre-5.4 cinnamon. Maybe a 'cinnamon-version' needs to go in the metadata.json file.

Sure, I'll add it.

---

> Moreover, the cursor circle doesn't appear under any circumstances in the above-mentioned system after fixing the try/catch issue. The applet doesn't even mention in which conditions it should, so a short 'readme' might be in order.

I might update the readme with more information, and try to test with earlier versions when having more time, if you find any workaround till then, you can send it to the [original repo](https://github.com/anaximeno/c-eyes) and I'll consider adding it.

mtwebster:
It's mainly the easing calls that prevent backward compatibility (we only added them in 5.4, but still support tweens).

Drugwash2:
@mtwebster Where could I get good info about easing and tweens for both v4 and v5.4+ Cinnamon? I have some problems with the tweener in the multi-line window list mod too.

anaximeno:
@mtwebster I added versions 5.4 and 5.6 to the cinnamon-versions metadata, other versions will be added should they be supported.

Drugwash2:
There's a typo in metadata.json: `multiverion` ==> `multiversion`.
However, if you fix that you'll notice the applet won't appear in the applets list anymore, because the declaration conflicts with the `cinnamon-version` declaration below.
So either loose `multiversion` and restrict usage to only 5.4 and later (the 5.6 should go, it's implicit) or keep `multiversion`, remove `cinnamon-version`, and add separate folders for **4.2** and **5.4** where you put alternative versions of **applet.js**, **helper.js**, **settings-schema.json** and a copy of the **circle** folder - unless you find a way to use a common path for it in both versions of **applet.js**.

For now I fixed another issue that prevented Cinnamon 4.2 to create the circle SVG images, so now it does create and display the circle image - it just doesn't remove it from the screen. The issue is consistent with **mtwebster**'s remark above about easing. Will look into it further, probably try a tweener instead.

anaximeno:
> There's a typo in metadata.json: `multiverion` ==> `multiversion`. However, if you fix that you'll notice the applet won't appear in the applets list anymore, because the declaration conflicts with the `cinnamon-version` declaration below. So either loose `multiversion` and restrict usage to only 5.4 and later (the 5.6 should go, it's implicit) or keep `multiversion`, remove `cinnamon-version`, and add separate folders for **4.2** and **5.4** where you put alternative versions of **applet.js**, **helper.js**, **settings-schema.json** and a copy of the **circle** folder - unless you find a way to use a common path for it in both versions of **applet.js**.

> For now I fixed another issue that prevented Cinnamon 4.2 to create the circle SVG images, so now it does create and display the circle image - it just doesn't remove it from the screen. The issue is consistent with **mtwebster**'s remark above about easing. Will look into it further, probably try a tweener instead.

I'll remove the multiversion and add the lowest compatible version on the metadata, thanks for the explanation on this. I'll also wait a bit for the changes you're making, if you get it to work on that version I'll change the version to a lower one.

Drugwash2:
Compatibility with Cinnnamon 4.2 is practically fixed, it will take just a little bit more time to reorganize settings in a slightly more ergonomical manner.

Drugwash2:
Since I could never wrap my head around those so-called git tools to create diffs and whatnot I have attached the full version of the modded applet. Please feel free to use it as you see fit.
The 4.2 version works in Cinnamon 4.2.4 (Mint 19.2). Haven't tested the 5.4+ version, problems with the virtual machine.

[c-eyes@anaximeno 1.6 MOD.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/10494559/c-eyes%40anaximeno.1.6.MOD.zip)

rcalixte:
> This isn't going to work at all on pre-5.4 cinnamon. Maybe a 'cinnamon-version' needs to go in the metadata.json file.

@anaximeno @mtwebster Wouldn't it be easier to update the metadata and then utilize a versioned directory structure? Creating a `5.4` directory while working on something for the 4.2+ versions of Cinnamon as well? Please stop me if this isn't a reasonable line of thinking.

Drugwash2:
@rcalixte Already done that in the mod attached above. Please check it out if you like. Should work in both 4.x and 5.4+.

anaximeno:
> > This isn't going to work at all on pre-5.4 cinnamon. Maybe a 'cinnamon-version' needs to go in the metadata.json file.
> 
> @anaximeno @mtwebster Wouldn't it be easier to update the metadata and then utilize a versioned directory structure? Creating a `5.4` directory while working on something for the 4.2+ versions of Cinnamon as well? Please stop me if this isn't a reasonable line of thinking.

Would there be necessary to create directories for each version of cinnamon supported? That could be a problem related to the maintainability of the applet, since each version would have to be updated as well. It is not a problem initially, but I will be in the long run.

anaximeno:
@Drugwash2, I made a small fix regarding the mouse circle settings, can you confirm it is working as expected on your computer?

mtwebster:
> Wouldn't it be easier to update the metadata and then utilize a versioned directory structure? Creating a 5.4 directory while working on something for the 4.2+ versions of Cinnamon as well? Please stop me if this isn't a reasonable line of thinking.

I just meant to set a minimum version of 5.4 so it could be merged as is

mtwebster:
You only need to add a version if/when something breaks.. so you may only have a folder for 4.2 and 5.4 - this would cover 4.2 up thru 5.4 (until something breaks compatibility with the 5.4 version), and so on. We try not to break api as much as possible - 5.4 was an aberration, though I still tried as much as possible to maintain compatibility.

Drugwash2:
No it doesn't, at all - the circle is not getting enabled anymore when clicking the eye.
And from what I gather you changed the idea behind the addition of the 'always' switch.
It was added in order to separate the permanent circle from the click circles. I found it annoying to have the circle visible at all times, but it is useful for acknowledging clicks only.

anaximeno:
> You only need to add a version if/when something breaks.. so you may only have a folder for 4.2 and 5.4 - this would cover 4.2 up thru 5.4 (until something breaks compatibility with the 5.4 version), and so on. We try not to break api as much as possible - 5.4 was an aberration, though I still tried as much as possible to maintain compatibility.

Ok, sure.

anaximeno:
> @anaximeno No it doesn't, at all - the circle is not getting enabled anymore when clicking the eye. And from what I gather you changed the idea behind the addition of the 'always' switch. It was added in order to separate the permanent circle from the click circles. I found it annoying to have the circle visible at all times, but it is useful for acknowledging clicks only.

I actually followed the same logic you introduced, but had to make some changes on some small errors I was getting. I am preparing a VM with Linux Mint 19.2 (the same as yours, I guess) to test the changes, I'll give you some feedback after that.

Drugwash2:
OK. There's still a bug when disabling/enabling the 'always' setting. Trying to figure out how to fix it.

c-eyes@anaximeno/files/c-eyes@anaximeno/4.2/applet.js
```diff
@@ -0,0 +1,553 @@
...
+
+		this.settings.bind(
+			"mouse-circle-size",
+			"mouse_circle_size",
+			this.on_property_updated
+		);
+
+		this.settings.bind(
+			"mouse-circle-enable",
+			"mouse_circle_always",
```
Drugwash2, line 108: line 108:
- 			"mouse_circle_always",
+ 			"mouse-circle-enable",

anaximeno:
> P.S. Unfortunately I just found out that when the circle is enabled - even only for clicks - a file drag'n'drop operation to the window manager on an autohidden panel is not possible anymore as the panel will not show up when hovering its respective screen border.
Initially I thought it was the update to Double Commander that created the issue but the operation became available again as soon as I disabled the circle in this applet.

I wrote some notes on the README regarding problems with the panel, since it is a small problem that can be worked around by deactivating the eye click animation, I don't see it as a release blocker. I put it as an issue on the repo, for further analysis on that problem.

Here's the most recent versions with the fixes and updates you made.

[c-eyes@anaximeno.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/10503310/c-eyes%40anaximeno.zip)

I will set this MR as ready, as soon you confirm it is working as expected.

Drugwash2:
Yes, it works. Just please remove the commas at the end of the `debounce` lines - it's unnecesary.

anaximeno:
Alright, ready for MR.

Thank you for the help @Drugwash2, @mtwebster, and @rcalixte.

Drugwash2:
You're welcome. Good luck ahead! 👍

Drugwash2:
Just for your information, I've worked on implementing an idle locator. Basically it works but occasionally there are artifacts - faint traces of the 'circle' remaining on the screen after moving cursor. Also kept the 'always' option available at least for testing purposes.
Won't publish that version yet for your review, there are still things to work on, and I have some real-life issues to deal with more urgently. Will let you know.

For now we can add a slight improvement to the Tweener part in v4.2 in order to replicate the original behavior.
Add the **x**, **y**, **scale_x**, and **scale_y** lines below:

```
		Tweener.addTween(actor,
			{
				x: mouse_x - (this.mouse_click_image_size * actor_scale / 2),
				y: mouse_y - (this.mouse_click_image_size * actor_scale / 2),
				scale_x: actor_scale,
				scale_y: actor_scale,
				opacity: 0,
```

anaximeno:
> Just for your information, I've worked on implementing an idle locator. Basically it works but occasionally there are artifacts - faint traces of the 'circle' remaining on the screen after moving cursor. Also kept the 'always' option available at least for testing purposes. Won't publish that version yet for your review, there are still things to work on, and I have some real-life issues to deal with more urgently. Will let you know.

Sure.

> For now we can add a slight improvement to the [Tweener](https://github.com/linuxmint/cinnamon-spices-applets/blob/fe41ada7fa7f3bafc9e5a07fa79738954147d92b/c-eyes%40anaximeno/files/c-eyes%40anaximeno/4.2/applet.js#L295) part in v4.2 in order to replicate the original behavior. Add the **x**, **y**, **scale_x**, and **scale_y** lines below:
> 
> ```
> 		Tweener.addTween(actor,
> 			{
> 				x: mouse_x - (this.mouse_click_image_size * actor_scale / 2),
> 				y: mouse_y - (this.mouse_click_image_size * actor_scale / 2),
> 				scale_x: actor_scale,
> 				scale_y: actor_scale,
> 				opacity: 0,
> ```

I'll add it in the next version.

PR #4656, opened 2023-01-29
[kdecapplet@joejoetv] Complete Rewrite, Version 2.0.0

JoeJoeTV:
I have rewritten the KDE Connect applet from the ground up to make it modular and add compatability for multiple KDE Connect versions.

New Features:
- Implemented new KDE Connect features(of the version in the LM 21/Ubuntu 22.04 repos)
- Added Settings for many aspects of the applet
- Split functionality into modules for each device to make development easier
- Added more icons to the menu
- Added compatability for multiple KDE Connect versions

Because I've completely rewritten the applet, only a few translation strings stayed the same and most of them have to be redone, so I have deleted the old files da.po, es.po, hu.po and it.po.


rcalixte:
@JoeJoeTV Thank you for this update! I've been using now for the past few weeks.

There's one minor bug that I just noticed. I've been able to replicate it but haven't checked the code to potentially fix it but I figure it's at least worth reporting. When I have multiple devices configured, but only one connected (applet screenshot attached), the option "Expand the first device sub menu automatically" is not honored and the first/only paired device sub menu is expanded. Once I enable/reconnect one of the other previously disabled devices, the menu setting works as expected under both conditions. I haven't tested other conditions but this is just the one that caught my eye that I was able to reproduce. (I'm not sure if this would be the case if only a single device were configured/paired total for example.)

I should add that I've tried reloading the applet, toggling the setting, etc. and I have been able to consistently reproduce at least under these conditions:

Steps to reproduce:
1. Pair multiple devices with KDE Connect.
2. Set "Expand the first device sub menu automatically" to off.
3. Have only one of the previously paired devices connected.
4. Click the applet and see the device sub menu expanded.

Optional steps:
5. Connect a second previously paired device.
6. Click the applet and see the "Expand the first device sub menu automatically" setting correctly honored whether on or off.

Thanks again!

JoeJoeTV:
Hi @rcalixte,

This is actually not a bug, but intended behaviour, as my thinking was, that one would want the first device sub menu to be expanded automatically, if only one device is available, so that one doesn't have to click to expand the sub menu automatically. But if multiple devices are available, I can't predict, which device sub menu the user would want to expand, so I added the setting to expand the first one.

If you want, I can also add another setting, so one can toggle this behaviour or I can put a mention of this behaviour above the setting or as a toolip.

rcalixte:
@JoeJoeTV I'll leave it for future users to decide on if this should be a setting or not. For now, I guess it's best to leave it as is but let the user know about the behavior (maybe only if the setting is set to off).

A tooltip **only** might be missed but maybe rewording "Expand the first device sub menu automatically" to make it clear that the setting only applies when multiple devices are connected might help avoid confusion, along with the tooltip. Maybe "When multiple devices are connect, expand" or something like that... hopefully shorter. 😅

If enough other users comment on it, I think it would make sense to break it out as a separate setting. One for expanding when there is exactly one device connected, one for when there is more than one device connected. Until/unless other users comment on it, it may just be premature work for this one quirky user. 😉

Thanks again!

Drugwash2:
@JoeJoeTV, @rcalixte

Regarding the autoexpand issue, my opinion would be to have this option as a dropdown list containing the options:

- never
- single only
- preferred only
- preferred or single

For a brief explanation:
- single only => expand any device when there is only one connected (ignore `preferred` setting)
- preferred only => expand only the preferred device, if connected, else expand none
- preferred or single => expand preferred device if connected, else expand any device if only one connected, else expand none

Obviously this would require an additional setting for a preferred device but I feel some people may already welcome such addition.

I'm not sure whether adding yet another option to expand the **first** device when multiple are connected would be useful - it all depends on how the items are sorted in the list: by order of addition, by order of connection during the session, aphabetically etc.

rcalixte:
> Regarding the autoexpand issue, my opinion would be to have this option as a dropdown list containing the options:

I really like the idea of the drop-down.

> * never
> * single only
> * preferred only
> * preferred or single

The first two options are easy. If there were a dependent field named Preferred Device that was a text box and that was only visible if one of the "preferred" options was selected, that would be great. If that's too complicated (or can wait until a later release), we could ship without it for now.

The tooltip could contain the finer details for each option as well.

Thanks for the input @Drugwash2!

Drugwash2:
You're welcome. Just random ideas flying through my mind.

> If there were a dependent field named Preferred Device that was a text box and that was only visible if one of the "preferred" options was selected, that would be great.

Actually my thought was to have the preferred device selected from a list (dropdown as well) of already connected devices - which would require updating and saving that list with each new device. Haven't checked the code to see how devices are actually detected and stored but using something similar to a drive's UUID would be most safe in case two or more devices with same name are connected either simultaneously or alternately. Using a simple text entry could not provide this kind of separation and would be prone to typos.

Tooltips should always be present for possibly ambiguous labels, but labels should be as relevant as possible without being too long.

rcalixte:
> Actually my thought was to have the preferred device selected from a list (dropdown as well) of already connected devices - which would require updating and saving that list with each new device.

This would require regular polling of KDE Connect's configuration data and a dynamic drop-down menu in the applet's configuration. Maybe an option in the sub menu for devices in the applet? Something like "Set as preferred device" with the option of whether or not to display the menu option in the settings. (This is getting a bit longer. 😞)


> Haven't checked the code to see how devices are actually detected and stored but using something similar to a drive's UUID would be most safe in case two or more devices with same name are connected either simultaneously or alternately. Using a simple text entry could not provide this kind of separation and would be prone to typos, but there should be a way to edit the names if necessary. Of course the dropdown would display the friendly device names, the only problem being with the identical names - there would have to be some visual hint to differentiate them.

It should/would look the same as the stored devices in the configuration for KDE Connect. As you said, if you have multiple devices of the same name, the device ID should still be unique. A user with duped names in their configuration has many other conflicts to manage already as well.

claudiux:
@JoeJoeTV 
The fr.po I offer in #4683 was made from this .pot file.

Drugwash2:
> Maybe an option in the sub menu for devices in the applet? Something like "Set as preferred device" with the option of whether or not to display the menu option in the settings.

A menu entry would be a friendly and acceptable way to set a device as preferred. But the dropdown list in Settings should still be present in case the user wants to change the preferred device at a later time.
I don't think there's the need for an extra option for showing or not the 'Add as preferred device' menu entry - if that's what you were referring to.

> It should/would look the same as the stored devices in the configuration for KDE Connect.

It's funny because I only installed and used KDE Connect (and this applet) only once before deciding it was not quite what I needed and uninstalled KDE Connect. The applet is still installed though. So I don't really know how devices are stored in KDE Connect's configuration. If it uses unique IDs then it should be fine. New devices with identical names could get automatically renamed to `Device Name [1]`... `Device Name [n]` or similar, and the user should be able to further edit the names in order to personalize them if desired.

JoeJoeTV:
@rcalixte @Drugwash2

The Applet is basically a front end to KDE Connect. It connects to it using it's DBus interface to get all the required information and send actions. The devices have names and UUIDs and also DBus interfaces for each device, so you can use the functions for each device provided by KDE Connect.

I was planning to implement a whole system, where you can reorder devices and modules in the device sub menus, but since I don't see, how you could do this using the default Cinnamon settings widgets, I have to use custom settings widgets, which I haven't fully figured out, how to do. I first wanted to make the applet functional.

The problem is, that I don't think you can create a dropdown menu in settings with dynamic values, only ones set in the settings-schema.json. I have already [asked in the Linux Mint forums](https://forums.linuxmint.com/viewtopic.php?f=120&t=389458) and got a reply by @Drugwash2 actually, but I currently don't have too much time, because of exams and I still need to fully figure out how to get dynamic data correctly into the settings widget and display it. I may look again today, but no promises.


@claudiux 

Should I directly merge the file into this PR or simple wait till both are merged?

Drugwash2:
> The problem is, that I don't think you can create a dropdown menu in settings with dynamic values [...]

You can but it's gonna be tricky. Custom Python script will be required. There's something along these lines in my mod of the **workspace-grid-icons@hernejj** applet. It may be a bit buggy.

Don't rush, follow your priorites. Good luck with the exams!

claudiux:
@JoeJoeTV 

> The problem is, that I don't think you can create a dropdown menu in settings with dynamic values, only ones set in the settings-schema.json. I have already [asked in the Linux Mint forums](https://forums.linuxmint.com/viewtopic.php?f=120&t=389458) and got a reply by @Drugwash2 actually, but I currently don't have too much time, because of exams and I still need to fully figure out how to get dynamic data correctly into the settings widget and display it. I may look again today, but no promises.
> 
I suggest you to create a sub-folder named 'kdecapplet' in the '~/.config' folder, containing a 'devices-data.json' file with all useful / reusable data; and to create in the 'settings-schema.json' of this applet a list containing two fields: Name and Description. The user can modify the order of this list and all descriptions.

> Should I directly merge the file into this PR or simple wait till both are merged?
Sure. Tell me when it's done, please. I'll close my own PR.

claudiux:
@JoeJoeTV My PR about fr.po was just merged.

JoeJoeTV:
@claudiux I just checked and it shouldn't be a problem, since there wasn't a french language file before, so it won't get deleted by this PR. I have deleted the old language files, as they have to be redone.

JoeJoeTV:
@claudix

> I suggest you to create a sub-folder named 'kdecapplet' in the '~/.config' folder, containing a 'devices-data.json' file with all useful / reusable data; and to create in the 'settings-schema.json' of this applet a list containing two fields: Name and Description. The user can modify the order of this list and all descriptions.

Thanks for the suggestion. THat would be an idea, but how can I make it so the user can only change the order of the list elements, because it would only be for moving them around and since one can't edit the UUID and the name from there, it would be annoying to still have that not working functionality in the settings.

That was my exact problem and why I looked at custom settings widgets.

JoeJoeTV:
@Drugwash2

Do you know if there is any documentation about the settings widgets, or do I have to figure the stuff out just from the python code from cinnamon itself?
I'm currently trying to figure out, what some classes and parameters are, like, is the "SettingsWidget" class you inherit you custom settings widget from the one defined in the xapp library or where is the type of the "settings" parameter defined that is passed to the widget?.

Drugwash2:
Unfortunately **real** documentation does not exist. Whatever exists is extremely basic and practically useless. Maybe a little bit of help can be provided by the doc packages for various libraries, if installed, accessible through **Devhelp** but that is far from being either comprehensive or complete.

Unless anyone has better ideas or knows of proper documentation I'd suggest you just look through both javascript and Python scripts in Cinnamon. And to some extent some useful information can be found in the already published x-lets.

The **app-launcher@mchilli** applet has an editable list - made with Python, of course - which can be helpful for your applet as well.
Personally I had to check out other x-lets and just use trial and error in order to achieve the desired results. And if I may say, both javascript and Python are way inferior to AutoHotkey for Windows which I've been working with a lot before switching to Linux. But we use whatever is available for the task at hand.

JoeJoeTV:
I have been working on adding the features for reordering and app-launcher@mchilli and the applet by @Drugwash2 has really helped with the still ongoing implementation of the the settings part.

Drugwash2:
Glad you found a way for advancing with your task. Just please be careful with that Python code, the app-launcher applet has a nasty bug or something which slows down Cinnamon considerably and persistently with every list editing operation, sometimes to the point that Cinnamon must be restarted in order to restore a usable state. At least that's what happens on my daily 19.2 system.

JoeJoeTV:
@Drugwash2 Hmm, do you have a clue, where the bug could come from?
My settings widgets will be simpler than the ones in app-launcher, so maybe they won't have the problem.
When I'm finished, I will push it to the branch and you can test it out, if you want.

The only issues I had while coding the applet that could freeze/crash cinnamon, were writing to many log messages at the same time, when having the looking glass open.

Drugwash2:
Unfortunately I couldn't find out the culprit. Everything seems to work out as expected, no errors or anything, no excess of logging. And it doesn't freeze or crash Cinnamon - just everything becomes more and more sluggish, like opening applications, panels unhiding/hiding etc. CPU usage seems to be increased too. maybe it's just my machine? Haven't seen any similar reports.
Hopefully your applet won't have that problems.

PR #4665, opened 2023-02-01
[Multi-Core System Monitor] Allow applet to create the full folder structure for configuration files

seliver:
@jaszhix 

It looks like the applet was not able to create the ~/.cinnamon/configs folder if it wasn't already in place.

As a workaround while this is not merged and deployed, run the following:

```sh
mkdir -p ~/.cinnamon/configs
```

And then restart cinnamon so that the applets are reloaded (I use Ctrl-Alt-Esc)


mtwebster:
That is no longer the location for configuration files.

They should now be put in `~/.config/cinnamon/spices`.

seliver:
in that case, if we can assume that "${GLib.get_user_config_dir()}cinnamon/spices" already exists, we can drop the `-p` from the `mkdir` command, right?

multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/5.6/applet.js
```diff
@@ -0,0 +1,395 @@
...
+      this.set_applet_icon_path(metadata.path + '/icon.png');
+      this.set_applet_tooltip(metadata.description);
+      return;
+    }
+
+    this.actor.set_offscreen_redirect(Clutter.OffscreenRedirect.ALWAYS);
+    this.childProcessHandler = null;
+    this.metadata = metadata;
+    this.panelHeight = panel_height;
+    this.configFilePath = `${GLib.get_user_config_dir()}/cinnamon/spices/${metadata.uuid}`;
```
seliver, line 165: basically the only thing changed in this PR ☝🏻

Gr3q:
> we can drop the -p from the mkdir command, right?

-p is also for not erroring out if it exists already.

Also you can create a new version, but this is not a big change so I recommend just using a version check against `imports.misc.config.PACKAGE_VERSION`.

brownsr:
Conflicts I'm afraid, since I pulled a more recent change.  Do please check if still needed and adjust accordingly if so.  Applet appears to have been abandoned, so the maintainer is changing to @claudiux 

Regards

Simon

claudiux:
This issue was solved with version 1.9.0 of this applet.

PR #4698, opened 2023-02-26
[Sensors@claudiux] Update German translations

mikehaertl:


claudiux:
Thank you @mikehaertl. That seems OK.

PR #4713, opened 2023-03-07
[kdecapplet@joejoetv] Version 2.1.0

JoeJoeTV:
I have updated the applet to version 2.1.0, which adds support for changing the order of modules for each device sub menu and changing the order of the devices themselves.
It also adds the ability to individually choose, which device sub menus should automatically expand.

Maybe @Drugwash2 could test, if the performance issues he observed with the app-launcher@mchilli applet are happening here too, as I have also used a custom SettingsWidget.

EDIT: I have also noticed, that the screenshot on the  cinnamon-spices page is still the old one, which I had reaplced in the 2.0.0 rewrite pull request. Do I have to do more than that to properly update it?

### DIFF
--- a/kdecapplet@joejoetv/files/kdecapplet@joejoetv/applet.js
+++ b/kdecapplet@joejoetv/files/kdecapplet@joejoetv/applet.js
@@ -160,12 +160,12 @@ function _(str) {
  * @short_description Device class representing a KDE Connect device
  */
 class Device {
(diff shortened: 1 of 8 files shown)


Drugwash2:
My apologies, for the time being I'm in the middle of something that takes up all my time and nerves, so unfortunately won't be able to test the changes anytime soon. If nobody else experiences that type of issues it may well be a local problem, so you shouldn't put much thought into it.

PR #4749, opened 2023-03-16
kdeccc@Severga: Add Simplified Chinese translation

Slinet6056:


Severga:
Please, do not merge. There is an error in the applet.

Severga:
The error will be fixed. And no need to update translation. May be merged.

PR #4777, opened 2023-03-28
multicore-sys-monitor@ccadeptic23: Update Simplified Chinese translation

Slinet6056:


Slinet6056:
@claudiux 
OK, I'll update it later.

PR #4790, opened 2023-04-02
weather@mockturtl: update for Dutch translation for version 3.2.13

Pjotr123:


Pjotr123:
Why is the Chinese translation causing errors for Dutch?

Slinet6056:
The zh_CN.po file has some errors after merging into the master branch from two different PR. It should be fixed by #4791.

PR #4811, opened 2023-04-15
[vboxlauncher@mockturtl] v0.4.1 - Make this applet 'not dangerous' - Add option 'Show Headless Mode'

claudiux:
@mockturtl

I offer a version 'not dangerous' of your applet, without Danger signal:
Capture d’écran du 2023-04-15 19-21-40

Also, I added a "Show Headless Mode" option (set to true by default) to allow users (like me) who never use headless mode to get a menu with direct access to their virtual machines.

I hope this will suit you.

Regards
Claudiux


mockturtl:
@claudiux Thank you!  I haven't used the vboxlauncher applet for several years.  Feel free to merge!

PR #4837, opened 2023-05-02
wallpaper-slideshow: Add applet

amaan-mohib:
# Wallpaper Slideshow

Starts a wallpaper slideshow persisting across reboots

## Need

Wallpaper slideshow already exists in Cinnamon, but the slideshow always starts from the beginning whenever rebooted.

So, to fix that this applet allows a wallpaper to alteast stay for the given delay irrespective of reboots.


brownsr:
is the gsettings key right ?  org.cinnamon.desktop.background

Regards

Simon

amaan-mohib:
> is the gsettings key right ? org.cinnamon.desktop.background
> 
> Regards
> 
> Simon

@brownsr yes, verified it from the rwall applet as well as nemo

PR #4849, opened 2023-05-11
Soup 3 cleanup

mtwebster:
We're hopefully going to switch Cinnamon to requiring soup 3 for 5.8 (distros are starting to drop soup 2), so I went thru and tried to update applets that hadn't been given soup 3 compatibility yet.

Most of these I couldn't test the actual http requests as they require accounts and api keys, but I hope to at least get the ball rolling so maybe the appropriate maintainers can cherry-pick from this, or I or they can open an individual PR for their applet.

@claudiux (VPN-Sentinel@claudiux)
@ludvigbostrom (devbar@ludvigbostrom)
@inventor96 (wakatimedisplay@inventor96)
@ImmRanneft (nightscout@ranneft)
not sure about others...

Thanks

PR #4852, opened 2023-05-13
[VPN-Sentinel@claudiux] v2.1.0 - Now compatible with Soup3

claudiux:
Related to discussion in #4849


claudiux:
Thank you @brownsr

PR #4792, opened 2023-04-02
power-profiles@rcalixte: Initial release of a power-profiles-daemon applet

rcalixte:
Closes linuxmint/cinnamon#10634

PR #4853, opened 2023-05-14
system-controls@rcalixte: Add Portuguese translation

rcalixte:


rcalixte:
@brownsr It would also be great to get eyes on #4792 as well. Thanks!

PR #4861, opened 2023-05-21
Add PanelDo

BeatLink:
Panel based todo applet for the Cinnamon desktop.


fredcw:
Some test results:

When clicking the applet it tries to open `~/Documents/todo.txt` in xed but the file opened is `~/~/Documents/todo.txt` which can't be saved. It seems that

`Util.spawnCommandLine(this._preferences.list_editor + " " + this._preferences.path_to_file);`

doesn't expand the "~"

Also line 46

`if (GLib.file_test(this.file, GLib.FileTest.EXISTS)) {`

doesn't seem to find the file when it does exist, maybe for the same reason.

An absolute path+filename does work as does "./Documents/todo.txt"

BeatLink:
I've changed the file path. Give it a try.

brownsr:
Could you please set yourself as the author, and not put the original author in, please ?  That's how we show abandoned applets, so would be a little confusing as is.

Regards

Simon

PR #4869, opened 2023-05-24
sysmonitor@orcus: added scaling functionality for network graph

SebiTimeWaster:
Added a minimum scaling user setting for the network graph since it is the only graph that does not have a fixed maximum (like 100% CPU or known usable RAM value) so the user can set a "soft" maximum by specifying the minimum scaling of the graph.

If the value exceeds the specified minimum the graph is auto-scaled.

The standard setting of "0" produces the same behaviour as it does without this PR.


SebiTimeWaster:
@OrcusCZ

SebiTimeWaster:
what happens if the owner does not react?

OrcusCZ (approved):

OrcusCZ:
@SebiTimeWaster Hi, I'm sorry for the delay. I wanted to check it out when you created it, but got distracted and forgot about it.
Thanks for the PR, looks good, approved.

SebiTimeWaster:
> @SebiTimeWaster Hi, I'm sorry for the delay. I wanted to check it out when you created it, but got distracted and forgot about it. Thanks for the PR, looks good, approved.

@OrcusCZ  no problem, this happens, i know that myself ^^

PR #4887, opened 2023-06-07
[Sensors@claudiux] v3.0.2 - Fixes the double separator error

### LINKED ISSUE #4886: [Sensors@claudiux] Double separator error
```
Applet version: 3.0.1
Cinnamon version: 5.6.8
Distribution: LMDE5
```
**Notify author of applet**
@claudiux

**Issue**
After last commit [8171665f] Sensors@claudiux: added functionality to user define a separator string (#4868)
all entered separators are displayed double `||` instead of `|`, `//` instead of `/` and so on

claudiux:
Fixes #4886.


claudiux:
Nothing more to add.

claudiux:
Thank you @brownsr.

Issue #4909, opened 2023-06-22
[desaturate-all@hkoosha] Error clicking on applet


claudiux:
```
 * Desaturate All (desaturate-all@hkoosha) 2023-06-06 21:28:22
 * Cinnamon version 5.6.8
 * Distribution Mint 21.1
 * Graphics hardware *and* driver used: no matter
 * Snippet or attachment of ~/.xsession-errors:
Gjs-Message: 08:49:39.664: JS WARNING: [/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function 26]: Too many arguments to method Clutter.Actor.has_effects: expected 0, got 1

 ```

**Notify author of applet**
@hkoosha

**Issue**
Each time the applet is clicked, an error message like the one above appears in the `~/.xsession-errors` file.

**Steps to reproduce**
Click on the applet. (I'm using the Xsession applet to watch the contents of  `~/.xsession-errors`.)

**Expected behaviour**
No error messages appear.

**Other information**
In applet.js, changing in line 24:
`if (Main.uiGroup.has_effects(this.effect)) {`
by:
`if (Main.uiGroup.has_effects() && Main.uiGroup.get_effects().indexOf(this.effect) > -1) {`
does the job.

hkoosha:
Thanks for reporting this! PR opened.

claudiux:
Thanks for reacting so quickly!

PR #4910, opened 2023-06-22
[desaturatea-all@hkoosha] replace call to `has_effects(this.effect)`

hkoosha:
Each time the applet is clicked, an error message like the following appears in the ~/.xsession-errors file.

```
Gjs-Message: 08:49:39.664: JS WARNING: [/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function 26]: Too many arguments to method Clutter.Actor.has_effects: expected 0, got 1
```

Addresses https://github.com/linuxmint/cinnamon-spices-applets/issues/4909

Tested on latest version of Manjaro-Cinnamon.


claudiux:
Tested on Mint 21.1. All is OK.

PR #4920, opened 2023-07-02
Remove erroneous top-level files

rcalixte:


fredcw:
lol. Nobody tell @claudiux

claudiux:
@fredcw Thank you.
@rcalixte I don't understand why these files are there. Surely they must be deleted.

Issue #4641, opened 2023-01-10
CinnVIIStarkMenu@NikoKrause Visual artifact on (All Programs) panel


scoson:
```
 * CinnVIIStarkMenu version 1.31
 * Cinnamon 5.6.5 
 * Linux Mint 21.1 Vera
 * Graphics:
  Device-1: Intel GeminiLake [UHD Graphics 600] vendor: Hewlett-Packard
    driver: i915 v: kernel bus-ID: 00:02.0
  Device-2: Luxvisions Innotech HP TrueVision HD Camera type: USB
    driver: uvcvideo bus-ID: 1-6:4
  Display: x11 server: X.Org v: 1.21.1.3 driver: X: loaded: modesetting
    unloaded: fbdev,vesa gpu: i915 resolution: 1366x768~60Hz
  OpenGL: renderer: Mesa Intel UHD Graphics 600 (GLK 2) v: 4.6 Mesa 22.0.5
    direct render: Yes


 ```

**Notify author of applet**
@NikoKrause

**Issue**
I've noticed the issue shown below since Mint Cinnamon 21 Vanessa release. It seems 2 others report the same on the Spices Applet page. I just did a clean install of Vera and am using Mint-y-blue theme and icons. I had to actually take a pic of this as it disappears with screenshot.

![IMG_20230110_141346_NR resized](https://user-images.githubusercontent.com/46092441/211661201-c04102ac-b1c4-4af1-969c-f0954c37def4.jpg)

fedaykin451:
I have noticed the exact same issue since upgrading to Mint Cinnamon 21, as well.  I notice it most often while scrolling or moving the cursor from right to left in the menu.

scoson:
Wondering if an update somehow fixed this issue within the last week? It seems to have resolved. I haven't seen this issue occur in several days. Just sayin..

fedaykin451:
> Wondering if an update somehow fixed this issue within the last week? It seems to have resolved. I haven't seen this issue occur in several days. Just sayin..

Still an issue for me.  I have noticed it doesn't reproduce every time.  Some days I don't see it at all.

scoson:
@fedaykin451 Yeah, I was mistaken, still a problem here to.

fedaykin451:
I installed today's update (2/7/2023), problem still occurs.

fedaykin451:
Problem still occurring after most recent update.

anaximeno:
[...]

PR #4927, opened 2023-07-07
[CinnVIIStarkMenu@NikoKrause] Fix bug causing triangle to appear on menu.

fredcw:
Use a different method to prevent category button missellection by calculating tangent of mouse movement instead of St.polygon leave event

Fixes: #4641

@NikoKrause


NikoKrause:
@fredcw Thanks for fixing that bug. I'll try to test it this weekend. (I'm still on 20.3 and need to upgrade to Mint 21 first.)

Do you know if this behaviour changed in the default Cinnamon menu as well?

fredcw:
@NikoKrause Actually, the main menu sets all the category buttons to reactive=false to prevent mis-selection. Your method was more concise but unfortunately there's some visual bugs in St that caused the St.polygon to appear randomly. My method, also used in Cinnamenu applet, simplifies things by using a fixed angle of about 73 deg above or below 3 o'clock instead of calculating the angles to the categoryBox corners. Now I think about it, it should be fairly easy to adapt my method to calculate angles to corners but the fixed angles seem to work pretty well anyway.

PR #4936, opened 2023-07-10
cpufreq@mtwebster: adding settings for configuring the lower and upper border

tomk3003:
@mtwebster: please have a look at this.

I just added two sliders for setting a percentage for the lower and upper border and modified num_to_color() to use the values calculated from that. This is straight forward and works for me out of the box after reloading the applet.


brownsr:
ping @mtwebster :-)

rcalixte (approved):

PR #4938, opened 2023-07-12
[spices-notifier@germanfr] v4.2 - Improves access to spices website

claudiux:
Leave a delay of 1 second between two requests to the spices site, so we don't get blocked.

Add a 'Refresh' option to the contextual menu.

### DIFF
--- a/spices-notifier@germanfr/files/spices-notifier@germanfr/applet.js
+++ b/spices-notifier@germanfr/files/spices-notifier@germanfr/applet.js
@@ -13,6 +13,7 @@ const http = new HttpLib();
 
 const UUID = 'spices-notifier@germanfr';
 
+const _TYPES = ['themes', 'applets', 'desklets', 'extensions'];
 const SPICES_URL = 'https://cinnamon-spices.linuxmint.com';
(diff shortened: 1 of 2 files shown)


claudiux:
Ping @fredcw @germanfr

fredcw:
The setting `this.uuidList` is used in on_xlets_loaded() but this setting isn't set anywhere?

claudiux:
You're right.

claudiux:
Everything seems fine now.

fredcw:
@claudiux Yes, seems to be working ok, thanks.

btw, it doesn't make any difference to the functioning of the applet but unlike regular functions, arrow functions take `this` from the local scope so you don't need `Lang` on line 183.

```
setTimeout(() => {
  this.get_xlets(_type);
  return false
}, 1000); // 1 second
```
and e.g.

`reload_all.connect('activate', () => this.reload());`

can be used instead of:

`reload_all.connect('activate', this.reload.bind(this));`

claudiux:
@fredcw Okay, thanks. I'll take that into account for the next version. I'd also like this applet to take into account the change in the number of stars.

PR #4953, opened 2023-07-19
[windowlist@cobinja.de] Remove unnecessary padding

nrdave:
mentioning @Cobinja so they see this

windowlist@cobinja.de, aka CobiWindowList, takes up some space in the panel that appears to be unused for anything. See below for examples in both horizonal and vertical panels.

horiz-old
vert-old

After some trial and error, I found that this extra space is primarily taken up by the CobiWorkspace module (class? element?). See below for screenshots showing how CobiWorkspace takes the space. The blue border is the applet itself, and the red border is just CobiWorkspace.
horiz-old-border
vert-old-border

After getting these screenshots, I noticed that on the horizontal panel, the right side of CobiWorkspace does not have any padding, which (aside from the padding looking bad) makes me think the padding wasn't intentional.

The changes I made remove this padding, by setting padding to 0 for CobiWorkspace. See below for results (both with and without borders):
horiz-new
horiz-new-border
vert-new
vert-new-border

The applet itself still has padding, and I'm not sure how multiple CobiWorkspace modules could be created, so I don't think this change causes any problems. However, one could alternatively reduce the padding in CobiWorkspace and remove the applet's own padding.

I only tested this on my system, running Cinnamon 5.8.4, but given that the CobiWorkspace onOrientationChanged function is the same across all versions, I doubt this will be an issue.

### DIFF
--- a/windowlist@cobinja.de/files/windowlist@cobinja.de/3.2/applet.js
+++ b/windowlist@cobinja.de/files/windowlist@cobinja.de/3.2/applet.js
@@ -1505,12 +1505,12 @@ CobiWorkspace.prototype = {
     if (orientation == St.Side.TOP || orientation == St.Side.BOTTOM) {
       this.actor.set_vertical(false);
       this.actor.remove_style_class_name("vertical");
-      this.actor.set_style("margin-bottom: 0px; padding-bottom: 0px; margin-top: 0px; padding-top: 0px;");
+      this.actor.set_style("margin-bottom: 0px; padding-bottom: 0px; padding-left: 0px; margin-top: 0px; padding-top: 0px;");
     }
     else {
       this.actor.set_vertical(true);
       this.actor.add_style_class_name("vertical");
-      this.actor.set_style("margin-right: 0px; padding-right: 0px; padding-left: 0px; margin-left: 0px;");
+      this.actor.set_style("margin-right: 0px; padding-right: 0px; padding-left: 0px; padding-top: 0px; padding-bottom: 0px; margin-left: 0px;");
     }
     for (let i = 0; i < this._appButtons.length; i++) {
       this._appButtons[i]._updateOrientation();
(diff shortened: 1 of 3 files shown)


klangman:
I tried to fix that a few times. Wonder why padding left & right 0 is not the same as "padding: 0px;"??

nrdave:
For the horizontal panel orientation specifically, I didn't actually need to add `padding-right: 0px` to the style, as for whatever reason there was no padding on the right. I just decided to change `this.actor.set_style("margin-bottom: 0px; padding-bottom: 0px; padding-left: 0px; margin-top: 0px; padding-top: 0px;");` to `this.actor.set_style("margin-bottom: 0px; margin-top: 0px; padding: 0px;");` for the sake of brevity and future-proofing (in case whatever caused the padding right to be 0 changes).

I hope I understood the question correctly.

(If you were talking about the vertical panel orientation, then both `padding-left: 0px` and `padding-right: 0px` are part of the style, and since I added `padding-top: 0px` and `padding-bottom: 0px`, it just made sense to just say `padding: 0px`)

Cobinja:
The padding comes from the theme. I only added those 0-values so that the applet fills the height of horizontal panels and the width of vertical panels.
Every applet that uses the css class "window-list-box" is affected for your used theme. Open an issue there.

Rejected because applets should undo themes as little as possible.

nrdave:
Fair enough. Might be worth it to document that somewhere, though

klangman:
Ah.. I didn't read the code carefully enough. I thought that the top/bottom was being set for vertical panels and the left/right was being set for horizontal panels. So I was confused why the change did anything.

So the solution here is to use "grouped-window-list-box" rather then "window-list-box" since the grouped-window-list applet does not have that annoying padding when using Mint-Y theme anyhow.

Cobinja:
I noticed that my applet used the css class .window-list-box for the container actor, but already had .grouped-window-list for everything else. I streamlined this in the marked patch.

Issue #4851, opened 2023-05-12
Question Regarding Usage : calc-js@ptandler


henri:
@ptandler

Just wondering how you commit a calculation to the history. Do you press a button or have to format the expression a certain way?

**Steps to reproduce**
1 + 2 [press enter key]

**Expected behavior**
see the expression and possibly the result in history

**Other information**
I am guessing this is not a bug - but just a misunderstanding of using this excellent applet?

ptandler:
What you describe as expected behavior is exactly how it should work: You type an expression and press enter and it shows up in the history.

Actually the result should show up immediately after typing, no need to press enter.

So this does not work for you?

Maybe, there is something in the logs, that gives a hint what is different in your environment?
https://stackoverflow.com/q/14025722/1480587

fredcw:
Same here. History menu just remains greyed out. Nothing in ~/.xsession-errors.

ptandler:
but evaluating the expression works?

ptandler:
ah, maybe you use a different cinnamon version?

I still have 5.2.7 on my machine (and mint 20.3 as the upgrade to 21 did not work)

I think I read that the CJS version was upgraded with this release

fredcw:
@ptandler Yes, i tested on cinnamon 5.6 and 5.8. Cjs is updated in 5.8 but that shouldn't make a difference. Clutter was updated in 5.4 https://github.com/linuxmint/cinnamon-spices-applets/issues/4364

PR #4996, opened 2023-08-09
[calc-js@ptandler] Fix enter/return key not recognised + history.

fredcw:
Fix enter/return key was not recognised meaning history didn't work.

Fixes: #4851

@ptandler

### DIFF
--- a/calc-js@ptandler/files/calc-js@ptandler/applet.js
+++ b/calc-js@ptandler/files/calc-js@ptandler/applet.js
@@ -335,7 +335,8 @@ class MiniCalc extends Applet.IconApplet {
         const keySymbol = event.get_key_symbol();
         // logInfo("keySymbol = ", keySymbol, " control = ", event.has_control_modifier());
         switch (keySymbol) {
-            case Clutter.Return:
+            case Clutter.KEY_KP_Enter:
+            case Clutter.KEY_Return:
             case Clutter.KP_Enter:
                 this.pushToHistory(input, result);
                 this.updateExpression(result);


rcalixte:
Has this been tested for backwards compatibility? Would it make sense to put the fix into a 5.4 directory and leave the existing in something like a 3.0 directory?

fredcw:
I'm pretty sure this is backward compatible because it's the same named constants I use in Cinnamenu but I'll check anyway...

fredcw:
Yes, tested on Mint 20.3 (cinnamon 5.2) which is an earlier version of Clutter and both keypad enter and return work so I guess `Clutter.KEY_Return` was backward compatible.

Btw, this applet doesn't work on Mint versions prior to 20.1 (cinnamon 4.8) due to it's use of `Intl.NumberFormat().formatToParts()` I guess this should probably documented in the readme as I don't think there's any way to prevent it from being installed on earlier versions of Mint and just failing to work.

Drugwash2:
> Btw, this applet doesn't work on Mint versions prior to 20.1 (cinnamon 4.8) due to it's use of `Intl.NumberFormat().formatToParts()`

It will work with a simple modification:

`this.numberDecimalSeparator = Intl.NumberFormat().format(1/10).replace(/\d/g, "");`

instead of

```
const numberParts = Intl.NumberFormat().formatToParts(1234.5);
this.numberDecimalSeparator = numberParts.find(part => part.type === 'decimal').value;
```

in **[initializeNumberFormat()](https://github.com/linuxmint/cinnamon-spices-applets/blob/91ffa7b153bab2e2518258d7f67e4974157da15c/calc-js%40ptandler/files/calc-js%40ptandler/applet.js#L125)**.
Of course this was very shortly tested in Mint 19.2 (Cinnamon 4.2.4) using en_US default locale.
If anyone has more extensive testing capabilities they could confirm - or not - the validity of this change.

It could even be put in a `try/catch` in order to retain the newer functionality for a future time when backward compatibility won't be an issue anymore:

```
initializeNumberFormat() {
  // see e.g. https://stackoverflow.com/a/51411310/1480587 and https://stackoverflow.com/a/62694190/1480587
  try{
    const numberParts = Intl.NumberFormat().formatToParts(1234.5);
    // e.g. [{"type":"integer","value":"1"},{"type":"group","value":"."},{"type":"integer","value":"234"},{"type":"decimal","value":","},{"type":"fraction","value":"5"}]
    // this.numberGroupSeparator = numberParts.find(part => part.type === 'group').value;
    this.numberDecimalSeparator = numberParts.find(part => part.type === 'decimal').value;
  } catch(e){
    this.numberDecimalSeparator = Intl.NumberFormat().format(1/10).replace(/\d/g, "");
  }
  // helpers for `updateResult()` in case `this.opt.convertLocaleNumberSeparators` is set
  // build reg exps to search for group separators (should be removed) ...
  // this.numberGroupRegExp = new RegExp(`\\d${escapeRegExp(this.numberGroupSeparator)}\\d`);
  /// ... and then for decimal separators (which should be replaced by ".")
  this.numberDecimalRegExp = new RegExp(`(\\d+)${escapeRegExp(this.numberDecimalSeparator)}(\\d+)`, "g");
}

```

ptandler:
Thank you so much for fixing this :heart_eyes: 

Should we move the "Mint versions prior to 20.1 (cinnamon 4.8)" issue to a separate issue/PR?

I think the approach with try/catch is a good way to provide compatibility.

I don't have a Mint < 20.3 at hand - and currently I also don't have much time to work on this.

PR #5004, opened 2023-08-16
InternetTimeApplet@stefan: Fix timezone for accurate Beat Time calculation

christopher-conley:
Change the hour/minute/second variable assignments to explicitly be in UTC format. Also, add one hour to the "getUTCHours()" call, as "Biel Mean Time" (BMT), which Beat Time is based upon, is defined by Swatch as UTC+1.

Previously this was calling Date's getHours/getMinutes/getSeconds function family, which returns LOCAL time. This had the effect of the Beat time returned by this desklet always being incorrect unless your computer's time is set to the UTC+1 timezone.

This commit fixes that and ensures that the correct universal Beat Time will be displayed regardless of the user's timezone. It also incorporates the calculation changes from @nstoeckigt via his previously-merged PR [here.](https://github.com/linuxmint/cinnamon-spices-desklets/pull/784)

Original author of this applet is @stefan12O, and I'm submitting this PR per his request [here.](https://github.com/linuxmint/cinnamon-spices-desklets/pull/909#issuecomment-1679223083)

### DIFF
--- a/InternetTimeApplet@stefan/files/InternetTimeApplet@stefan/applet.js
+++ b/InternetTimeApplet@stefan/files/InternetTimeApplet@stefan/applet.js
@@ -26,11 +26,17 @@ NetBeatApplet.prototype = {
 
 	refresh: function() {
 		let d = new Date();
-		let h = d.getHours();
-		let m = d.getMinutes();
-		let s = d.getSeconds();
-		let tzoff = 60 + d.getTimezoneOffset();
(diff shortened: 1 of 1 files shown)


stefan12O:
As agreed, thank you again.
Btw., I wasn't aware, that there was a change by Nico in 2022, somehow this bypassed my attention.

stefan12O (approved):

rcalixte (approved):

PR #5006, opened 2023-08-21
[CassiaWindowList@klangman] Add new smart numeric hotkeys

klangman:
1. Added an option in the "Hotkeys" setting tab (enabled by default) called "Smart numeric hotkeys" which allows you to switch to the first 9 windows of a specific application as ordered on the window list. Creating a Hotkey containing the "1" key will automatically be extended to 1-9. For example, using a key sequence of "Alt+1" with a "description" of "firefox.desktop" will register hotkeys Alt+1 through Alt+9. Using those hotkeys will switch to one of the first 9 firefox windows based on the order of the windows on the window-list.

2. Fix a number of issues with hotkey sequence text not being properly added to button tool-tips.

3. Fix a number of issues with hotkeys not working properly after deleting, rearranging or modifying hotkey entries.

4. Fixed some issue around deciding which hotkeys are available for manual assignment using the button context menu.

5. When a windows titlebar is not visible on the screen, a new context menu item "Move titlebar on to screen" has been added to the button context menu which will move the window so it's titlebar will be visible.

6. Fixed an issue where sometimes the "urgent" status of a pinned button will remain even after the window that caused the urgent status is closed.

7. Fixed some other minor bugs and some issues running on Mint 19.x

### DIFF
--- a/CassiaWindowList@klangman/files/CassiaWindowList@klangman/4.0/applet.js
+++ b/CassiaWindowList@klangman/files/CassiaWindowList@klangman/4.0/applet.js
@@ -47,6 +47,7 @@ const Settings = imports.ui.settings;
 const SignalManager = imports.misc.signalManager;
 const CinnamonDesktop = imports.gi.CinnamonDesktop;
 const ModalDialog = imports.ui.modalDialog;
+const Config = imports.misc.config;
 
 const UUID = "CassiaWindowList@klangman";
 
@@ -119,6 +120,7 @@ const ICON_NAMES = {
    writer: 'x-office-document'
 }
 
+const majorVersion = parseInt(Config.PACKAGE_VERSION.substring(0,1));
 
 // The possible user setting for the caption contents
 const CaptionType = {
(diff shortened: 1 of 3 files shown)


CassiaWindowList@klangman/files/CassiaWindowList@klangman/4.0/applet.js
```diff
@@ -389,6 +406,72 @@ function getKeyAndButtonMouseAction(mouseActionList, modifier, context, mouseBtn
    return -1;
 }
 
+function moveTitleBarToScreen(window) {
+   if (majorVersion < 5) {
```
rcalixte, line 410: You can move the applet code into versioned folders to avoid having to introduce these kinds of conditionals. Take a look at other applets for an example.

rcalixte (approved):

CassiaWindowList@klangman/files/CassiaWindowList@klangman/4.0/applet.js
klangman: I did consider doing that, but went with this approach for now to keep the code common for a little longer. I'll give up and split off to a version 5 source tree if too much more of this ugliness starts creeping in.

PR #5023, opened 2023-09-03
darkMode@linuxedo: add support to XDG Desktop Portal

Wilfison:
Add support to XDG Desktop Portal.

XDG Desktop Portal support was added to all supported desktops (Cinnamon, MATE and Xfce). Among other things this brings a global dark mode setting.

This setting only affects applications which support dark mode.


darkMode@linuxedo.com/README.md
```diff
@@ -8,13 +8,18 @@ For more details and feature requests: [Dark Mode in Cinnamon Desktop Environmen
 License: [GPL-v3](https://github.com/linuxedo/cinnamon-dark-mode-applet/blob/main/LICENSE)
 
 ## Changelog
-* Jul 12, 2022
+
+- Set 3, 2023
```
rcalixte, line 12: Small typo here.

rcalixte, line 12: Typo here.

rcalixte (changes requested):

rcalixte:
cc @slgobinath

darkMode@linuxedo.com/README.md
```diff
@@ -8,6 +8,10 @@ For more details and feature requests: [Dark Mode in Cinnamon Desktop Environmen
 License: [GPL-v3](https://github.com/linuxedo/cinnamon-dark-mode-applet/blob/main/LICENSE)
 
 ## Changelog
+* Set 3, 2023
```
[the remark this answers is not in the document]
Wilfison: Correct, my mistake in the translation

rcalixte (approved):

rcalixte:
@Wilfison Have you tested this on versions before Cinnamon 5.8? I'm not sure if this change needs to be implemented in a versioned structure or not. I'll let @slgobinath chime in too before merging.

rcalixte:
> @rcalixte You were right, I added a version check for the minimum supported to be 5.8

You should make use of directory versioning, similar to all of the other applets. This will remove the need for a conditional in the code. You can take the code from before this PR and move it to a 3.0 directory and then put the code for the new PR in a 5.8 directory. This simplifies the code for future maintenance. If you need a hand, I can help with this. Just say the word! Thanks!

Wilfison:
@rcalixte Thanks for the feedback. I will follow your instruction.

Do you think it's better for me to close this pull request with multiple commits and create a new one with just a single commit?

rcalixte:
> Do you think it's better for me to close this pull request with multiple commits and create a new one with just a single commit?

It's fine. We can squash and merge so that the final merge will just be one commit.

rcalixte:
@Wilfison I went ahead and pushed a few commits to clean things up a bit. We should be good to merge now. Thanks for your contribution!

sudoshindo:
@slgobinath @rcalixte @wilfison 

Hello.

I made a suggestion over at the spices applet page, but I just want to make sure the comment is seen.

Here is a copy of what I said there:

**Suggestion:**

Make the background specific/explicit instead of random. Setting up a light/dark wallpaper pair is kind of a hassle because you have to make 2 directories containing only 1 image each. It might be better if you get rid of the "random selection" part of the code so that the wallpaper files are easier to manage for users.

It would allow us to have a `/wallpapers/sunset` dir that contains `sunset-dark.png` and `sunset-light.png` instead of doing something like `/wallpapers/dark/sunset.png` and `/wallpapers/light/sunset.png` and we could even avoid grouping the pairs into subdirectories if we wanted to since we could specify the exact image paths.

PR #5035, opened 2023-09-22
[octopussy@centurix] Add confirmation dialogs for dangerous actions

EvilGremlin:
Another small fix. I don't know what i'm doing, please tell if there is better way to do such dialogs.


rcalixte:
cc @Centurix

Centurix (approved):
Nice, thanks!

rcalixte (approved):

PR #5038, opened 2023-10-04
Create Issue form/template

rcalixte:
Similar to linuxmint/cinnamon-spices-desklets#925
Similar to linuxmint/cinnamon-spices-extensions#486
Similar to linuxmint/cinnamon-spices-themes#759


fredcw:
This seems to mostly unnecessary and I don't see how it's an improvement The only real issue atm is that the proper author name is often omitted or incorrect. A drop down list would be the perfect solution but there seems to be a problem atm because, in applets at least, it's too long to fit on the page so half the applets are missing and won't this also have to be updated after every new applet/author change?

Also many of the required fields don't need to be required - "steps to reproduce" and "expected behaviour" are often too obvious to need stating and even "cinnamon version" and "distribution" sometimes don't need stating because again, it might be obvious to both the bug reporter and the applet author e.g "after last update, feature x stopped working". And the "code of conduct" tick box is not only pointless, it could itself be seen as offensive or disrespectful.

rcalixte:
First, thanks for the feedback.


> This seems to mostly unnecessary and I don't see how it's an improvement The only real issue atm is that the proper author name is often omitted or incorrect. A drop down list would be the perfect solution but there seems to be a problem atm because, in applets at least, it's too long to fit on the page so half the applets are missing and won't this also have to be updated after every new applet/author change?


I hadn't seen the issue with the dropdown missing elements but my displays are also higher resolutions so it could be a blind spot on my end. I'll test further when I get a chance. But yes, the long-term idea is that this will all be fully automated at some point. The discussion is here: linuxmint/discussions#65.


> Also many of the required fields don't need to be required - "steps to reproduce" and "expected behaviour" are often too obvious to need stating and even "cinnamon version" and "distribution" sometimes don't need stating because again, it might be obvious to both the bug reporter and the applet author e.g "after last update, feature x stopped working".


This is all fluid and I left the PR open for others to review. The fields are free-form where no logic is being parsed so any text input would suffice. Hopefully this is a nice middle of the road. I agree that in the situations where things are more obvious, this could be undue, but there are other avenues for that feedback to reach the author, such as the Spices website. In situations where recent updates cause issues for users, it also quite often ends up being versions of Cinnamon different from what the author has available (either older or newer) or other distributions where there are enough differences for issues to present themselves, such as package or library dependencies. Gathering that input when the Issue is opened ideally helps the author discover a root cause and fix faster.


> And the "code of conduct" tick box is not only pointless, it could itself be seen as offensive or disrespectful.


This was actually lifted from the current Issue template in the main Cinnamon repository ([link](https://github.com/linuxmint/cinnamon/blob/master/.github/ISSUE_TEMPLATE/bug_report.yaml)). Unfortunately, sometimes there needs to be a reminder for cordiality. Many other projects have also adopted a Code of Conduct as well. Feedback is welcome for how this can be presented so as not to be off-putting.

rcalixte:
> A drop down list would be the perfect solution but there seems to be a problem atm because, in applets at least, it's too long to fit on the page so half the applets are missing

I see where the disconnect is now. The preview isn't the same as the actual form itself.

This is what the full dropdown looks like in the actual form:

anaximeno:
@rcalixte if someone wants to fill a feature request, will they be able to fill it here, or will this template be mandatory when filling an issue on the spices repos? Or will there be a redirection to the discussion section like in the cinnamon and other repos?

rcalixte:
> @rcalixte if someone wants to fill a feature request, will they be able to fill it here, or will this template be mandatory when filling an issue on the spices repos? Or will there be a redirection to the discussion section like in the cinnamon and other repos?


With the changes proposed here, this is what the page that the "New Issue" button generates will look like:


It's also worth mentioning that the form has no validation so hopefully it won't be too imposing. (i.e. Every required field can be "N/A" if needed.) This is all fluid though. Nothing is set in stone so feedback is appreciated.

The "Feature Request" button links to the Discussions. As of today, many users also use the Spices website for feature requests and bug reports.

anaximeno:
I just took a look at this and it seems good.

fredcw:
@rcalixte My concern is that people may see all the required fields and decide that it's too much effort to fill it all out and decide not to bother. But I'd rather get a bad bug report than no bug report at all. I think the xlets bug report form can afford to be far less formal than on other mint repos because with the other repos there are few developers available to deal with a lot of bugs but with the xlets repos, I think it's more often the case that a single developer only occasionally has to deal with a bug and therefore has far more time to ask for further information if needed.

anaximeno:
@fredcw Even though I agree it's better to have a bad report than not having one, I don't believe the current template (at least the one for the applets) would dissuade users from creating a report since the only required fields are the cinnamon version, distribution, applet name, and description of what happened, they are all essential in my view for a report of a xlet, while easy to obtain for the users (you probably already know the applet name, and what happened, also it's very likely you know what's your distribution, and it's straightforward to obtain the cinnamon version if you don't know). 

I also find this bug report template far easier than the ones in the main projects (which I believe can be a bit dissuasive).

anaximeno:
I've been wondering though, if it isn't better to let feature requests be made here than in the discussion section

rcalixte:
> I've been wondering though, if it isn't better to let feature requests be made here than in the discussion section

#5056 is open for feedback.

rcalixte:
The changes based on the feedback here have been merged for this repository. If there is no further feedback, we can merge the others as well.

fredcw:
@rcalixte  You ignored the feedback.

rcalixte:
> @rcalixte You ignored the feedback.

Can you expand on this? Have you looked at the current form? There have been a few submissions for Issues with the new form as well if you want to take a look at those as well.

fredcw:
Yes, I've looked at the current form and I've seen the new submissions. Apart from getting rid of the tick box, you've ignored everything else I said. I don't know what else you want me to say. You're demanding of bug reporters (collaborators) that they waste their time filling out what is very often completely irrelevant information. Open source is about collaboration, not making demands of people.

rcalixte:
> Open source is about collaboration, not making demands of people.

It's also about community isn't it? Everyone has a role in that community and providing actionable bug reports or feature requests that make it easy for developers to take action only make collaboration easier for everyone right? This is the same as when there are beta releases for Linux Mint and great bug reporters help issues get fixed very quickly.




> Apart from getting rid of the tick box, you've ignored everything else I said.

I'm sorry you feel like this. We've gone through many iterations just in this thread based on your feedback and I felt like we'd come to a reasonable compromise. Your concern is that the template is too difficult for members. Is this feedback that you have seen elsewhere that can be shared? Again, this is just a spin from what is already on the main Cinnamon repository as well. The goal isn't friction but to make the resulting contribution easier to process.

fredcw:
I didn't come here for an argument. If you're going to ignore what I say, then that's you're choice but I won't be telling anyone to file to bug report on the applets repo when the form makes unreasonable demands and wastes people's time.

anaximeno:
I believe we are exaggerating a bit in some points here, people who have bugs want them to be fixed so most of them would not care to make a bit of effort for that, in open-source software most of the time instead of putting telemetry to monitor applications executing in the users' devices, the users have to come to where their software is developed and fill an issue for the developers to fix the bug since the users are the one filling the issues they won't always know what is more essential in that case. As I noted above at least the version of cinnamon, and the distribution are good information because some bugs are distro-specific (I've run into a similar problem with an extension), and others are specific to a version of cinnamon (I've had similar problems too). Even though not all applets receive issues all the time as you stated above, we must not forget that the developers are contributors like people who file these issues here and it will be an extra effort if they always have to make these questions before having to fix the issue (which also takes time in some cases). In community (and open source) projects, we all should be open to discussing things and understanding each other's points of view, even when not agreeing with them.

PR #5046, opened 2023-10-13
Add Korean translation

chaeya:
Add ko.po file and modify text variable to add Korean translation

- Changed text variable value in applet.js
- Added ko.po and pot file

### DIFF
--- a/sync@szlldm/files/sync@szlldm/applet.js
+++ b/sync@szlldm/files/sync@szlldm/applet.js
@@ -63,7 +63,7 @@ MyApplet.prototype = {
 	dirty = dirty.substring(7).trim();
 	writeback = writeback.substring(11).trim();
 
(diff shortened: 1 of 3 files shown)


rcalixte:
@chaeya The validation failed with the following message below:
```txt
[sync@szlldm] Error during validation: [sync@szlldm] The following errors were found in translation file 'ko.po':
msgfmt: files/sync@szlldm/po/ko.po: warning: PO file header fuzzy
                                    warning: older versions of msgfmt will give an error on this
HINT: Most errors in translation file `ko.po` can usually be prevented and solved by using Poedit.


Errors were found! Once you fix the issue, please re-run 'validate-spice <uuid>'' to check for further errors.
```

You can install and use Poedit to prevent these problems.

PR #5049, opened 2023-10-21
[smallcalc_applet@lerc.sds] Allows the user to define the precision for the result

claudiux:
@lerc
See https://github.com/orgs/linuxmint/discussions/193

### DIFF
--- a/smallcalc_applet@lerc.sds/files/smallcalc_applet@lerc.sds/applet.js
+++ b/smallcalc_applet@lerc.sds/files/smallcalc_applet@lerc.sds/applet.js
@@ -46,7 +46,8 @@ function makeApplet(orientation, panel_height, applet_id) {
     var base = 10;
 
     var settings = new Settings.AppletSettings(applet, uuid, applet.instance_id);
-    settings.bindProperty(Settings.BindingDirection.IN, "definitions", "definitions", definitionsChanged);
+    settings.bind("definitions", "definitions", definitionsChanged);
+    settings.bind("precision", "precision", null);
 
     let parser = MathJS.parser();
     loadDefinitions();
@@ -110,7 +111,7 @@ function makeApplet(orientation, panel_height, applet_id) {
             dragging = true;
             drag_device = event.get_device();
             drag_device.grab(osdBox);
-            
+
             Main.pushModal(osdBox);
 
             dragStartMousePosition = event.get_coords();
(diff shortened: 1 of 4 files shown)


fredcw:
@claudiux This sets the maximum number of decimal places rather than the precision. For example, if you set the precision to 10 and then do 1.3*3*100000000 you'll still get 3900000000.0000005.

The math.js library that this applet uses apparently has a precision function [Round-off errors](https://mathjs.org/docs/datatypes/numbers.html#roundoff-errors) Maybe just round to a precision of 14 as suggested in this reference?

fredcw:
This seems to work:
```

if (typeof(result) === 'number') {
    result = MathJS.format(result, {precision: 14});
}
```

claudiux:
@fredcw 
Thank you! Can you test the most recent changes, please?

fredcw:
@claudiux It seems to work fine. A couple of suggestions though: given that a precision of 16 (and possibly 15 also) may mean that the last 1 or 2 digits may be inaccurate, it may be misleading to be able to set the precision to these values when they may in fact not be precise. Perhaps the maximum (and default value) should be 14? and I guess that should be plenty for a simple calculator. Also, the "As is" seems to effectively be the same as 16 so that seems unnecessary.

I like the green equals button but I think the button active colour should be a lighter version of the green, same as the Clear button is a lighter version of red. Maybe `background-color: rgba(29, 212, 94, 0.6);`

claudiux:
@fredcw Can you test these new changes, please?

fredcw:
Seems to work OK but I meant the green highlight colour when the button is pressed. Something like this would be ideal I think:
```
.button.equals {
    background-color: rgba(29, 212, 94, 0.6);
}

.button.equals:active {
    background-color: rgba(29, 212, 94, 0.8);
}
```

The colouring of the close button is a nice touch too.

Precision option of 0 seems the same as 16 as well so I guess that one isn't necessary. You still have the 15 and 16 options though.

claudiux:
I've removed the precision 0 option and now use a scale of 1 to 16 (not 1). Ok for the css, I'll modify it immediately.

fredcw:
@claudiux Nice. Looks good.

Lerc:
I should take another look at the calculator,  I kind of lost track of it  when it moved to this repo.

claudiux:
Many thanks to @fredcw! Without him, this applet wouldn't be so perfect.

PR #5056, opened 2023-10-23
Create Feature Request form/template

rcalixte:
Similar to linuxmint/cinnamon-spices-desklets#935
Similar to linuxmint/cinnamon-spices-extensions#488
Similar to linuxmint/cinnamon-spices-themes#761


anaximeno:
I was a bit busy this week, but it seems OK to me.

PR #5058, opened 2023-10-24
[moonphase@techi-freki] Adds percentage display to label and tooltip.

sphh:
Display of phase name and/or phase percentage can be independently configured for the label and the tooltip.

### DIFF
--- a/moonphase@techi-freki/files/moonphase@techi-freki/js/moon.js
+++ b/moonphase@techi-freki/files/moonphase@techi-freki/js/moon.js
@@ -7,11 +7,12 @@ const Gettext = imports.gettext;
 Gettext.bindtextdomain(UUID, get_home_dir() + "/.local/share/locale")
 
 class Moon {
-    constructor(currentDate, useAltIcons) {
+    constructor(currentDate, useAltIcons, showNameLabel, showPercentageLabel, showNameTooltip, showPercentageTooltip) {
         this.currentDate = currentDate;
         this.age = this._getAge();
         this.currentPhaseIcon = this._getCurrentPhaseIcon(useAltIcons);
-        this.currentPhaseName = this._getCurrentPhaseName();
+        this.currentPhaseName = this._getCurrentPhaseName(showNameLabel, showPercentageLabel);
+        this.currentTooltip = this._getCurrentPhaseName(showNameTooltip, showPercentageTooltip);
     }
     // translation
     _(str) {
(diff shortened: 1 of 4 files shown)


sphh:
@Techi-Freki

sphh:
We could also add the number of days to the next full/new moon. Would that be appreciated? (But thinking this even further, this would call for a format string, complicating it for the average user … What do you think?)

claudiux:
It works fine. I was thinking of having the date of (or number of days before) the next Moon phase, displayed in the tooltip.

claudiux:
@sphh Please change the version number in `metadata.json`.

sphh:
I can change the version number, no problem, but I wanted to leave this to @Techi-Freki. Firstly, maybe s/he wants to make some changes, add a new feature or has a certain idea about the version number. I would have changed it to 0.1.0.

I also think, adding more display options calls for a format string. Let's wait a couple of days and see, what @Techi-Freki thinks about it.

claudiux:
Ping @Techi-Freki

Techi-Freki:
Hey guys, this all looks good to me. I'm surprised this thing is getting as much engagement as it is. Thanks! I appreciate it.

rcalixte (approved):

sphh:
@Techi-Freki: Glad that you like it.

What is your opinion about adding other information as mentioned in https://github.com/linuxmint/cinnamon-spices-applets/pull/5058#issuecomment-1777653682 ff.?

Techi-Freki:
I'm ok with it. The only other thing I was looking to implement was an indicator that displays a lunar eclipse and maybe moon set and rise times. But I might need to gather more info from the user (geolocation) which I'm not really wanting to do at this point because I have a strong privacy streak in me, that might change soon.

I'm really not sure how to properly display those other than showing them in the settings section because I think showing it in the cinnamon task bar may be a bit much. I'm open to suggesting on the front end display of it (I'm a back end dev by trade, so front end and design isn't my strong suite).

I already have those icons added as part of the icon set class. As far as the versioning is concerned, this is my first open source project that I've released and the v0.0.1 was a bit of a mistake on my part. It should have been 1.0.0 but we can change it as we deem necessary. Thanks again for both of your additions. This has been pretty exciting. I'm looking to get some Spanish and maybe Italian and Japanese translations added.

claudiux:
Hi @Techi-Freki 

Displaying lunar eclipse and moon set and rise times: very good idea!
For geolocation, simply add a switch in the settings, in a "Privacy" section. Depending on the value of this switch, the user may be offered the display of the lunar eclipse and the moonrise and sunset times. A replica of this switch can be displayed in the applet's contextual menu, or in a (non-contextual) menu that would then have to be created.


To know the status of translations for your applet:
https://github.com/linuxmint/cinnamon-spices-applets/blob/translation-status-tables/.translation-tables/tables/moonphase%40techi-freki.md 

Ping @Dragone2 @haggen88 for Italian and Spanish translations.

PR #5059, opened 2023-10-24
IntelTurboBoost@wsmenezes: Intel Turbo Boost spice. Pull request for initial release.

wsmenezes:
Simple systray applet to enable / disable Intel Turbo Boost.


IntelTurboBoost@wsmenezes/README.md
```diff
@@ -0,0 +1,128 @@
...
+~~~
+
+On **Redhat/Fedora** systems, install them with:
+
+~~~bash
+$ yum install msr-tools
+$ yum install python3-polib
+~~~
+    
+## Installation / Removal
```
rcalixte, line 84: Just an FYI, this is handled once it is added to the repository of Spices and available from the Applets dialog in Cinnamon.

rcalixte (approved):

IntelTurboBoost@wsmenezes/README.md
wsmenezes: Yes, thanks for catching that. Do you think it would be better to remove that to avoid any confusion? Please let me know your thoughts. Thanks again for the review.

rcalixte: It's your call! As long as users aren't confused. I had to check myself that the applet didn't require manual compilation as that wouldn't be the best user experience. Maybe borrow from the README of other applets and encourage users to install via their system if possible, then provide alternatives if something goes wrong?

rcalixte: If you look on your own system, you should see this applet in the system Applets dialog and you can make sure it's working as intended.

wsmenezes: > It's your call! As long as users aren't confused. I had to check myself that the applet didn't require manual compilation as that wouldn't be the best user experience. Maybe borrow from the README of other applets and encourage users to install via their system if possible, then provide alternatives if something goes wrong?

Got it. I'll do that, replicate some section like you suggested. It's better. I'll request an update.

Issue #5065, opened 2023-10-26
hwmonitor@sylfurd Feature Request: option to hide some graphics


Cyrille37:
### Applet name and maintainer

hwmonitor@sylfurd @hultan

### What would you like to see?

Hi

Thanks a lot for this great applet, it's nice.

Do you think possible to add settings to hide (or show) some graphics. For example I would like to not display the "BAT" and "MEM" graphics.

Thanks & Cheers :-)

fredcw:
@Cyrille37 Due to a bug, the page tab titles in the config dialog are not visible when first opened. Widen the config window to access the tabbed pages and turn off the graphs you don't want.

Cyrille37:
Yeah! Thanks :-)

But it's not the first time, but the with which shows tabs.

![image](https://github.com/linuxmint/cinnamon-spices-applets/assets/368965/531e3d59-6cb1-47e2-95a4-e3db9eb83600)

![image](https://github.com/linuxmint/cinnamon-spices-applets/assets/368965/5d659aca-44e4-4017-beb7-6d66b6fb2498)

Hultan:
> @Cyrille37 Due to a bug, the page tab titles in the config dialog are not visible when first opened. Widen the config window to access the tabbed pages and turn off the graphs you don't want.

Oh, so that is why the tabs are sometimes not visible? I have never seen this happen on my machine, so I have given up on solving this bug. Thanks for fixing this bug @fredcw! I hope someone accepts the PR soon, I don't have permission to do that.

BTW, feel free to take over this applet by changing the `author` field in `info.json` file. I doubt that I will do more work on this applet.

Cyrille37:
Thanks guys

I made the update, it's better :-) but it's not exhaustive :-(

Tabs are now visible with a smaller window
![image](https://github.com/linuxmint/cinnamon-spices-applets/assets/368965/cdcd1774-cc11-4148-a471-2927bbc293e8)

But with an even smaller window they disappear again
![image](https://github.com/linuxmint/cinnamon-spices-applets/assets/368965/36f31566-b13c-4c30-9cdb-aebe04aa4de2)

Thanks & cheers :-)

fredcw:
[...]

Issue #4439, opened 2022-08-23
hwmonitor@sylfurd : Config pages are not visible


cgvirus:
```
 * Applet version Version 1.3.1 
 * Cinnamon version 5.4.11
 * Linux Mint 21 Cinnamon
 ```

**Notify author of applet**
hwmonitor@sylfurd

**Issue**

- The Configure pages are not visible.
- Only showing general settings, theme (graph) and theme graph data, The first page

**Steps to reproduce**

- right click on the applet>configure.

**Expected behaviour**

- Should show all configuring pages

fredcw:
Ping @Hultan 

@cgvirus It's not a fix but as a temporary measure, making the configuration dialog wider causes the tabs to appear.

cgvirus:
> Ping @Hultan
> 
> @cgvirus It's not a fix but as a temporary measure, making the configuration dialog wider causes the tabs to appear.

OH! And here I am going through js for finding whats up! XD 
Thanks!

Drugwash2:
Could anyone please try the suggestion I posted **[here](https://forums.linuxmint.com/viewtopic.php?f=120&t=375881)** and confirm whether it works or not with the latest Cinnamon (and with older versions too)?
If it does maybe the Mint developers would implement it natively since as is any further update to Cinnamon would instantly revert the change.

Issue #3818, opened 2021-06-22
[hwmonitor@sylfurd]: Settings window too small


jkirk:
**Facts**

 * Applet version/Build date (to be found in "About.." or in "metadata.json")
 1.3.1 -> https://github.com/linuxmint/cinnamon-spices-applets/tree/009831231f8bd0a04e7f8c45f8b1ad2931e24a1c
 (**please update the version string.**)
 * Cinnamon version (cinnamon --version)
 Cinnamon 4.8.6
 (**please add support for Cinnamon 4.8.**)
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
 Debian GNU/Linux 11 (bullseye)

**Notify author of applet**
@hultan

**Issue**
The settings window is too small. The settings pages are not visible at first:

![Screenshot from 2021-06-22 11-07-27](https://user-images.githubusercontent.com/288637/122897178-0c293580-d34a-11eb-8202-13e81b4a9ad6.png)

Only after expanding the width, the settings pages get visible:

![Screenshot from 2021-06-22 11-14-11](https://user-images.githubusercontent.com/288637/122898201-fb2cf400-d34a-11eb-84c6-d54e3d90e7d9.png)

**Steps to reproduce**

Right click hwmonitor-applet -> click configure

**Expected behaviour**

The settings window should be large enough to make all settings pages visible by default.

Thanks for maintaining the applet and looking into this. 🤗

Hultan:
To my knowledge, hwmonitor has no control over the settings window size. I believe this is handled by Cinnamon. Is it still a problem for you? I guess that, since it had been a month, you have rebooted your machine, otherwise try that.

PR #5067, opened 2023-10-27
hwmonitor@sylfurd: Reduce number of pages in config dialog.

fredcw:
Reduce number of tabbed pages by combining the NET and DISK tabs so that they appear when config dialog is first opened.

If the tab bar is too wide for the window, it doesn't show until the window is widened.

fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/5065
fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/4439
fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/3818

### DIFF
--- a/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.8/settings-schema.json
+++ b/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.8/settings-schema.json
@@ -2,16 +2,15 @@
     "layout": {
       "type": "layout",
       "pages": [
-        "General",
-        "CPU",
-        "MEM",
-        "NET_IN",
-        "NET_OUT",
-        "DISK_READ",
-        "DISK_WRITE",
-        "BAT"
+        "General_page",
+        "CPU_page",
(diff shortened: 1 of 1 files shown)


rcalixte:
@claudiux This should have waited on approval from @Hultan.

fredcw:
@rcalixte I didn't mention the author because I took this comment https://github.com/linuxmint/cinnamon-spices-applets/issues/5065#issuecomment-1782549158 as implied approval.

rcalixte:
Cool. I hadn't seen that. Are you willing to take this over? 😅

We can set the author to `None` if not so that future contributions aren't blocked.

fredcw:
Yes, I think none is probably best as I'm not too familiar with the applet and don't have any ideas for it that I'd like to implement.

Hultan:
I really like this change, this is how I should have done it from the beginning. But I did not now about the way the number of tabs could affect the settings dialog functionality when I added all those tabs. Great job @fredcw!

PR #5090, opened 2023-11-01
[MuteToggler@jebeaudet.com] Automatically decides which 'amixer' parameter to use (Closes #3979)

### LINKED ISSUE #3979: [MuteToggler@jebeaudet.com] Not working on every installation
```
 * Applet version/Build date: 2021-07-16 09:28:06
 * Cinnamon version: 5.0.5
 * Distribution: Mint 20.2
 * Graphics hardware *and* driver used
 * Snippet or attachment of ~/.cinnamon/glass.log: N/A
 * Snippet or attachment of ~/.xsession-errors: No entries

 ```

**Notify author of applet**
@jebeaudet 

**Issue**
I have `MuteToggler` installed on two computers, both with Mint 20.2. It works on one, but not on the other. I could track it down to the problem, that `Capture` does not exist on one computer:

1st computer:
```sh
$ amixer scontrols
Simple mixer control 'Master',0
Simple mixer control 'Headphone',0
Simple mixer control 'Speaker',0
Simple mixer control 'PCM',0
Simple mixer control 'Mic',0
Simple mixer control 'Mic Boost',0
Simple mixer control 'IEC958',0
Simple mixer control 'IEC958',1
Simple mixer control 'IEC958',2
Simple mixer control 'IEC958',3
Simple mixer control 'IEC958',4
Simple mixer control 'Capture',0
Simple mixer control 'Auto-Mute Mode',0
Simple mixer control 'Digital',0
Simple mixer control 'Internal Mic Boost',0
Simple mixer control 'Loopback Mixing',0
```

2nd computer
```sh
$ amixer scontrols
Simple mixer control 'IEC958',0
Simple mixer control 'IEC958',1
Simple mixer control 'IEC958',2
Simple mixer control 'IEC958',3
Simple mixer control 'IEC958',4
```

Toggling the microphone with the help of the `Sound applet` works on both computers.

**Expected behaviour**

Show at least an error message, if toggling does not work.

**Other information**

Because toggling the mic with the `Sound applet` works, that left me wondering, if it is not easier (for the user) to implement the feature of `MuteToggler` into the `Sound applet`. Most of the features are already there: In the preferences dialog, it is possible to set the action of a middle click to `Toggle Mute input`. The only feature missing is to show it on the icon, e.g. by an emblem/badge with a green or a red, crossed out mic. (This would also save some space!)

What do you think?

sphh:
On some installations it is not enough to call 'amixer'. There an
additional parameter is needed: 'amixer -D pulse' for amixer to find the
proper channel. This PR checks, if it can find the 'Capture' channel
with 'amixer' and if not successful with 'amixer -D pulse'.

### DIFF
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -86,8 +86,7 @@ MyApplet.prototype = {
                 ];
             Util.spawn_async(cmd, (stdout) => {
                 try{
-                    var string = new TextDecoder().decode(stdout);
-                    if(string().indexOf("] [on]") != -1){
+                    if(stdout.indexOf("] [on]") != -1){
                         this.set_not_muted_icon();
                     } else {
                         this.set_muted_icon();
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -13,14 +13,14 @@ MyApplet.prototype = {
     
     _init: function(metadata, orientation, panel_height, instance_id) {
         try {
-            global.log("Initializing MuteToggler applet")
+            global.log("MuteToggler: Initializing applet")
             Applet.IconApplet.prototype._init.call(this, orientation, panel_height, instance_id);
             this.uuid = UUID;
             
             this.set_applet_tooltip(_("Click to mute/unmute microphone"));
 
             try {
-                global.log("Initializing MuteToggler Settings")
+                global.log("MuteToggler: Initializing settings")
                 this.settings = new Settings.AppletSettings(this, this.uuid, instance_id);
                 this.settings.bindProperty(Settings.BindingDirection.IN,
                                      "keybinding",
@@ -42,7 +42,7 @@ MyApplet.prototype = {
             this.is_audio_muted();
             
             this.refresh_loop();
-            global.log("Done initializing MuteToggler applet")
+            global.log("MuteToggler: Done initializing applet")
         } catch (e) {
             global.logError(e);
         }
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -62,7 +62,7 @@ MyApplet.prototype = {
     is_audio_muted: function() {
         try {
             let cmd = [
-                "bash",
+                "sh",
                 "-c",
                 "amixer sget Capture"
                 ];
@@ -95,7 +95,7 @@ MyApplet.prototype = {
     on_applet_clicked: function(event) {
         try{
             let cmd = [
-                "bash",
+                "sh",
                 "-c",
                 "amixer set Capture toggle"
             ];
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -4,6 +4,7 @@ const Main = imports.ui.main;
 const Mainloop = imports.mainloop;
 const Util = imports.misc.util;
 const Settings = imports.ui.settings;
+const GLib = imports.gi.GLib;
 
 const UUID = 'MuteToggler@jebeaudet.com';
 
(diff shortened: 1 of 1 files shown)


rcalixte:
cc @jebeaudet

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
```diff
@@ -37,12 +37,29 @@ MyApplet.prototype = {
...
+                let [res, stdout] = GLib.spawn_command_line_sync(cmd);
+                var string = new TextDecoder().decode(stdout);
+                if (res && string.indexOf("'Capture'") != -1) {
+                    this.amixer = "amixer" + param;
+                    global.log("MuteToggler: Use mixer command '" + this.amixer + "'"
+                    break;
+                } 
+            }
+            if (this.amixer == "") {
+                throw "No working mixer command found!";
```
jebeaudet, line 55: exceptions are catched lower and we'll end up with an invalid command "", could you log an error and fallback on the default `amixer` command at least?

jebeaudet (changes requested):
Thanks for the PR!

This unfortunately won't work with a hotplugged mic which would switch from alsa to pulseaudio but it's better than nothing without adding some mic detection logic in there.

I've asked another user who had trouble that reached me by email to try and test your PR out, I'd be curious to see if it works for them.

Until then, if you could just add a default amixer command in case your logic didn't find a capture then I think we could merge it.

Thanks again, I'm trying to see if I can get a linux mint vm up and running to try and help you.

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
sphh: I see. I have a built in microphone, so never thought about this.

What do you think about not using this logic during start-up, but whenever the microphone is toggled? That should take care of hot-plugged microphones.

sphh: Or maybe even easier: Use the command `sh -c "amixer set Capture toggle || amixer -D pulse set Capture toggle"`? This works, because if `amixer` cannot find the the device, its returned error code is >0.

sphh: ~Or maybe even easier: Use the command `sh -c "amixer set Capture toggle || amixer -D pulse set Capture toggle"`? This works, because if `amixer` cannot find the the device, its returned error code is >0.~
I just tried to implement this, but unfortunately I could not get it to work in the applet code (it works on the terminal).

sphh:
In my latest commit, it (silently) falls back to 'amixer', if no suitable parameter can be found.

The more I think about this, the more I believe, that this functionality should be part of the standard sound applet. There you can already configure the middle click to mute the microphone. Adding a hotkey for this feature should be rather easy. Unfortunately it does not show it on the icon …

jebeaudet (approved):
LGTM

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
sphh: IMHO there are three sides to this:
1. There is the configurable hotkey for muting the microphone in the system setting. This functionality is the same as that for the MuteToggler hotkeys.
2. The sound applet already have a middle-click functionality. It can be used to toggle the microphone (and some other things). So this functionality is already there (but with the middle click instead of clicking onto MuteToggler). I just added a shift-middle click, so that you can toggle play _and_  (separately) toggle the microphone.
3. The MuteToggler shows the current status of the microphone. This is not shown in the standard sound applet. This is the only missing feature so that the sound applet can take over all functionality of the MuteToggler.

The standard sound applet uses the standard icons. I could not find an icon family which shows the output and input state in one icon. I guess, an alternative approach is needed, such as overlaying a badge onto an existing icon. Do you happen to know, how this can be done?

jebeaudet: Glad to hear that it's now all built in in the sound applet, can't remember if it was at the time I did this. But yea, a mute indicator would be the best and could put the final nail in the coffin for this applet.

I unfortunately have no idea how to do a badge. Reverse engineering the sound applet should not be too bad if you want to try it, IIRC, the sound applet icon showed a red X when the sound output is muted no? You could try to find this behavior in the codebase and see how it's done to mimic the same thing for the current active mic input.

jebeaudet:
@rcalixte If you want to merge this, it lgtm.

rcalixte (approved):

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
sphh: I just opened a discussion (https://github.com/orgs/linuxmint/discussions/209).

With regard to the badges: [SpicesUpdate@claudiux](https://github.com/linuxmint/cinnamon-spices-applets/tree/master/SpicesUpdate@claudiux) uses text badges implemented as [Label](https://gjs-docs.gnome.org/st13~13-label/) on its icon. Maybe it's easy to replace the label with an [Icon](https://gjs-docs.gnome.org/st13~13-icon/)?

PR #5106, opened 2023-11-16
[MuteToggler@jebeaudet.com] Allow the use of symbolic icons

claudiux:
@jebeaudet 
This PR allows users to choose between solid and symbolic icons.


jebeaudet (approved):
If you're interested in taking over the applet, I'm open to it since I'm unfortunately not using linux mint as much as I want to anymore since changing job.

If you want to, just open another PR with the `author` and `original_author` and I'll approve it.

This being said, thanks again

claudiux:
It's not my intention, but I'm a fan of this applet and I've modified its code to adapt it to my use.

I promise I won't make any more modifications from now on. ;-)

You've done a great job. Thank you so much!

PR #5133, opened 2023-11-21
[timer@Severga] Preserves timer despite applet restart

claudiux:
@Severga 

Hi,
When mintupdate updates xlets in use, Cinnamon restarts and the timer is reset.

This PR preserves timer (and "turn off" status) despite the applet restart.

Are these changes to your liking?

Regards
Claudiux


Severga:
Hello, Claudiux!
I thought about this feature while writing the code. But I thought only about machine reboot, and decided not to implement it. I hadn't thought about the situation you said (applet reloading). I think in this situation this feature is really good and necessary. I'm for this feature.
Thank you for your contribution!

rcalixte (approved):

PR #5139, opened 2023-11-25
[MonitorInputSource@klangman] Initial applet commit

klangman:
A new applet that allows the user to change the input source for all the monitors attached to their computer.


MonitorInputSource@klangman/files/MonitorInputSource@klangman/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+   "uuid": "MonitorInputSource@klangman",
+   "name": "Monitor Input Source",
+   "description": "Change your monitors video input source directly from the cinnamon panel",
+   "max-instances": "1",
```
rcalixte, line 5: Would it make sense to enable more than one instance of the applet? I'm thinking in the case of multiple panels across multiple displays.

rcalixte (approved):

MonitorInputSource@klangman/files/MonitorInputSource@klangman/metadata.json
klangman: Yes.. I didn't think that through fully. I'll make that change. Thanks Rick.

rcalixte: It's no problem! Your contributions have been great! I was excited to try this out but it appears my monitors may not support ddcutil (maybe I need to dig further). Either way, thanks for this!

klangman: I have two older Lenovo monitors at work and I was surprised that ddcutil was able to send commands to them. I don't know a whole lot about DDC/CI but based on what I see in Wikipedia, the spec that allow for sending commands to monitors was introduced in 1998 but adoption was slow so it might be only in 2010 that monitor support was wide spread (but that's just me reading between the lines on the wki page). 

If you manage to get your monitors to work with ddcutil and find that there is some information I can add to the README.md or any code changes that would make it work in more cases please let me know.

PR #5148, opened 2023-11-26
Remove libmpv1 or libmpv2 dependency for LM and Debian

claudiux:
Fixes the bug reported by @Neonteepee on https://cinnamon-spices.linuxmint.com/applets/view/360.

### DIFF
--- a/Radio3.0@claudiux/files/Radio3.0@claudiux/CHANGELOG.md
+++ b/Radio3.0@claudiux/files/Radio3.0@claudiux/CHANGELOG.md
@@ -1,3 +1,6 @@
+### v1.10.1~20231126
+  * Remove libmpv1 or libmpv2 dependency for LM and Debian.
+
 ### v1.10.0~20231126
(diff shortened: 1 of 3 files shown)


cubetti2004:
Hi, yes please remove libmpv1 or libmpv2. It does not work on Linux MInt 21.2 and LMDE6. Further I tried to record the radio to .flac, but this does not work too. If I try to play with vlc or Audicisous, I got the error message "Error playing file:///home/user1/Music/Radio3.0/O%20Come%2C%20O%20Come%2C%20Emanuel%20%20By%3A%20Eugene%20Ormandy%3A%20Philadelphia%20Orchestra%20-%20WALMRadio.com_2023-11-26_05-14-08%20PM_2023-11-26_05-14-08%20PM.flac: The file could not be decoded. It may be invalid, corrupt, or in an unsupported format." If I change the file type from .flac to .mp3 it works. Can you please change the recording or converting to real .flac files please.

PR #5159, opened 2023-11-28
healthyeyes@ipolozov - Fixed a bug where the colour would change back immediately

GingerGigiCat:
Fixed a bug where the colour would change back immediately after clicking it if the time was the same. Also added some logging.

### DIFF
--- a/healthyeyes@ipolozov/files/healthyeyes@ipolozov/applet.js
+++ b/healthyeyes@ipolozov/files/healthyeyes@ipolozov/applet.js
@@ -5,6 +5,23 @@ const Main = imports.ui.main;
 const Mainloop = imports.mainloop;
 const UUID = 'healthyeyes@ipolozov';
 
+var myDate = new Date();
+var last_hour = myDate.getHours() - 1;
+var last_minute = myDate.getMinutes() - 10; // Takes away a few minutes in case the time is already at the time it should change
+
+function log(message, type = "debug") {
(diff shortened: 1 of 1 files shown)


rcalixte (approved):

healthyeyes@ipolozov/files/healthyeyes@ipolozov/applet.js
```diff
@@ -24,19 +41,30 @@ MyApplet.prototype = {
...
-        if (minute == 0 || minute == 30) {
+        
+        //log(hour);
+        //log(minute);
+        //log(last_hour);
+        //log(last_minute);
+	
+        // Checks if it should change, and makes sure that it isn't the same time as when it last changed
+        if ((minute == 0 || minute == 30) && (minute != last_minute || hour != last_hour)) {
+            log("change!");
```
rcalixte, line 56: I wasn't sure if this was intentional or inadvertent. If it shouldn't be there, we can push the fix quickly.

GingerGigiCat: That log was meant to be left in, but if it's bad to have regular outputs then it could be removed

rcalixte: It's up to you. Sometimes debug things slip past. 😅

PR #5199, opened 2023-12-08
Add Russian translation to BgRadio@spacy01

blogdron:


blogdron:
Im created one more commit, because i miss some words before. Other words non-translated because it radio station and TV channels names.

PR #5202, opened 2023-12-08
Reloading Cinnamon Menu to get correct dimentions

mindbeat:
Added a reload of the Cinnamon menu applet after we change the resolution / scaling. 
This is so we get the correct scaling of the menu.

### DIFF
--- a/rswitcher@nixdev.com/files/rswitcher@nixdev.com/rswitcher
+++ b/rswitcher@nixdev.com/files/rswitcher@nixdev.com/rswitcher
@@ -129,3 +129,6 @@ fi
 
 # Run the xrandr function to create the new settings.
 create_resolution
+
+# Reload start menu to get correct dimentions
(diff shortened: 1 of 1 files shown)


rswitcher@nixdev.com/files/rswitcher@nixdev.com/rswitcher
```diff
@@ -129,3 +129,6 @@ fi
 
 # Run the xrandr function to create the new settings.
 create_resolution
+
+# Reload start menu to get correct dimentions
+dbus-send --session --dest=org.Cinnamon.LookingGlass --type=method_call /org/Cinnamon/LookingGlass org.Cinnamon.LookingGlass.ReloadExtension string:'menu@cinnamon.org' string:'APPLET'
```
rcalixte, line 134: You can also use /org/Cinnamon and org.Cinnamon and the ReloadXlet method call. I think that is the proper scope but this looks like it would still work too.

rcalixte (approved):

rswitcher@nixdev.com/files/rswitcher@nixdev.com/rswitcher
mindbeat: Thanks for the feedback! I will check it out. :+1:

PR #5226, opened 2023-12-14
Update for Hungarian translation of the Cinnamon applets

bossbob88:
This is an update of the hu.po files of the cinnamon applets.


claudiux (approved):

claudiux:
Thank you @bossbob88. For Radio3.0, two messages have not yet been translated: https://github.com/linuxmint/cinnamon-spices-applets/blob/translation-status-tables/.translation-tables/po/Radio3.0@claudiux/_hu.po

Issue #5320, opened 2023-12-28
download-and-upload-speed@cardsurf: null value assigned to the text label


anaximeno:
### Applet version/Build date

download-and-upload-speed@cardsurf 1.6.5 (2023-06-04 16:03:13)

### Cinnamon version

Cinnamon 5.8.4

### Distribution

LMDE 6

### Applet name and maintainer

download-and-upload-speed@cardsurf @cardsurf

### What happened?

![Captura de ecrã de 2023-12-08 17-21-13](https://github.com/linuxmint/cinnamon-spices-applets/assets/63073056/0ca03b9f-5b13-4095-813b-8a51505ecae7)

### Other information

I believe this was also causing some residual text to appear while updating the upload and download speed values, because it didn't happen anymore after the fix.

PR #5321, opened 2023-12-28
 download-and-upload-speed@cardsurf: fix null text label value

anaximeno:
Solves https://github.com/linuxmint/cinnamon-spices-applets/issues/5320


rcalixte:
cc @cardsurf Merging since this looks like a useful bug fix.

rcalixte (approved):

PR #5344, opened 2024-01-01
[qredshift@quintao] v1.7.6 - Adds option 'Activate as soon as Cinnamon starts up'

claudiux:
@raphaelquintao 
This PR adds an option "Activate QRedshift as soon as Cinnamon starts up".
This way, QRedshift can be activated even if it was previously deactivated.


haggen88:
If you want you can add the updated es.po file:  [es.po.zip](https://github.com/linuxmint/cinnamon-spices-applets/files/13802575/es.po.zip)

claudiux:
@haggen88 Done! Thank you for all your Spanish translations!

PR #5366, opened 2024-01-10
[system-controls@rcalixte] v1.3.0

claudiux:
@rcalixte

Hi Rick,

I offer some changes for your system-controls@rcalixte applet:

* Remove options that are unavailable when locks (disable-log-out, disable-switching-user, disable-lock-screen) are enabled, without having to restart the applet.
* Add OSD when restarting Cinnamon.
* Close the menu before activating any action.

I hope you'll find it to your liking.

Regards
claudiux

### DIFF
--- a/system-controls@rcalixte/CHANGELOG.md
+++ b/system-controls@rcalixte/CHANGELOG.md
@@ -1,4 +1,9 @@
 
+### 1.3.0
+* Remove options that are unavailable when locks (disable-log-out, disable-switching-user, disable-lock-screen) are enabled, without having to restart the applet.
+* Add OSD when restarting Cinnamon.
+* Close the menu before activating any action.
+
 ### 1.2.2
 
 * Remove unavailable options in the Wayland session for now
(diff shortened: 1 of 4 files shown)


rcalixte:
> * Remove options that are unavailable when locks (disable-log-out, disable-switching-user, disable-lock-screen) are enabled, without having to restart the applet.

I wasn't aware that this was an issue but if it is, I would prefer to implement it differently. Could you open an Issue with the details?




> * Add OSD when restarting Cinnamon.

I'd opted to keep this disabled to keep the applet simple. If it was going to be enabled, I would prefer an option but that would take away from the simple nature of the applet and I would prefer to keep it simple.




> * Close the menu before activating any action.

Similar to the first point, this wasn't a known problem and I haven't seen it in the course of regular usage. Can you also open an Issue with details of what you're experiencing?




In the future, please feel free to reach out to discuss particulars so that efforts aren't spent in vain.

PR #5369, opened 2024-01-11
[PanelTranslator@klangman] Initial Commit

klangman:
This is an applet that uses google via translate-shell to translate text


rcalixte (approved):

PanelTranslator@klangman/files/PanelTranslator@klangman/languages_0_9_6_12.js
```diff
@@ -0,0 +1,130 @@
...
+   {code:"gu",      englishName:"Gujarati"},
+   {code:"pap",     englishName:"Papiamento"},
+   {code:"yua",     englishName:"Yucatec Maya"},
+   {code:"ht",      englishName:"Haitian Creole"},
+   {code:"ps",      englishName:"Pashto"},
+   {code:"zu",      englishName:"Zulu"},
+   {code:"ha",      englishName:"Hausa"},
+   {code:"fa",      englishName:"Persian"},
+   {code:"haw",     englishName:"Hawaiian"},
+   {code:"pl",      englishName:"Polish"}
```
rcalixte, line 127: You could change the `englishName` key to just `name` and then use translatable strings, e.g. `name: _('French')`. Translations will be very long but it would solve the issue.

rcalixte: > If a newer translate-shell is install (newer then currently available in the default Mint repositories) then the English and non-English names are read directly from translate-shell and this hard-coded list is not used.

So this is essentially a short-term problem. The next version of Mint/Cinnamon will have 0.9.7 which is the same as LMDE. Could you branch a new version of this  in a 6.2 folder since everywhere will have that newer version of translate-shell at that time?

klangman: Actually.. Do you happen to know of a GUI widget I can use that support St.Scrollable to act as a "text box" for the from/to test entries?  I tried to use St.ScrollView with an St.Entry as a child but it didn't work, just got scroll bars but no entry field.

rcalixte: Maybe different params for the St.BoxLayout? Maybe @mtwebster has a better idea. I don't think I've tried to use the St class for something like that before.

PR #5377, opened 2024-01-13
directory-menu@torchipeppo: Initial release

torchipeppo:
This applet allows the user to open a directory of their choice in a dropdown menu directly, without passing through the file manager. It is intended to replicate the functionality of a built-in plugin for XFCE that works in the same way. [1]

I made this applet a few years ago because I missed its functionality when switching from an old machine with Xubuntu to a Mint/Cinnamon setup, and I have been using it ever since with no problems (on Cinnamon 5.0.5), so I believe it should be ready for submission here.

I do have one gripe with the implementation, for which I would ask the reviewers' opinion: I had to use the `popup` method of the GTK `Menu`, which the documentation marks as deprecated. [2]
This was deemed necessary because, despite my research efforts, it was the only way I could find to have the dropdown menu show up despite not having an underlying GTK application. In particular, the alternatives suggested in the documentation are not suitable because they do require an application. And I could find no workarounds to somehow make a Cinnamon applet "count" as a GTK application, either.
As for why I couldn't use Cinnamon's built-in submenu, the answer is that they and GTK menus have different look-and-feels and functionality: a Cinnamon submenu needs to be clicked through and only expands vertically, while a GTK submenu appears and disappears automatically and opens to the side, which is more practical IMHO and is less likely to run out of screen space when opening multiple of them as the user navigates through subfolders.
Using the GTK menu also makes this applet much closer to the original XFCE plugin.

Therefore, if the team feels that this applet using a deprecated GTK functionality is not that much of a problem, then it's fine; otherwise, any advice for a "cleaner" approach would of course be appreciated.

[1] https://gitlab.com/xfce/xfce4-panel/-/tree/master/plugins/directorymenu
[2] https://docs.gtk.org/gtk3/method.Menu.popup.html


rcalixte (approved):

rcalixte:
> Therefore, if the team feels that this applet using a deprecated GTK functionality is not that much of a problem, then it's fine; otherwise, any advice for a "cleaner" approach would of course be appreciated.

I can take a look in a little bit to see if I can get one of the supported methods working. If you want, you can also put the code and related pieces in a 5.0 directory (similar to how other applets use versioning if you take a look through) and it will be available for everything Cinnamon 5.0+. This looks awesome though. Thanks for your contribution!

torchipeppo:
Well, thank you for merging!
My worry with using `popup` is more that it's not future-proof in case Cinnamon ever switches to a future GTK version where `popup` is removed for good, rather than it not being compatible with previous versions: so setting up a 5.0 directory is an idea, but not necessarily something to do urgently, assuming I understood your advice correctly.
And if you do end up dedicating some of your time to this and find a solution, I'd be grateful if you let me know.

rcalixte:
I have an idea. Have you tried using PopupMenu instead of Gtk.Menu? You would need to re-implement the `populate_menu_with_directory` function to use the new object structure but I think it would work better. Take a look at existing applets for an example. I'll see if I can find a link to relevant documentation too.

fredcw:
I just tried this applet. Nothing happens when I click on the applet and nothing reported in ~/.xsession-errors. Tried click tolerances between 20 and 300ms

OS Manjaro, Cinnamon 6.0.4

torchipeppo:
@fredcw Thanks for the report! That indeed seems to be the case, I just tested on an empty VM running 21.2 live. I'll look into it in my spare time, but I did find a small weirdness in these few minutes. I got the menu to show up by performing these steps:

* Open the Cinnamon debugger (Super + L by default)
* Click on the dropper button on the bottom left
* Click on the applet

You may see the menu items replicated multiple times the first time you do this: this is normal, as the menu is populated every time you click the applet, but items are disposed of only when the menu disappears, which clearly isn't happening normally as the menu doesn't even appear when clicking on the applet under normal circumstances.

I don't know what to make of this weird finding yet, and I don't know how long the investigation is going to be, but I'd be glad if you could at least try to reproduce these steps and see if they have the same effect for you.

@rcalixte Normally I would agree, but as I've already discussed in my opening message, don't Cinnamon's sub-PopupMenus expand vertically on top of the parent menu, rather than appearing off to the side like GTK menus? (The "Launch player" item of the volume applet's menu should be an example of such a Cinnamon submenu.) I fear this would consume screen space much faster. Plus, having to click to open each submenu feels impractical to me, but that might just be due to personal taste or my usage of the applet.

rcalixte:
> @rcalixte Normally I would agree, but as I've already discussed in my opening message, don't Cinnamon's sub-PopupMenus expand vertically on top of the parent menu, rather than appearing off to the side like GTK menus? (The "Launch player" item of the volume applet's menu should be an example of such a Cinnamon submenu.) I fear this would consume screen space much faster. Plus, having to click to open each submenu feels impractical to me, but that might just be due to personal taste or my usage of the applet.

Take a look at [popupMenu.js](https://github.com/linuxmint/cinnamon/blob/master/js/ui/popupMenu.js#L2186C1-L2704) as there are methods to open the menu on either side when opened.

fredcw:
@rcalixte I think what @torchipeppo is referring to is the separate sub menu that opens next to the menu that's already open when you hover over a menu item. As far as I know, this is not possible with cinnamon/St, only one menu can be open at any one time in cinnamon.
spg5

rcalixte:
I'm aware. I'm referring to the code that I believes support this functionality. I'm checking to see if there are any applets making use of this already.

fredcw:
@rcalixte Yes, class PopupMenu is opened on top/bottom/left/right of a widget but only one can be open at any one time on the desktop and it's normally the applet itself.

rcalixte:
The code in the desklet devTools@scollins seems promising but I would need to spend time with it. If this functionality turns out not to be possible, maybe we should just remove this applet until a solution can be implemented? It seems like a nice idea.

torchipeppo:
> maybe we should just remove this applet until a solution can be implemented?

I do agree that right now the applet is pretty broken on the newer versions, where I'd assume most users are. Maybe we could leave this up for a little while, just in case the solution turns out to be quick or we get some extra input, and then remove the applet if the situation is unchanged?

> The code in the desklet devTools@scollins seems promising

I actually did use this desklet for unrelated reasons a while ago! I'll check it out myself.

Anyway, I'm going to install a Mint 21.3 VM, since @fredcw reported the issue there too, and move development there. I do have full interest in finding a solution that works on the newer versions, since I will inevitably have to upgrade my system sooner or later.

rcalixte:
So good news and bad news.

The good news: from what I can see, it should be possible to do the submenus using GTK. I can see working implementations from programs like Cairo-Dock and like you mentioned Xfce as well. Both are GTK-based.

The bad news, both Cairo-Dock and Xfce are written in C so you would need to reverse engineer the GTK introspections to be able to implement them in JavaScript for an applet. That is also assuming the full library calls are available for Cinnamon's JavaScript engine already. This is likely going to take some time and effort to accomplish. Have we thought about how we would like to proceed here?

fredcw:
@rcalixte I think it should be removed until it can be fixed. There doesn't seem to be any point keeping it there while it doesn't work.

rcalixte:
I thought it worked for older versions. We could just render it ineffective for Cinnamon versions that aren't supported?

fredcw:
Actually, on older versions, (I tested Mint 20.3) if you right on the panel to open the panel context menu and then click on the applet, the applet window opens (so that both menus are simultaneously open) which causes cinnamon to freeze. (nothing in .xsession-errors)
Screenshot from 2024-01-18 02-20-18

rcalixte:
I stand corrected. We'll remove it now.

@torchipeppo Feel free to reopen and resubmit once you've got things working.

torchipeppo:
Got it. @fredcw 's double-menu freeze is certainly unfortunate, I just reproduced it on my system.
I also agree that it doesn't make sense to keep the applet as it is now.

@rcalixte
> both Cairo-Dock and Xfce are written in C so you would need to reverse engineer the GTK introspections to be able to implement them in JavaScript for an applet

I have been doing something similar, so far I believe I managed to get the non-deprecated popup_at_rect to work by translating to GJS some of the steps that the XApp Status Icons do in C (no source link right now b/c I'm not on the VM where I have it saved, but it's in a linuxmint repo). It's little progress, because the menu still doesn't appear except through Melange as before, but progress nonetheless.

That said, I think I'll return once I have something ready for a second submission. My sincere thanks to everyone for your time and feedback, and I'll see you at an indeterminate future time. :)

rcalixte:
> That said, I think I'll return once I have something ready for a second submission. My sincere thanks to everyone for your time and feedback, and I'll see you at an indeterminate future time. :)

I'm looking forward to it! Please let us know if we can do anything to help. There is also the open Issue (5387) that works for future communication as well.

anaximeno:
@torchipeppo I'm not sure but this built-in [applet](https://github.com/linuxmint/cinnamon/tree/master/files/usr/share/cinnamon/applets/xapp-status%40cinnamon.org) of Cinnamon might be a good reference as it displays status icons on Cinnamon.

torchipeppo:
Yes, that applet caught my eye early in the process because XApp Status Icons are one of the few items in Cinnamon that use GTK menus (on right click). The applet itself appears to be just a container for the icons, though: the actual logic of processing UI events, including handling the popup menus, is in a piece of C code that I found recently and that I mentioned in my last comment. I'm still looking into it.

In other news, it appears GitHub picked up on an early fix for the freeze issue on Cinnamon <=5.2, which I pushed to [my personal WIP repo](https://github.com/torchipeppo/directory-menu-for-cinnamon). I don't plan to open a PR right now because the main issue on Cinnamon 5.4+ is still unsolved, but I might if I notice it takes too much time. Meanwhile, it's there if anyone wants to test it.

PR #5383, opened 2024-01-16
qredshift@quintao: hotfix: disable by default under Wayland

anaximeno:
Redshift is not supported yet by Wayland and running the applet while Cinnamon is running under Wayland may cause the system to completely freeze, even not initilizing successfuly. Some people also reported on Reddit having this same issue https://www.reddit.com/r/linuxmint/comments/195dsmu/blank_screen_on_experimental_wayland_session/

cc @raphaelquintao


rcalixte (approved):

rcalixte:
We should update the .pot file for translators too.

anaximeno:
Sure, I'll update them.

Issue #5387, opened 2024-01-17
[directory-menu@torchipeppo] Does not open any menu on Cinnamon 5.4+


claudiux:
### Applet version/Build date

Version 0.1 (2024-01-13 19:12:39)

### Cinnamon version

5.4+

### Distribution

Mint 21+

### Graphics hardware *and* driver used

No matters

### Applet name and maintainer

directory-menu@torchipeppo @torchipeppo 

### What happened?

Successfully tested on Cinnamon 5.0 and 5.2, this applet does not work on Cinnamon from 5.4 to 6.0.

No message in `~/.xsession-errors`, but the menu does not appear when clicking on the applet icon.

torchipeppo:
This issues should be solved by these PRs now:
- https://github.com/linuxmint/cinnamon-spices-applets/pull/5451
- https://github.com/linuxmint/cinnamon-spices-applets/pull/5453

PR #5388, opened 2024-01-18
Remove directory-menu@torchipeppo since it is not working

rcalixte:


anaximeno:
Wouldn't it be better to just point them to older versions of cinnamon lower than 5.4 instead? Unless this is the only way they would not show up in current versions, it seems a bit extreme.

rcalixte:
> Wouldn't it be better to just point them to older versions of cinnamon lower than 5.4 instead? Unless this is the only way they would not show up in current versions, it seems a bit extreme.

[This](https://github.com/linuxmint/cinnamon-spices-applets/pull/5377#issuecomment-1897670714) was the reason it was removed. Apparently it wasn't working on earlier versions either. If this isn't the case, we can definitely revert and just disable it until it's fixed for 5.4. Can you confirm it's fine for 5.2 and lower?

claudiux:
I can confirm that this applet works on 5.0 and 5.2.

rcalixte:
Were either of you able to confirm the right-click issue identified by @fredcw?

torchipeppo:
Just reproduced @fredcw 's issue reproduced on Cinnamon 5.0.7

anaximeno:
Oh, ok @rcalixte, I thought it was an older applet since @claudiux said it was working on older versions, if it is a new one then there's no problem. 

@torchipeppo I believe you can submit a new PR again if you manage to solve the [issue](https://github.com/linuxmint/cinnamon-spices-applets/issues/5387).

PR #5451, opened 2024-02-04
directory-menu@torchipeppo: Updated release, working on Cinnamon 5.4+

torchipeppo:
This is a new release of my applet to solve the issues raised in https://github.com/linuxmint/cinnamon-spices-applets/pull/5377

First off, this PR solves issue https://github.com/linuxmint/cinnamon-spices-applets/issues/5387 :

* After exploring a few possibilities, the best way I could find to have the applet work on Cinnamon 5.4+ is to handle all GTK calls in a Python script, as opposed to GJS. Fortunately, there are a few other applets that do this already that I took as reference.
* During the process, the (deprecated) Gtk.Menu.popup() call was replaced with popup_at_rect(), following the model of XApp status icons at https://github.com/linuxmint/xapp/blob/master/libxapp/xapp-status-icon.c . This also enables the tangible benefit of aligning the popup menu with the applet icon.

Other:

* The freezing issue at https://github.com/linuxmint/cinnamon-spices-applets/pull/5377#issuecomment-1897670714 is taken care of by preventing the applet to take any action if any Cinnamon popup menus are open
* Added settings to change icon and tooltip. Besides customization, this allows to distinguish b/w multiple instances of the applet.
* This version triggers on mouse button *release* rather than press. This rids the applet of the awkward "click tolerance timeout" setting it had in the previous version.

Known issues:
- This version is not as fast to load as the GJS-only version. I have no quantitative data, but since the code is essentially the same, I'd say that the extra delay is due exclusively to starting a new Python process. This hypothesis is supported by the fact that only the "root" menu takes longer to appear, while submenus are much faster despite using the exact same code as the root to be generated and populated.
- It appears that popup_at_rect never shows the menu if the mouse is held at the time it's called, causing the Python process to hang indefinitely. Triggering the applet on mouse release prevents this from being a problem normally, but it can still happen if the user *double-clicks* on the applet: the second release works fine, but the first triggers another Python process that never terminates. So, just don't double-click on the applet. None of the built-in ones requires double-clicking, after all.
- If this applet is on a panel set to automatically hide, the panel will hide as soon as any submenu opens. This is otherwise completely harmless. I don't think there is an easy solution here; attempting to "grab" the same way Cinnamon's PopupMenuManager does just ends up freezing the whole desktop environment. And even the XApp Status Icons have the same issue (I tried the client for Zoom, the video conference app, which puts a status icon there).

Tested on Cinnamon 5.0.7 (Mint 20.2 host) and Cinnamon 6.0.4 (Mint 21.3 VM). Also quickly tried it on Mint 21.1 and 22.2 Live VMs just to make sure I didn't hit any single-version weirdness.


rcalixte (approved):

directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
```diff
@@ -0,0 +1,267 @@
...
+            "action": "launch_default_for_uri",
+            "uri": uri,
+            "timestamp": timestamp,
+        }))
+
+    def open_terminal_at_path(self, path):
+        print(json.dumps({
+            "action": "open_terminal_at_path",
+            "path": path,
+        }))
```
rcalixte, line 144: Should these do more than print?

rcalixte:
@torchipeppo Apologies again as it appears this might have been a rushed approval. I checked that the menu navigation worked since that was the main challenge previously but it doesn't appear that the actual opening of paths/files works at the moment. I've tested some code locally to no avail as of right now but I'll keep going to see if I can get this working.

directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
rcalixte: #5453

torchipeppo:
Alright, I get the window manager warning on my Mint 21.3 VM, but not the kernel error. The windows still open correctly. The VM in question is almost clean since I installed it specifically for testing, so IIRC the only things installed there are Visual Studio Code, the directory-menu applet.

And I get neither of those messages on Mint 20.2

Let me add some extra version info for my machines I didn't mention before to compare with yours, just in case.
|                               | Mint 20.2 host                             |                               Mint 21.3 VM |
|--------------------------|--------------------------------------------|--------------------------------------------|
|`python3 --version` |                                          3.8.10 |                                        3.10.12 |
|   Kernel version *   | Linux 5.4.0-169-generic x86_64  | Linux 5.15.0-91-generic x86_64 |

* just because the message says "kernel", I don't know if it might be relevant

rcalixte:
> * just because the message says "kernel", I don't know if it might be relevant

It's not a visible error to the user but it will flood their logs each time they use the applet. I can circle back to this later if you need a second pair of eyes.

Like you said, the functionality is there. I'll test this on other machines too.

Can you take a look at #5453 when you get a chance?

directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
torchipeppo: Not really, see https://github.com/linuxmint/cinnamon-spices-applets/pull/5453#pullrequestreview-1861465938

torchipeppo:
> Like you said, the functionality is there
So the requested window still opens on kernel v6?

> Can you take a look at https://github.com/linuxmint/cinnamon-spices-applets/pull/5453 when you get a chance?
Left a comment there

rcalixte:
> > Like you said, the functionality is there
> 
> So the requested window still opens on kernel v6?

Neither function launches for me until I made the changes in the PR but that PR isn't correct either.

torchipeppo:
> Neither function launches for me until I made the changes in the PR...

That makes some sense to me: if the kernel error makes your Python process crash, it might follow that the GJS callback doesn't get executed.

By the way, I think the warning is connected to submenus, as I see one warning per time I open (or close, idk) one, but not if I just stick to the root menu

rcalixte:
> By the way, I think the warning is connected to submenus, as I see one warning per time I open (or close, idk) one, but not if I just stick to the root menu

Correct, the Window warning looks like it is only triggered for the submenus.

The kernel message appears whenever the menu closes so it looks like it's independent of the execution or the action.

torchipeppo:
I think the warning is not *just* related to this applet: I got it to log also by right-clicking on the desktop and on nemo, but not on Firefox or Visual Studio Code (all on my Mint 21.3 clean-ish VM).


Can anyone attempt to replicate this?
- Right-click on your desktop
- Open any sub-menu, e.g. by hovering over "Create New Document"
- Close the menu
- `cat .xsession-errors` and check if a warning was logged that is similar to the first line here https://github.com/linuxmint/cinnamon-spices-applets/pull/5451#issuecomment-1925776122

You should also be able to see the warning by:
- Right-clicking on Nemo (good sub-menus to open are "Arrange items" if you clicked on empty space or "Open with" if you clicked on a file)
- Opening a sub-menu from the menu bar of Nemo (e.g. View -> Sidebar)
- Doing the same thing with Cinnamon's default PDF viewer (File -> Recents)

You should not see the warning if you try this on:
- Firefox (e.g. right-click on a tab -> Move Tab)
- Visual Studio Code (e.g. File -> Open Recent)

rcalixte:
> I think the warning is not _just_ related to this applet: I got it to log also by right-clicking on the desktop and on nemo, but not on Firefox or Visual Studio Code (all on my Mint 21.3 clean-ish VM).

I can duplicate this. Fascinating. We should open an Issue for Nemo.

torchipeppo:
This also happens on Cinnamon's PDF viewer though

rcalixte:
> This also happens on Cinnamon's PDF viewer though

Oh boy. Probably the Cinnamon repository then.

torchipeppo:
Also `xviewer` (image viewer, File -> Open with), so yeah, either Cinnamon or the XApps (I don't know if they have a separate repo since I believe they are available for other distros/DEs?)

rcalixte:
> Also `xviewer` (image viewer; File -> Open with), so yeah, either Cinnamon or the XApps (I don't know if they have a separate repo since I believe they are available for other distros/DEs?)

They are individual repositories but it looks like the issue is one with Cinnamon itself. We can handle that separately.

torchipeppo:
@rcalixte Meanwhile I've been thinking about this PR, what exactly happens with your version? Does the app/terminal open? Do you still get the kernel error or not? Because I think it is possible to do the gsettings lookup in the Python script, so we might actually move the execution to the Python side

rcalixte:
> @rcalixte Meanwhile I've been thinking about your PR, what exactly happens with your version? Does the app/terminal open? Do you still get the kernel error or not? Because I think it is possible to do the gsettings lookup in the Python script, so we might actually move the execution to the Python side

We're thinking the same thing. 😄

In the original version here, I was able to get the directory listing. That was why I merged. When I went to use either action, neither worked. I couldn't get a folder to open and I couldn't get a terminal at either path. That led me to looking around and the logs and whatnot.

I just pushed a change that takes the execution to the Python side and works for me. Thanks for an awesome applet! I can't tell you how much time this is going to save me! 😁

PR #5453, opened 2024-02-04
directory-menu@torchipeppo: Add execution to 'Open' functions

rcalixte:
Add translation template
Add a terminal option for non-Debian-based distros

This addresses the execution functionality of the applet.

cc @torchipeppo

### DIFF
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
@@ -46,31 +46,6 @@ class CassettoneApplet extends Applet.IconApplet {
         this.actor.connect('button-release-event', Lang.bind(this, this.on_button_release_event));
     }
 
-    // essentially an independent JS translation of xapp_favorites_launch from the Favorites Xapp.
-    launch(uri, timestamp) {
-        let display = Gdk.Display.get_default();
-        let launch_context = display.get_app_launch_context();
-        launch_context.set_timestamp(timestamp);
-        Gio.AppInfo.launch_default_for_uri_async(uri, launch_context, null, this.launch_callback);
-    }
-
-    // emulates how nemo handles opening in terminal (using the same flags as Util.spawn)
-    open_terminal_at_path(path) {
-        let gnome_terminal_preferences = Gio.Settings.new("org.cinnamon.desktop.default-applications.terminal");
-        let default_terminal = gnome_terminal_preferences.get_string("exec");
-        let argv = [default_terminal];
-        let spawn_flags = GLib.SpawnFlags.SEARCH_PATH
-                    | GLib.SpawnFlags.STDOUT_TO_DEV_NULL
-                    | GLib.SpawnFlags.STDERR_TO_DEV_NULL;
-        GLib.spawn_async(path, argv, null, spawn_flags, null);
-    }
-
-    launch_callback(source_object, result) {
-        if (!Gio.AppInfo.launch_default_for_uri_finish(result)) {
-            log("An error has occurred while launching an item of the Directory Menu.")
-        }
-    }
-
     normalize_tilde(path) {
         if (path[0] == "~") {
             path = "file://" + GLib.get_home_dir() + path.slice(1);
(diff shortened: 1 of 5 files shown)


directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
```diff
@@ -136,12 +137,17 @@ def launch(self, uri, timestamp):
             "uri": uri,
             "timestamp": timestamp,
         }))
+        subprocess.run(['/usr/bin/xdg-open', uri], check=False)
```
torchipeppo, line 140: [1/3] In my design for this version, the launch and open_terminal_at_path functions *are* supposed to do nothing but printing. This sends the JSON string on stdout to be received by applet.js, which then takes care of executing the requested action.

directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
```diff
@@ -136,12 +137,17 @@ def launch(self, uri, timestamp):
...
 
     def open_terminal_at_path(self, path):
         print(json.dumps({
             "action": "open_terminal_at_path",
             "path": path,
         }))
+        terminal = self.terminal_cmd if self.terminal_cmd else "/usr/bin/x-terminal-emulator --working-directory"
```
torchipeppo, line 147: [2/3] Same as the comment for the launch function above

directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
```diff
@@ -163,6 +164,7 @@ class CassettoneApplet extends Applet.IconApplet {
             "x": x,
             "y": y,
             "orientation": o,
+            "terminal_cmd": this.terminal_cmd,
```
torchipeppo, line 167: [3/3] The callback function  beginning at line 170 or 172 is the one that actually takes care of executing the action clicked on the menu, as per the comments on the Python file. The callback is executed after the Python process terminates, with the "response" argument being whatever the process wrote to stdout.

torchipeppo (commented):
Hi, I notice that this PR is adding execution functionality to the Python code which is already taken care of on the GJS side, was there some problem with the submitted version?

I added line-specific comments to illustrate the communication between the Python and GJS sides.

I will admit this is not the most elegant design, but it is the best I could do with my knowledge: I was not aware of `xdg-open`, and `Gio.AppInfo.launch_default_for_uri_async` and `GLib.spawn_async` gave me problems I can't remember right now in Python.

directory-menu@torchipeppo/files/directory-menu@torchipeppo/popup_menu.py
rcalixte: Hmm. That wasn't working for me. I'll test this on another machine. I do like the gsettings lookup as a cleaner alternative that doesn't require a new preference.

rcalixte: Did you test this with other terminals than gnome-terminal?

torchipeppo: My default terminal is `terminator` actually

rcalixte: > My default terminal is `terminator` actually

Same! I'm a newcomer as of a few months. You have great taste! 😁

rcalixte:
@torchipeppo This is ready for your review. 🙏🏼

torchipeppo:
The code looks good, and from a quick test it looks like this version can open folders, terminals and files on my Mint 20.2 machine without issue. Thank you so much for your help :D

Just a couple minor points, to align the linuxmint repo with my personal one with all the dev history:
* I'd like to move your code there, with your permission;
* And I think we should increment the version number, since v0.3 is supposed to be my submission from a few hours ago. Moving the execution logic sounds like more than patch-level to me, so maybe this can be v0.4.0?

torchipeppo:
Edited the previous message with links.

torchipeppo (changes requested):
Remembered where the review button was

rcalixte:
> * I'd like to put your code there too, with your permission;

No permission needed! Feel free! That's why we open source. 🤓



> * Moving the execution logic sounds like more than patch-level to me, so maybe this can be v0.4.0?

You got it.

rcalixte:
Alright. Now we're ready. 😄 


Feel free to update the README when you're ready too. We'll get that merged for you quickly.

torchipeppo:
And with that, I'd say we can merge this, as far as I'm concerned! ^_^

torchipeppo:
I don't think I can push anything on this PR, though, right?

torchipeppo (approved):
Only the version number changed since the last review, so it's good

rcalixte:
> I don't think I can push anything on this PR, though, right? So this should be merged first, and then I'd open a PR for the readme

Correct. Sorry I didn't make that part clear.

PR #5454, opened 2024-02-04
add sct@skulptist.de v1

Matthias-Hermsdorf:
This applet uses sct to toggle through screen temperatures. You can set up to 7 steps and which icon you want to use.


sct@skulptist.de/files/sct@skulptist.de/metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+    "uuid": "sct@skulptist.de",
+    "author": "Matthias Hermsdorf",
+    "version": "1.0.0",
+    "name": "SCT Toggle",
+    "description": "Toggle screen temperatures with sct. \nSet up to 7 custom color temperatures and toggle through by clicking on the applet",
+    "max-instances": "1"
```
rcalixte, line 7: Can this support more than one instance for multi-monitor setups?

sct@skulptist.de/info.json
```diff
@@ -0,0 +1,3 @@
+{
+    "author": "Matthias Hermsdorf"
```
rcalixte, line 2: This should be your GitHub username. The name of the applet probably should too but that is a loosely defined convention.

rcalixte (changes requested):
Mainly the info.json needs to be corrected and you need to include a screenshot.png file in the root of your directory.

Feel free to take a look through the README and shout if you have any questions! This looks great!

sct@skulptist.de/info.json
Matthias-Hermsdorf: I will change the author. Changing the applet name is possible. My page is https://skulptist.de and my own name is so long. If you insist, I would change it.

sct@skulptist.de/files/sct@skulptist.de/metadata.json
Matthias-Hermsdorf: sct has no option to set the screen color temperature for different screens. So I have had set the max-instances to 1.
I will try what will happen with more instances.

sct@skulptist.de/info.json
rcalixte: I don't insist. 😅

It's fully up to you either way.

sct@skulptist.de/files/sct@skulptist.de/metadata.json
Matthias-Hermsdorf: At the moment every applet would have its own state. Think about having 7 steps configurated and click on one applet to the 5. and then clicking on another instance, it would switch from 1 to 2. 
With more instances they should share the currentStep state. Yes, its implementable.

sct@skulptist.de/README.md
```diff
@@ -0,0 +1,29 @@
+# SCT Toggle 
+
+A Linux Mint Cinnamon Applet to toggle through color temperatures.
```
rcalixte, line 3: Just a comment, Cinnamon Spices are distro-independent.

rcalixte (approved):

sct@skulptist.de/README.md
Matthias-Hermsdorf: Oh, you are right.
Are they tied to Cinnamon or are there other runtimes that use the spices?

rcalixte: By default, the code is part of Cinnamon itself. Distros could modify this if they wanted to of course but I don't know of any that do.

As far as compatibility, only the Themes could be ported or compatible elsewhere depending on how they are structured. The recently published Actions can be used anywhere with Nemo as well, Cinnamon or not.

Matthias-Hermsdorf: Ok, I will modify that part in the next version.

Matthias-Hermsdorf:
Thanks for merging

sct@skulptist.de/README.md
rcalixte: No rush! 😅

rcalixte:
> Thanks for merging

Thanks for your contribution!

PR #5458, opened 2024-02-04
directory-menu@torchipeppo: Rephrase the description for better comprehension of the functionality

anaximeno:
cc @torchipeppo

I believe this way it will be clearer to the users what the functionality of this cool applet is.

PR #5459, opened 2024-02-04
directory-menu@torchipeppo: Adapt popup menu for translation

anaximeno:
This will allow the texts "Open in folder" and "Open in terminal" on the popup menu to be translated into other languages.

cc @torchipeppo

Blocked by #5458

### DIFF
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/po/directory-menu@torchipeppo.pot
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/po/directory-menu@torchipeppo.pot
@@ -45,3 +45,11 @@ msgstr ""
 #. settings-schema.json->show-hidden->description
 msgid "Include hidden files"
 msgstr ""
+
+#. popup_menu.py:71
+msgid "Open Folder"
+msgstr ""
+
+#. popup_menu.py:76
+msgid "Open in Terminal"
+msgstr ""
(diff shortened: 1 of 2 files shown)


rcalixte:
You would probably want to update the template as well.

anaximeno:
Updated

rcalixte (approved):

rcalixte:
@torchipeppo This was merged because it is essentially a bug-fix for translation support.

torchipeppo:
No problem, that's a benefit for everyone. Maybe I'll even contribute an Italian localization one of these days (though I am *not* claiming the task right now).

As a newcomer here, I'm curious: is it common practice (or even necessary) to mention specific lines of code in the translation template? Isn't it a problem if they get misaligned?

rcalixte:
> As a newcomer here, I'm curious: is it common practice (or even necessary) to mention specific lines of code in the translation template? Isn't it a problem if they get misaligned?

I'm assuming you're referring to the line numbers from the code that ends up in the comments. The comments in the templates and the translations don't matter for functionality. It's to aid translators if they need to know where a string is located and what a string pertains to. We all know English is very ambiguous and hard to translate sometimes. We could remove the comments and the translations would work the same though. Think of the translations as a database of strings. All that happens in each localization is a string look-up, one-to-one. So if strings change by even a single character, it would be best to update the translation template. For just line numbers in the comments, that alone doesn't merit updating the template, though it would help new translators and translations that haven't been completed yet. The easier it is to find when/how a string will be shown and the context around it should help translators do the awesome work they do a little easier.

torchipeppo:
Ok, I see, so # lines are just comments. I had assumed they were something more semantically meaningful, like references or labels.

Thank you! :)

PR #5460, opened 2024-02-04
directory-menu@torchipeppo - add spanish translation

haggen88:


rcalixte:
It's okay that the check failed. There's two more strings coming in #5459. Feel free to push the new strings and the fixes here @haggen88. Thanks for always being so quick!

rcalixte:
Ready for you! Shout if you need a hand!

rcalixte:
@haggen88 You don't need to close this. You can push the new commit here and we can squash and merge it.

PR #5464, opened 2024-02-05
sct@skulptist.de - add spanish translation

haggen88:


rcalixte:
@haggen88 I've rolled in the other changes that just got pushed out but I'm wondering if this is a bug in Poedit itself. This is happening far too often and I don't think it's anything that you are doing. And nothing changed on our end related to this.

rcalixte (approved):

PR #5485, opened 2024-02-08
Fix "Main.Util undefined" error

### LINKED ISSUE #5444: [CustomApplicationsMenu@LLOBERA] Error launch command on first click
### Applet version/Build date

2.1 (28-07-2019)

### Cinnamon version

6.0.4

### Distribution

Mint 21.3

_No response_

CustomApplicationsMenu@LLOBERA @Nicolas01

### What happened?

The applet loads fine, and shows the menus I configured.

But when I click on any of the entries of the menu, the command will not be executed and the following errors are recorded on xsession-errors

```text
(cinnamon:2406): Gjs-WARNING **: 12:16:44.524: Some code accessed the property 'Util' on the module 'main'. That property was defined with 'let' or 'const' inside the module. This was previously supported, but is not correct according to the ES6 standard. Any symbols to be exported from a module must be defined with 'var'. The property access will work as previously for the time being, but please fix your code anyway.
(cinnamon:2406): Gjs-WARNING **: 12:16:44.525: JS ERROR: Exception in callback for signal: activate: TypeError: Main.Util is undefined

```
After I click for the second time at the same menu entry, the error is no longer shown and the command will be executed properly for the rest of the session.

So the error only occurs when clicking for the first time in any of the menu items, after starting the session, or after restarting cinnamon.

My configuration entry is this

```json
{
    "active": true,
    "iconName": "application-vnd.ms-powerpoint",
    "displayName" : "Presentación Distribuidores",
    "command": "soffice /home/xxxxxxx/Documentos/xxxxx/presentaciones/presentacion_dist.odp"
},
```

This is an image recording the event, the output of the terminal window error, two consecutive clicks on the same menu entry, first one logs an error and the command will not be executed, second click starts the command correctly.

Peek 2024-02-02 12-16

### Other information

_No response_

Nicolas01:
While clicking for the first time on an item menu, the command is not executed and the following error is recorded on xsession-errors:
`JS ERROR: Exception in callback for signal: activate: TypeError: Main.Util is undefined`

fix #5444

### DIFF
--- a/ShutdownMenuWithIcons@LLOBERA/files/ShutdownMenuWithIcons@LLOBERA/applet.js
+++ b/ShutdownMenuWithIcons@LLOBERA/files/ShutdownMenuWithIcons@LLOBERA/applet.js
@@ -1,8 +1,8 @@
 // Shutdown Menu With Icons Cinnamon Applet
-// Developed by Nicolas LLOBERA - nllobera@gmail.com from the System Shutdown and Restart Applet by Shelley
-// version: 2.2 (28-08-2013)
+// Developed by Nicolas LLOBERA from the System Shutdown and Restart Applet by Shelley
+// version: 2.3 (08-02-2024)
 // License: GPLv3
-// Copyright © 2013 Nicolas LLOBERA
+// Copyright © 2024 Nicolas LLOBERA
 
 const Gettext = imports.gettext;
 const Gio = imports.gi.Gio;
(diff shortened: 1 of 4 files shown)


ShutdownMenuWithIcons@LLOBERA/files/ShutdownMenuWithIcons@LLOBERA/applet.js
```diff
@@ -1,23 +1,23 @@
...
+const Gtk = imports.gi.Gtk;
 const St = imports.gi.St;
 const Util = imports.misc.util;
-const GLib = imports.gi.GLib;
+const Applet = imports.ui.applet;
+const PopupMenu = imports.ui.popupMenu;
+const Settings = imports.ui.settings;
+
+const UUID = "ShutdownMenuWithIcons@LLOBERA";
 const AppletUUID = "ShutdownMenuWithIcons@LLOBERA";
```
rcalixte, line 18: Just a note here, you're defining the same thing twice. You could use one and just replace every reference.

rcalixte (approved):

PR #5503, opened 2024-02-12
[suspend@janax] Add  double-click and count-down options

klangman:
I believe this applet has been abandoned, so I am offering the following changes. I could also submit this as a separate applet if you think that better given the extent of the changes?

1. Rewritten in Javascript EC6
2. Added an option to require a double-click to suspend (reducing accidental suspends).
3. Added a count-down option to allow the user some time to abort the suspend by clicking on the applet icon a second time. The count-down is displayed using a icon-overlay badge like the one used by the grouped-windowlist applet for showing the window count.
4. Added an option to configure the count-down duration
4. Added a README.md.
5. Made the applet allow multiple instances (mainly for people with multiple monitors).
6. Added an option to use a symbolic or full color icon on the panel.

By default it still instantly suspends with a single click.

### DIFF
--- a/suspend@janax/files/suspend@janax/README.md
+++ b/suspend@janax/files/suspend@janax/README.md
@@ -1,0 +1,9 @@
+Immediately suspends the computer (by default)
+
+## Features
+
+1. Suspend instantly on a single click or only suspend on a double-click to avoid accidentally suspending the computer.
+2. Optionally add a count-down timer (1-9 sec) giving you the chance to cancel the suspend operation. 
+3. A single click on the panel button during the count down will cancel the suspend operation.
+4. Choose between a full color or a symbolic panel icon for the applet.
+
--- a/suspend@janax/files/suspend@janax/applet.js
+++ b/suspend@janax/files/suspend@janax/applet.js
@@ -2,42 +2,135 @@ const Lang = imports.lang;
 const Applet = imports.ui.applet;
 const GLib = imports.gi.GLib;
 const Gettext = imports.gettext;
+const Settings = imports.ui.settings;
+const Mainloop = imports.mainloop;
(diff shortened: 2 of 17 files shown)


rcalixte:
> I believe this applet has been abandoned, so I am offering the following changes. I could also submit this as a separate applet if you think that better given the extent of the changes?

I think this is fine. Do you want to add yourself as a maintainer? And do you want to add a version to the metadata?

klangman:
I don't mind being the maintainer, since there really shouldn't be a lot to do. It's a very simple applet that survived for years without many changes.  As for the version, sure, We can call it 1.1, 2.0 or something like that I guess.

rcalixte:
> I don't mind being the maintainer, since there really shouldn't be a lot to do. It's a very simple applet that survived for years without many changes. As for the version, sure, We can call it 1.1, 2.0 or something like that I guess.

Done. If you're okay with the changes, I'll go ahead and merge.

klangman:
Looks good to me.

rcalixte (approved):

klangman:
@rcalixte I wonder why the README.md content is not showing up on https://cinnamon-spices.linuxmint.com/applets/view/236 ??

rcalixte:
> @rcalixte I wonder why the README.md content is not showing up on https://cinnamon-spices.linuxmint.com/applets/view/236 ??

It needs to be at the root of the Spices, not a subdirectory. You can include the CHANGELOG there as well if you would like.

klangman:
> It needs to be at the root of the Spices, not a subdirectory. You can include the CHANGELOG there as well if you would like.

Ah.. I saw the README.txt and didn't think about it at all before just putting the README.md in the same spot. I'll fix that  tonight. Thanks!

Issue #4711, opened 2023-03-07
color-picker@fmete breaks when using display scaling. Almost always picks color #000000


mrmatteastwood:
I don't know how to find out the applet's UUID, unfortunately.

```
 * Applet version/Build date: "About" screen says "Updated on 2023-02-14 09:54:44"
 * Cinnamon version (cinnamon --version): 5.6.7
 * Distribution - Linux Mint 21.1 Vera (base: Ubuntu 22.04 jammy)
 * Graphics hardware *and* driver used: NVIDIA GeForce RTX 3080 Ti Laptop GPU/PCIe/SSE2 v: 4.6.0 NVIDIA 525.85.05
 * Snippet or attachment of ~/.cinnamon/glass.log: can't find that file on my system
 * Snippet or attachment of ~/.xsession-errors: attached

[xsession-errors.txt](https://github.com/linuxmint/cinnamon-spices-applets/files/10910819/xsession-errors.txt)

 ```

**Notify author of applet**
@icarter09

**Issue**
Color picker doesn't pick the correct colors. It mostly picks color #000000, regardless of where the user clicks. Video: https://youtu.be/24Eutz8kiDo. Issue only occurs when display scaling is set to a value other than 100%.

**Steps to reproduce**
1. Install color picker and required packages, xclip and python3-xlib
2. Set display scaling to 125%
3. Try to pick colors

**Expected behaviour**
Color picker should pick the correct colors.

**Other information**
Full PC specs: https://termbin.com/jmf4i

BlessedDisco:
Created a pull request to fix this issue. Hopefully it can be merged soon

PR #5522, opened 2024-02-21
Fix color-picking when display scaling is used

BlessedDisco:
Fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/4711

The provided fix allows the color picker to also work when display scaling is used. Previously when using the color picker with display scaling, it would report either #000000 or an incorrect color (when picking a color somewhere in the upper left color of the screen).

Tested locally with a multi-monitor setup with different resolutions for each monitor and with multiple different scaling settings.


rcalixte:
cc @icarter09

rcalixte:
Merging as this is a bug fix and Isaac doesn't appear to be active.

PR #5563, opened 2024-03-06
directory-menu@torchipeppo: Add support for Flatpak terminals

rcalixte:
cc @torchipeppo This adds support for terminals installed via Flatpak (e.g. [Black Box](https://flathub.org/apps/com.raggesilver.BlackBox))

Shout if you want to change the version too.

### DIFF
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/metadata.json
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/metadata.json
@@ -2,7 +2,7 @@
     "uuid": "directory-menu@torchipeppo",
     "name": "Directory Menu",
     "description": "Quickly navigate through directories on your system using a dropdown menu.",
-    "version": "0.5",
+    "version": "0.5.1",
     "author": "torchipeppo",
     "max-instances": "-1"
 }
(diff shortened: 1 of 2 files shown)


torchipeppo (approved):
I see no issues with that.
Would you mind incrementing the patch version number?

rcalixte:
> Would you mind incrementing the patch version number?

I assume 0.6 is okay?

torchipeppo:
Sorry about the misunderstanding, by "patch" I meant the third number (which is not explicitly there at the moment), resulting in 0.5.1
But at this point I feel like I might be getting too pedantic over something ultimately inconsequential, so I'll be fine either way.

rcalixte:
> Sorry about the misunderstanding, by "patch" I meant the third number (which is not explicitly there at the moment), resulting in 0.5.1 But at this point I feel like I might be getting too pedantic over something ultimately inconsequential, so I'll be fine either way.

It's fine! I'm glad I checked with you first. I just spaced on the semantic versioning myself so I'm glad to get it right according to your wishes.

torchipeppo:
All good. I only thought about it because I read it on a whim semi-recently, too.
Anyway, thank you for the update! ^_^

PR #5614, opened 2024-03-21
[CassiaWindowList@klangman] New mouse actions and some fixes

klangman:
1. Added 9 mouse button actions for window tiling
2. Added mouse button action to move a window to the current workspace
3. Fixed several cases where the icon saturation effect was not changing to reflect a change in a window state
4. Changed icon saturation effect for grouped window list buttons to only consider the last-focused window rather then requiring all window to have the same qualifying state before using the effect
5. Changed how the applet was deciding on code paths for features that are only possible in newer releases. This fixes some issues when running on Cinnamon 5.x releases prior to 5.4.
6. Added version numbering, started with version 2.0 based on the fact that I have already far exceeded the feature set that I originally set out to implement.
7. Added a CHANGELOG.md file


rcalixte (approved):

klangman:
@rcalixte I guess I didn't check the translation change closely enough, sorry.  I started using "cinnamon-spices-makepot" rather then the "makepot" script I had created, but I guess "cinnamon-spices-makepot" is not looking at the SetupWizard.ui file? Maybe if I rename it to SetupWizard.glade it would be picked up? or should I go back to using my makepot script?

Thanks for spotting my error here.

rcalixte:
> Thanks for spotting my error here.

It's actually an oversight with the script so totally not an error on your part. I'm working on some changes that I hope to push today so your process doesn't have to change. The intention is to have a standardized script as a precursor to hopefully automating away some of the process.

PR #5651, opened 2024-03-31
placesCenter@scollins v2.3 - Allows to show entire path in Recent Documents tooltips

claudiux:
@collinss 
Hi Stephen,
This PR is to show the uri of each recent document in its tooltip.
This is useful when several documents have the same name (like `applet.js` - which applet is it?).
I hope you find these changes satisfactory.

Regards
claudiux


claudiux:
@collinss Up?

PR #5663, opened 2024-04-03
Sensors@claudiux: Fix typo on de.po

mikehaertl:


rcalixte (approved):

claudiux:
@mikehaertl Thank you! Can you translate this single string shown here? https://github.com/linuxmint/cinnamon-spices-applets/blob/translation-status-tables/.translation-tables/po/Sensors@claudiux/_de.po

mikehaertl:
@claudiux We also call it "Tooltip" in German. So should be fine to leave the translation empty in that case, right?

claudiux:
@mikehaertl Yes, you can leave it empty.

PR #5668, opened 2024-04-07
directory-menu@torchipeppo: Change to TextIconApplet and other improvements

anaximeno:
Hi @torchipeppo 

I made some changes here:

1. Turned the applet from an IconApplet to a TextIconApplet this way it will be possible to display a text label in the front of the icon as shown in the image below:

Captura de ecrã de 2024-04-07 17-38-14

Captura de ecrã de 2024-04-07 17-38-54

2. Updated the Character Limit setting to show a switch instead to enable the character limiting, and only when it is enabled that the user has to define the character limit (the logic on the Python script continues the same)

Captura de ecrã de 2024-04-07 17-44-29

3. I also removed the tooltip text entry in favor of the panel text added in 1. and instead made it static for "Directory Menu" which is the title of the applet

Please tell me if you approve these changes.

### DIFF
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
@@ -36,6 +36,7 @@ class CassettoneApplet extends Applet.TextIconApplet {
         this.settings.bind("show-hidden", "show_hidden", null, null);
         this.settings.bind("icon-name", "icon_name", this.set_applet_icon_symbolic_name, null);
         this.settings.bind("label", "label", this.set_applet_label, null);
+        this.settings.bind("limit-characters", "limit_characters", null, null);
         this.settings.bind("character-limit", "character_limit", null, null);
         this.starting_uri = this.normalize_tilde(this.starting_uri);
 
@@ -140,7 +141,7 @@ class CassettoneApplet extends Applet.TextIconApplet {
             "x": x,
             "y": y,
             "orientation": o,
-            "character_limit": this.character_limit,
+            "character_limit": this.limit_characters ? this.character_limit : -1,
         }
 
         Util.spawn_async(['python3', `${this.metadata.path}/popup_menu.py`, JSON.stringify(args)]);
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
@@ -21,8 +21,22 @@ const Gio = imports.gi.Gio;
 const St = imports.gi.St;
 const Cinnamon = imports.gi.Cinnamon;
 const Settings = imports.ui.settings;
+const Gettext = imports.gettext;
+
+
 const UUID = "directory-menu@torchipeppo";
 
+Gettext.bindtextdomain(UUID, GLib.get_home_dir() + "/.local/share/locale");
+
+function _(text) {
+	let locText = Gettext.dgettext(UUID, text);
+
+	if (locText == text) {
+		locText = window._(text);
+	}
+
+	return locText;
+}
 
 
 class CassettoneApplet extends Applet.TextIconApplet {
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
@@ -25,8 +25,7 @@ const UUID = "directory-menu@torchipeppo";
 
 
 
-class CassettoneApplet extends Applet.IconApplet {
-
+class CassettoneApplet extends Applet.TextIconApplet {
     constructor(metadata, orientation, panel_height, instance_id) {
         super(orientation, panel_height, instance_id);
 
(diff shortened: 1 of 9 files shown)


rcalixte:
> 3\. I also removed the tooltip text entry in favor of the panel text added in 1. and instead made it static for "Directory Menu" which is the title of the applet

I actually liked this because it helped differentiate multiple instances on the same panel. It would be nice to have some flexibility here.

rcalixte:
> 1. Turned the applet from an IconApplet to a TextIconApplet this way it will be possible to display a text label in the front of the icon as shown in the image below:

Can you add a screenshot of the before and after? I guess I'm not seeing the difference quite yet.

anaximeno:
> > 3. I also removed the tooltip text entry in favor of the panel text added in 1. and instead made it static for "Directory Menu" which is the title of the applet
> 
> I actually liked this because it helped differentiate multiple instances on the same panel. It would be nice to have some flexibility here.

Wouldn't the change in 1. help with that?

anaximeno:
@rcalixte I improved the screenshots with a before and after comparison for better comprehension of the changes

rcalixte:
> > > 3. I also removed the tooltip text entry in favor of the panel text added in 1. and instead made it static for "Directory Menu" which is the title of the applet
> > 
> > 
> > I actually liked this because it helped differentiate multiple instances on the same panel. It would be nice to have some flexibility here.
> 
> Wouldn't the change in 1. help with that (For example if each instance starts from one place you could change the panel text with the name of the start place)?

Maybe make them independent options?

anaximeno:
> > > > 3. I also removed the tooltip text entry in favor of the panel text added in 1. and instead made it static for "Directory Menu" which is the title of the applet
> > > 
> > > 
> > > I actually liked this because it helped differentiate multiple instances on the same panel. It would be nice to have some flexibility here.
> > 
> > 
> > Wouldn't the change in 1. help with that (For example if each instance starts from one place you could change the panel text with the name of the start place)?
> 
> Maybe make them independent options?

They aren't really "dependent", it's just that it didn't seem necessary after the change in 1. But since it seems that, there's a case where you would find it useful (even with the option in 1), then I will add that back, is that ok?

rcalixte:
> They aren't really "dependent", it's just that it didn't seem necessary after the change in 1. But since it seems that, there's a case where you would find it useful (even with the option in 1), then I will add that back, is that ok?

I may just be confused.

anaximeno:
> I may just be confused.

As a quick summary, originally when you added the directory-menu applet to the panel, it would only show an icon corresponding to the applet, what I have done was change the type of the applet to a TextIcon applet which is an applet that can display both an icon and (optionally) some text with the icon in the panel.

rcalixte:
> As a quick summary, originally when you added the directory-menu applet to the panel, it would only show an icon corresponding to the applet, what I have done was change the type of the applet to a TextIcon applet which is an applet that can display both an icon and (optionally) some text with the icon in the panel.

Ah yeah, for some reason I thought it already had panel text as an option.

Unrelated, how do you (and @torchipeppo) feel about an option for a key binding that opens the menu? It would default to empty of course but it would help for folks who want to navigate solely via the keyboard.

anaximeno:
> Unrelated, how do you (and @torchipeppo) feel about an option for a key binding that opens the menu? It would default to empty of course but it would help for folks who want to navigate solely via the keyboard.

I'm in favor of that, since you can also navigate through menus and submenus with the keyboard, it would make sense to have a key binding open and close it too.

rcalixte:
> I'm in favor of that, since you can also navigate through menus and submenus with the keyboard, it would make sense to have a key binding to open and close it too.

The Escape key already closes the menu so we would just need a binding for the open menu function I think.

torchipeppo:
@anaximeno @rcalixte 
Let me go through the points one by one:

> Turned the applet from an IconApplet to a TextIconApplet

I can see how that would help in certain setups, so I think it's a good addition. I will require that the label is optional *(I assume you just set it to the empty string?)* and the configurable tooltip remains, though, if only because I expect an extra text label to be detrimental for my personal use case. But, again, this is part of a community, and I always like seeing a range of customization options on the stuff I use, so the more people this can serve, the better.

So, in conclusion, more customization is good as long as it doesn't break the current look.

> Character limit switch

Nice! The "-1 to disable" thing has always been nothing more than a quick way to get the job done.

> Improve the adaptation to vertical panels

I don't think this point is further explained? I'll have to check the code.

> An option for a key binding that opens the menu

I'm also in favor.

directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
```diff
@@ -36,15 +49,21 @@ class CassettoneApplet extends Applet.IconApplet {
...
-        this.set_applet_icon_symbolic_name(this.icon_name);
         this.set_applet_tooltip(_(this.tooltip_text));
+        this.set_applet_icon_symbolic_name(this.icon_name);
+        this.set_show_label_in_vertical_panels(false);
+        this.set_applet_label(this.label)
 
         this.actor.connect('enter-event', Lang.bind(this, this.on_enter_event));
         this.actor.connect('button-release-event', Lang.bind(this, this.on_button_release_event));
+
+		this.setAllowedLayout(Applet.AllowedLayout.BOTH);
```
torchipeppo, line 66: I still don't understand the "improved vertical panels" thing. Is it this line?

torchipeppo (approved):

directory-menu@torchipeppo/files/directory-menu@torchipeppo/applet.js
anaximeno: Yes, generally, when you moved the applet to a vertical panel it would alert to the user that it might not work well there, but for what I have tested it works without any issues in a vertical panel, also theres one more line above that disables the text on vertical panels if set.

anaximeno:
@torchipeppo thank you for your points. The text label is set to an empty string by default (in the image above I decided to put a text there to show how it would work), and after taking with @rcalixte we decided that the tooltip remains (there are points on why I decided to remove an add it again in the comments above) and improving the adaption to the vertical panel is indeed related to the line you mentioned in the code review below. In conclusion the default look (the one you get after the installation) keeps the same, only the user will have the opportunity to add a text label to the panel if they want to, and the character limiting will work a bit differently.

torchipeppo:
@anaximeno It's all good from my side then.

Thank you for your contribution! ^_^

rcalixte (approved):

PR #5673, opened 2024-04-08
ddcci-multi-monitor@tim-we: Added changing brightness on scrolling applet icon.

GingerGigiCat:
Added ability to scroll on the applet icon to change the brightness without having to actually open it and move the slider. (With a lot of inspiration and code taken from the built-in [sound@cinnamon.org](https://github.com/linuxmint/cinnamon/tree/8214ed15927d59e4a7054404c2c584eeb4ba4a05/files/usr/share/cinnamon/applets/sound%40cinnamon.org)) 
I haven't tested it on a multi-monitor setup, but I think it would just increase or decrease the brightness of all the monitors.

I also changed a few quirks of the applet, such as not showing the real brightness of the monitor when opened, just the last brightness it was set to by the applet.

### DIFF
--- a/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
+++ b/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
@@ -119,7 +119,7 @@ class DDCMultiMonitor extends Applet.IconApplet {
         this.set_applet_icon_symbolic_name("display-brightness");
         this.set_applet_tooltip(DEFAULT_TOOLTIP);
         this.actor.connect('scroll-event', (...args) => this._onScrollEvent(...args));
-        this.lastTooltipTimeoutID = setTimeout(() => {}, 1);
+        this.lastTooltipTimeoutID = null;
 
         this.menu = new Applet.AppletPopupMenu(this, orientation);
         this.menuManager = new PopupMenu.PopupMenuManager(this);
--- a/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
+++ b/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
@@ -24,6 +24,7 @@ function log(message, type = "debug") {
 }
 
 class Monitor {
+    #promises;
     constructor(index, name, bus) {
         this.index = index;
         this.name = name;
@@ -31,6 +32,7 @@ class Monitor {
         this.bus = bus;
         this.menuLabel = null;
         this.menuSlider = null;
+        this.#promises = Promise.resolve();
     }
 
     updateBrightness() {
@@ -72,13 +74,14 @@ class Monitor {
     setBrightness(value) {
         this.brightness = Math.round(value);
         this.updateMenu();
-
-        return new Promise((resolve, reject) => {
-            Util.spawnCommandLineAsync(
-                `ddcutil --bus=${this.bus} setvcp 10 ${this.brightness}`,
-                resolve,
-                reject
-            );
+        this.#promises = this.#promises.then(() => {
+            return new Promise((resolve, reject) => {
+                Util.spawnCommandLineAsync(
+                    `ddcutil --bus=${this.bus} setvcp 10 ${this.brightness}`,
+                    resolve,
+                    reject
+                );
+            });
         });
     }
 
@@ -116,6 +119,7 @@ class DDCMultiMonitor extends Applet.IconApplet {
         this.set_applet_icon_symbolic_name("display-brightness");
         this.set_applet_tooltip(DEFAULT_TOOLTIP);
         this.actor.connect('scroll-event', (...args) => this._onScrollEvent(...args));
+        this.lastTooltipTimeoutID = setTimeout(() => {}, 1);
 
         this.menu = new Applet.AppletPopupMenu(this, orientation);
         this.menuManager = new PopupMenu.PopupMenuManager(this);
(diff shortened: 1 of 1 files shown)


rcalixte:
cc @tim-we

ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
```diff
@@ -188,6 +195,46 @@ class DDCMultiMonitor extends Applet.IconApplet {
...
+        let direction = event.get_scroll_direction();
+        if (direction == Clutter.ScrollDirection.SMOOTH) { // Prevents being triggered twice
+            return
+        }
+
+        if (direction == Clutter.ScrollDirection.DOWN) {
+            this.monitors.forEach((monitor) => {
+                monitor.brightness = Math.max(0, monitor.brightness - BRIGHTNESS_ADJUSTMENT_STEP); // Decrease brightness by the step amount, making sure to not go below 0
+                // Tried making this into a function, it did not go well
+                this.set_applet_tooltip("Brightness: " + monitor.brightness + "%");
```
tim-we, line 210: > Tried making this into a function, it did not go well

What do you mean by this?

I don't think its a good idea to set this in the monitor loop as the tooltip will then show only the brightness of the last monitor. I suggest you use something like 
```js
const tooltipMessage = this.monitors.map(monitor => {
    monitor.brightness = /* ... */
   /* ... */
   return `${monitor.name}: ${monitor.brightness}%`;
}).join("\n");

to display all brightness values for all monitors.

**Comment by tim-we** [comment]
Unfortunately I cannot test this at the moment because I'm not using linux mint right now.

File: ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js [hunk 2]
```diff
@@ -188,6 +195,46 @@ class DDCMultiMonitor extends Applet.IconApplet {
...
+                // Tried making this into a function, it did not go well
+                this.set_applet_tooltip("Brightness: " + monitor.brightness + "%"); 
+                this._applet_tooltip.show();
+                setTimeout(() => { // Wait before actually changing the brightness, to prevent weirdness when scrolling multiple steps
+                    monitor.setBrightness(monitor.brightness); 
+                    setTimeout(() => {
+                        this._applet_tooltip.hide();
+                        this.set_applet_tooltip(DEFAULT_TOOLTIP);
+                    }, 1000);
+                }, 180);
```
tim-we, line 218: > to prevent weirdness when scrolling multiple steps

Could you elaborate on that? I know that ddc/ci behaves weird when you issue a new command before a previous one has completed. Maybe we should update `setBrightness` accordingly. It could store the promise in a private field of `Monitor` and append the new operation with `.then(...)`.
I'm only assuming thats the issue here and I think that would be a cleaner solution.

ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
GingerGigiCat: > What do you mean by this?

I should have been more specific, I meant all of lines 210 to 218. For some reason when I tried to put it into a function, the function would just not get called. It doesn't help that I am not great at javascript.

> I don't think its a good idea to set this in the monitor loop

I completely agree, thanks for the suggestion.
I'm not sure where or how exactly to add the code you suggested, but here works (at least for 1 display):

```js
if (direction == Clutter.ScrollDirection.DOWN) {
    let tooltipMessage = this.monitors.map(monitor => {
        monitor.brightness = Math.max(0, monitor.brightness - BRIGHTNESS_ADJUSTMENT_STEP);
        setTimeout(() => { // Prevent weirdness when scrolling multiple steps
            monitor.setBrightness(monitor.brightness);
        }, 200);
        return `${monitor.name}: ${monitor.brightness}%`;
    }).join("\n");

    this.set_applet_tooltip(tooltipMessage)
    this._applet_tooltip.show()
    setTimeout(() => {
        this._applet_tooltip.hide();
        this.set_applet_tooltip(DEFAULT_TOOLTIP);
    }, 2500);
    
}

else if (direction == Clutter.ScrollDirection.UP) {
    let tooltipMessage /* ... */
```

Should I push this into the pull request, or did you have something else in mind?

ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
GingerGigiCat: > Could you elaborate on that?

Without the pause, the screen brightness would jump around, often ending on the wrong brightness. I assume this is because of ddc/ci sending the brightness changes in a different order?

> Maybe we should update `setBrightness` accordingly. It could store the promise in a private field of `Monitor` and append the new operation with `.then(...)`.

While I did have to do some searching to figure out what that means, that's a much better idea than what I did!

My new code for the brightness setting:
```js
class Monitor {
    #promises;
    constructor(index, name, bus) {
        this.index = index;
        this.name = name;
        this.brightness = 50;
        this.bus = bus;
        this.menuLabel = null;
        this.menuSlider = null;
        this.#promises = Promise.resolve();
    }
```
...
```js
    setBrightness(value) {
        this.brightness = Math.round(value);
        this.updateMenu();
        this.#promises = this.#promises.then(() => {
            return new Promise((resolve, reject) => {
                Util.spawnCommandLineAsync(
                    `ddcutil --bus=${this.bus} setvcp 10 ${this.brightness}`,
                    resolve,
                    reject
                );
            });
        });
    }
```
...
```js
class DDCMultiMonitor extends Applet.IconApplet {
```
...
```js
        if (direction == Clutter.ScrollDirection.DOWN) {
            let tooltipMessage = this.monitors.map(monitor => {
                monitor.brightness = Math.max(0, monitor.brightness - BRIGHTNESS_ADJUSTMENT_STEP);
                monitor.setBrightness(monitor.brightness);
                return `${monitor.name}: ${monitor.brightness}%`;
            }).join("\n");

            this.set_applet_tooltip(tooltipMessage)
            this._applet_tooltip.show()
            setTimeout(() => {
                this._applet_tooltip.hide();
                this.set_applet_tooltip(DEFAULT_TOOLTIP);
            }, 2500);
            
        }
        else if (direction == Clutter.ScrollDirection.UP) {
```


Should I push this into the pull request?

tim-we: Yes, looks good to me. Only thing that we should change is the `setTimeout` thing for hiding the tooltip. If I understand correctly this will get scheduled for every scroll step. So if I scroll up and down for 2.5s weird things will start to happen. `setTimeout(...)` returns an id for that timeout which can be used to cancel the timeout. See https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout#examples
I would suggest to create a method for scheduling the hiding of the tooltip and this method could handle the clearing of the timeout as well as creating a new timeout. 

> While I did have to do some searching to figure out what that means

Sorry about that :D Anyway, I hope you learned something, I think promises are really cool and a handy tool for async stuff.

Thanks for your contribution to this applet!

tim-we: Also feel free to just commit & push these changes to this branch, instead of a long comment like this. That makes it easier to review.

ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
```diff
@@ -111,7 +117,9 @@ class DDCMultiMonitor extends Applet.IconApplet {
         super(orientation, panelHeight, instance_id);
         this.detecting = false;
         this.set_applet_icon_symbolic_name("display-brightness");
-        this.set_applet_tooltip("Adjust monitor brightness via DDC/CI");
+        this.set_applet_tooltip(DEFAULT_TOOLTIP);
+        this.actor.connect('scroll-event', (...args) => this._onScrollEvent(...args));
+        this.lastTooltipTimeoutID = setTimeout(() => {}, 1);
```
tim-we, line 122: You can initialize this with `undefined` or `null`, passing an invalid timeout ID to `clearTimeout` [does nothing; no exception is thrown](https://developer.mozilla.org/en-US/docs/Web/API/clearTimeout#notes).

tim-we (approved):

ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
GingerGigiCat: Oh interesting, I was expecting it to be like a promise in that way.

GingerGigiCat:
Are the changes good to be merged now? @tim-we

tim-we:
@GingerGigiCat looks good to me, but I don't have write permissions and cannot merge this. Also as I mentioned before, I no longer use Linux mint so I did not test this. But I assume you did, at least for 1 display, can you please confirm?

@rcalixte can you merge this?

GingerGigiCat:
Yes, I have tested for 1 display but I am not able to test it for multiple displays.

rcalixte:
@tim-we Should we update the maintainer for this? Would you like to still be contacted for future pull requests?

rcalixte (approved):

GingerGigiCat:
> If not, @GingerGigiCat would you like to be updated as the new maintainer for this applet?

If tim does not want to be the maintainer anymore, then I could step in as the new one, although I do not know the specifics of the ddc/ci mechanics so I may not be able to fix it if it breaks for any reason.

rcalixte:
> although I do not know the specifics of the ddc/ci mechanics so I may not be able to fix it if it breaks for any reason.

That makes both of us. 😅 

It would be a learning experience or someone else from the community could come forward with a solution. At the very least, having someone who uses the applet as the maintainer would be a step up in my (humble) opinion.

tim-we:
@rcalixte If you like, you can change the maintainer. I too think it makes sense to have someone that can at least test it. I'm happy to stay in the loop and get contacted for future PRs if thats possible. If I find the time I will still do a code review.

rcalixte:
> Edit: I see you already changed the maintainer. Will I still be contacted for future PRs?

#5862 hasn't been merged yet but I can do that now if it's okay. You won't be automatically tagged on future Issues or Pull Requests but we can still try to include you for future updates.

PR #5679, opened 2024-04-10
[sound150@claudiux] v6.7.3 - Fixes #5678

### LINKED ISSUE #5678: Media player list should be expanded as standard

sound150@claudiux @claudiux

I just installed your applet and it is working much better than the inbuilt sound applet. Thank you very much!
In both applets the list of active media players (forground or background sources) is kollapsed in main view.
So each time I want to change the source I first have to expand it to find the desired player.
I added 2 screenshots combined in 1 image:
sound150
It would be great to have the list expanded ALWAYS!
Thank you for your work, your answer and any improvement!

claudiux:
Fixes #5678.
Adds an option to keep open the player list in menu.

### DIFF
--- a/sound150@claudiux/files/sound150@claudiux/5.4/applet.js
+++ b/sound150@claudiux/files/sound150@claudiux/5.4/applet.js
@@ -2060,6 +2060,8 @@ class Sound150Applet extends Applet.TextIconApplet {
 
     on_applet_clicked(event) {
         this._openMenu();
+        if (this.settings.getValue("keepPlayerListOpen"))
+            this._launchPlayerItem.menu.open();
     }
 
     _openMenu() {
--- a/sound150@claudiux/files/sound150@claudiux/5.4/settings-schema.json
+++ b/sound150@claudiux/files/sound150@claudiux/5.4/settings-schema.json
@@ -42,6 +42,7 @@
       "type": "section",
       "title": "Menu",
       "keys": [
+        "keepPlayerListOpen",
         "playerControl",
         "extendedPlayerControl"
       ]
@@ -116,6 +117,11 @@
         "volume-down"
       ]
     }
+  },
+  "keepPlayerListOpen": {
+    "type": "switch",
+    "default": false,
+    "description": "Keep the player list open"
   },
   "playerControl": {
     "type": "switch",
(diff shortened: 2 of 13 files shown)


Dan-cer:
Thank you for the quick fix. Now the player list is open, and that is great!
My focus, however, was the list below of playback control. It is important to have that opened too because to switch the source I have to click into that list, because Youtube in browser isn't listed above:
sound150_1

claudiux:
This requires more development and debugging. Please be patient.

Dan-cer:
Thanks for your quick answer my friend! Yes I can imagine that and I'm very patient, I promise! 
A few days ago I read a post in which the author complained that the sound applet in Cinnamon does not display visible buttons for the various players. Now I have come to know and love your applet and am pleased that you want to provide exactly this feature. Good luck!

Dan-cer:
Hi claudiux, I think it's great that the button assignment setting within the applet settings ensures that they ALWAYS work. That's why it would be an advantage to also include the key assignment for "next track" and "previous track" in the applet settings. Because these keys no longer work for me, even though I have defined them in the Mint settings. The Mint settings were and are very often gotten inactive as soon as more than one media source is in play. I have known this for years.
Thank you!

claudiux:
You mean in the Shortcuts tab? You want me to add "next track" and "previous track"?

Dan-cer:
Yes, dear friend!

claudiux:
Can you open another Feature Request please?

Dan-cer:
**I now want to change my request, making it more simple:**
Why? Because today I learned the different intentions of the 2 sections I wanted to be merged. Clicking on any item in the upper section will start the selected app while the function of the section below is to select between media sources that are already started.
To merge them is therefore not an ideal solution. 
So I withdraw my request.
However, there seems to be a little issue left. As you see in my screenshot, Lollypop is displayed twice in the second section. Maybe because I started the app twice, I haven't found out yet.
**Could you just fix it to have only 1 instance of each active app** (may it a player or browser)?

Thank you very much in advance!
Kind regards!
sound-applet-new

claudiux:
When there is more than one way to control the same application, each way appears in the list; this is normal.

Dan-cer:
I checked it now:
1. with starting Lollypop its entry is displayed twice. Closing Lollypop removes both entries.
2. with starting YouTube in browser, Vivaldi is displayed twice. Immediately.
So it seems to be a bug.

claudiux:
There are two ways of controlling the mpv player here, even if mpv is not launched twice:

image

Dan-cer:
Ok. Meanwhile I found out, that the double entries are merging after a while (half a minute or so). As I mostly have a very short look on the applet I hadn't realized it before.
So it might be self-correction...
Finally, your fine applet need no fix?!
Have you an idea to exclude some players (Telegram...)?

PR #5690, opened 2024-04-12
directory-menu@torchipeppo: Add support for keybindings

rcalixte:
cc @torchipeppo @anaximeno 

Ref: #5668


anaximeno:
@rcalixte I tested it and it works well. 

I suggest that when clicking the keybinding a second time the menu (if it is open) should be closed, it might be a bit more intuitive compared with how keybinding works for most of the applets that come out of the box with Cinnamon.

rcalixte:
> I suggest that when clicking the keybinding a second time the menu (if it is open) should be closed, it might be a bit more intuitive compared with how keybinding works for most of the applets that come out of the box with Cinnamon.

I thought of that and tested a few things but the issue is getting it to be consistent but the menu is drawn by the Python script e.g. if the menu is opened via mouse click and then the keybinding shortcut is triggered, it's difficult to hook that with a proper callback. Given the target audience is likely to be power users (both installing and using the applet but also configuring a binding), I settled on having the Esc key be a consistent "cancel the menu" vector. Some consistency is better than none? 😅

I'm open to modifications if you can find something that works with every situation though.

anaximeno:
> I thought of that and tested a few things but the issue is getting it to be consistent but the menu is drawn by the Python script e.g. if the menu is opened via mouse click and then the keybinding shortcut is triggered, it's difficult to hook that with a proper callback. Given the target audience is likely to be power users (both installing and using the applet but also configuring a binding), I settled on having the Esc key be a consistent "cancel the menu" vector. Some consistency is better than none? 😅
> I'm open to modifications if you can find something that works with every situation though.
> (The construction of this applet is a bit of a hack but I'm largely fine because of the value that it provides. There are limitations that I'm willing to live with. 😅)

Ok, I'll give it a look into it, if I find something that works (if I don't I am also fine with how it is working currently, as it will be a significant improvement for some cases).

I also encountered an issue while testing regarding the setting of the keybinding as I described in the [pull request](https://github.com/rcalixte/cinnamon-spices-applets/pull/1) I sent in your repo with a possible solution.

torchipeppo (commented):
Sorry for the delay. Looks good, but I'm worried about what might happen if the keybind is pressed while another menu (whether GTK or Cinnamon) is open.
This is particularly relevant for Cinnamon panel menus, because attempting to open the Directory menu while a panel's context menu was open caused Cinnamon to freeze entirely (see applet.js lines 121-128).

anaximeno:
I tested for the case @torchipeppo mentioned before but didn't notice the issue, I guess it must've been because the focus goes to the popup menu after opening (and keybindings mostly work in relation to the focused window, in this case, the focused popup menu), so initially the issue wouldn't happen for most cases, I believe, though still, I am not sure it that will always be the behavior for the future, therefore still securing for could be a good idea, @rcalixte maybe the verification on the lines **122**:**128** should be moved (or by duplication adding it) to inside the **open_cassettone** function?
```js
if (global.menuStackLength) {
    // If we attempt to open this GTK menu while a Cinnamon panel menu is open,
    // Cinnamon will freeze.
    // This can happen with the panel's context menu (but not an applet's).
    // Returning is a simple fix, but it would be nicer (and riskier?) if it caused the open menu to close.
    return false;
}
```

rcalixte:
> @rcalixte maybe the verification on the lines **122**:**128** should be moved (or by duplication adding it) to inside the **open_cassettone** function?

I just rebased with this change. I still haven't had a chance to test it thoroughly but I hope to soon. Please shout if you or @torchipeppo have a chance to test/review this as well.

torchipeppo (approved):
So, I tested the keybind branch both with and without the additional menuStackLength check and in both cases Cinnamon stays alive when I press the keybind while another Cinnamon menu is open.

That said, I too think that keeping the extra check where it is makes more sense, since that is where the two activation paths (mouse and keybind) meet. It's probably safer against possible subtle changes to Cinnamon, too.

rcalixte:
Thank you both!

PR #5729, opened 2024-04-19
Fix prayer times appearance

NadhirBoukhenifra:
Set "expand" to "false" to address the issue of prayer times not being displayed correctly.


NadhirBoukhenifra:
Before:
before
After:
after

rcalixte:
Thank you!

NadhirBoukhenifra:
@rcalixte Glade to help.

PR #5961, opened 2024-05-01
[moonphase@techi-freki] Geolocation Features

Techi-Freki:
Adds geolocation popup
Adds moon rise, transit, and set times to popup
Adds phase info  to popup
Swaps out SunCalc for SunCalc3
Updates .pot file for new terms for localization
Advances version to 1.0.0


moonphase@techi-freki/files/moonphase@techi-freki/js/translator.js
```diff
@@ -0,0 +1,15 @@
...
+        this.uuid = uuid;
+    }
+
+    translate(str) {
+        GetText.bindtextdomain(this.uuid, `${get_home_dir()}/.local/share/locale`);
+        let translated = GetText.dgettext(this.uuid, str);
+        if (translated !== str) return translated;
+        return str;
+    }
+}
```
rcalixte, line 15: Using this as an internal class instead of the standard gettext invocation means that the default scripts will not pick up strings for translation properly which can pose a challenge for translators. You can see this if you run `./cinnamon-spices-makepot moonphase@techi-freki` on your system and check the output.

Techi-Freki: Thanks Rick! I wasn't aware of that script. I have removed that class and the script looks like it worked correctly.

rcalixte: Just one more quick note that you could make it global and only have to define it once. (Check other applets for an example here.) But this works and we can get this merged if you're ready.

rcalixte (approved):

PR #5975, opened 2024-05-03
cinnamon-spices-applets: Update readme

claudiux:
Adds how to determine the expiration date of a Linux Mint version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,8 @@ To ensure that these tables are always up-to-date, they are automatically regene
 # Javascript feature backward compatibility table
 ## (incomplete)
 To determine the backward compatibility of a javascript feature, look at it's page on [mdm web docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript) to see in which version of Firefox it was introduced. Then compare with the table below.
+To find out the expiration date of a Linux Mint version, take the release date of the major version (20 or 21 or 22, without dot) and add 5 to the year; the month is always April.
+Example: Major version 21 was released on July 31, 2022, so it expires in April 2027. Sub-versions 21.1, 21.2 and 21.3 also expire in April 2027.
 
 <table><tr><th colspan="1">Mint Version</th><th colspan="1">Codename</th><th colspan="1">Release Date</th><th colspan="1">Cinnamon<br/>Version</th><th colspan="2">Firefox version<br/>(CJS JS engine)</th></tr>
 <tr><td colspan="1">17</td><td colspan="1">Qiana</td><td colspan="1">31 May 14</td><td colspan="1">2.2.14</td><td colspan="1"></td></tr>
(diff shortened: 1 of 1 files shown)


rcalixte:
Just a few thoughts:
* This doesn't cover LMDE or other distros.
* Applets (and Spices) themselves are meant to work anywhere so it feels out of place here to call this out.
* Why this instead of the Wikipedia page for Mint which has the benefit of being maintained release by release?

I don't mean to heap things on so apologies if it feels that way.

claudiux:
>     * This doesn't cover LMDE or other distros.

I'm not the author of this table.
 
>     * Applets (and Spices) themselves are meant to work anywhere so it feels out of place here to call this out.

The authors of Spices are therefore obliged to maintain them for all versions of Cinnamon, even those that have expired?

>     * Why this instead of the Wikipedia page for Mint which has the benefit of being maintained release by release?

We should at least provide a link to this Wikipedia page.

> I don't mean to heap things on so apologies if it feels that way.

Well, one time more or less... I don't feel like offering anything anymore.

fredcw:
I think it would best to just mark the unsupported mint versions in the table. I'm at work atm, but I can do this later. I guess it's up to applet maintainers if they want to support older versions.

rcalixte:
> I think it would best to just mark the unsupported mint versions in the table. I'm at work atm, but I can do this later. I guess it's up to applet maintainers if they want to support older versions.

Are the first two columns even needed? What matters is the version of Cinnamon and the underlying JS engine right? It would also make it easier to maintain long-term. (I also think the Cinnamon Version column only needs the major and minor.)

fredcw:
@rcalixte I think the mint version is useful, codename not so much. But in any case, I think it's handy to have this here as the only other place it's documented is wikipedia. The shipped stable cinnamon version is not documented anywhere else afaik.

rcalixte:
> @rcalixte I think the mint version is useful, codename not so much. But in any case, I think it's handy to have this here as the only other place it's documented is wikipedia. The shipped stable cinnamon version is not documented anywhere else afaik.

Wouldn't someone looking for this info on their system check the System Info dialog? Granted, all of that information is there, but the Cinnamon version is what matters, not the distro. (I have been looking at this through the lens of the table being useful to anyone using Cinnamon, not necessarily just Mint users. Let me know if that's incorrect since I think that's the reason we see this slightly differently.)

rcalixte:
I'm thinking that we only need three columns here. (Correct me if I'm wrong.)

|Cinnamon<br />Version|Release<br />Date|CJS<br />Version|
|:-:|:-:|:-:|
|6.0.x|12 Jan 24|102|

fredcw:
@rcalixte Actually, since I made it, I've barely used it for its stated purpose. I use it quite often to see which cinnamon version goes with which Mint version. For instance, when communicating with users, I find it better to write "cinnamon 4.8 (Mint 21.1)" rather than just the cinnamon version and this information isn't really available anywhere else. In any case, I don't see the harm in keeping this info here in case anyone finds it useful.

PR #5984, opened 2024-05-07
Added password-generator@spencerlommel.com

SpencerLommel:


password-generator@spencerlommel.com/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+Using a terminal, you can do the following sequence of commands:
+
+```
+git clone https://github.com/SpencerLommel/cinnamon_password_generator.git
+mkdir -p ~/.local/share/cinnamon/applets/password-generator@spencerlommel.com/ && mv cinnamon_password_generator/* ~/.local/share/cinnamon/applets/password-generator@spencerlommel.com/
+rm -rf cinnamon_password_generator
+```
+
+Then press `Alt`+`F2`, type `r` in the dialog window, then press `Enter`.
```
rcalixte, line 12: Once published, these instructions will be more difficult than necessary as users will be able to install from their systems. Take a look at other applets for an example.

rcalixte:
TWo quick notes: first, you will want to run `./cinnamon-spices-makepot password-generator@spencerlommel.com` to generate the translation template that translators can use to translate this applet to other languages. Second, while it isn't mandatory, the standard ID for an applet is name@github_author. You can keep the domain if you would like but it is just a heads up for you to consider if you would like.

password-generator@spencerlommel.com/README.md
SpencerLommel: That makes sense, I just copied the one from before I planned on adding it to the store so I forgot to change it haha. Will fix that now

anaximeno (changes requested):
Hi @SpencerLommel 

You'll also probably need to add these lines above your applet's class to define the function `_` used for translation (use your applet's UUID there):

https://github.com/linuxmint/cinnamon-spices-applets/blob/74c7ca93e6cfe64b2b73ab2c54a26173f2f8620e/Exit%40claudiux/files/Exit%40claudiux/5.4/applet.js#L15-L20

After making this change and the comments I added below, run the command @rcalixte suggested above to (re)generate the translation file.

password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
```diff
@@ -0,0 +1,77 @@
...
+        this.set_applet_tooltip("Generate Random Password");
+
+        this.menu = new Applet.AppletPopupMenu(this, orientation);
+        this.menuManager = new PopupMenu.PopupMenuManager(this);
+        this.menuManager.addMenu(this.menu);
+
+        this.menuBox = new St.BoxLayout({ vertical: false });
+        this.textBox = new St.Entry({
+            style_class: 'popup-menu-item',
+            hint_text: "Please enable at least one value in settings.",
```
anaximeno, line 28: Here should be set for translation basically call a function named `_`, this way `_("Text that will be displayed to the user and should be translated")`

password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
```diff
@@ -0,0 +1,77 @@
...
+
+        // Bind settings
+        this.settings.bind("password-length", "passwordLength", this.onSettingsChanged);
+        this.settings.bind("include-lowercase", "includeLowercase", this.onSettingsChanged);
+        this.settings.bind("include-uppercase", "includeUppercase", this.onSettingsChanged);
+        this.settings.bind("include-symbols", "includeSymbols", this.onSettingsChanged);
+        this.settings.bind("include-numbers", "includeNumbers", this.onSettingsChanged);
+
+        this.set_applet_icon_name("dialog-password");
+        this.set_applet_tooltip("Generate Random Password");
```
anaximeno, line 19: Here should be set for translation basically call a function named `_`, this way `_("Text that will be displayed to the user and should be translated")`

anaximeno (approved):

rcalixte (approved):

rcalixte:
Thank you for your contribution!

PR #5991, opened 2024-05-08
Auto copy setting, sound feedback, reorganized settings

SpencerLommel:
- Reorganized settings into password settings and copy settings
- Added a feature that auto copies the password to the clipboard when then icon is clicked (off by default)
- Added a checkbox to play a sound when the password is copied to the clipboard
- Added a soundfilechooser to change the sound that is played when the password is copied
- Ran makepot so new settings should be set for translations

### DIFF
--- a/password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
+++ b/password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
@@ -23,6 +23,8 @@ class CinnamonRandomPasswordApplet extends Applet.IconApplet {
         this.settings.bind("include-uppercase", "includeUppercase", this.onSettingsChanged);
         this.settings.bind("include-symbols", "includeSymbols", this.onSettingsChanged);
         this.settings.bind("include-numbers", "includeNumbers", this.onSettingsChanged);
+        this.settings.bind("auto-copy", "autoCopy", this.onSettingsChanged);
+
 
         this.set_applet_icon_name("dialog-password");
         this.set_applet_tooltip(_("Generate Random Password"));
@@ -57,10 +59,15 @@ class CinnamonRandomPasswordApplet extends Applet.IconApplet {
     }
 
     on_applet_clicked(event) {
-        this.menu.toggle();
         let newPassword = this.generateRandomPassword();
         this.textBox.set_text(newPassword);
+        if (this.autoCopy) {
+            St.Clipboard.get_default().set_text(St.ClipboardType.CLIPBOARD, newPassword);
+        } else {
+            this.menu.toggle();
+        }
     }
+    
 
     generateRandomPassword() {
         let charset = "";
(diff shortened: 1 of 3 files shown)


password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
```diff
@@ -51,17 +56,38 @@ class CinnamonRandomPasswordApplet extends Applet.IconApplet {
...
+        } else {
+            this.menu.toggle();
+        }
+    }
+
+    playClickSound() {
+        if (this.playSound && this.soundFilePath) {
+            let file = Gio.File.new_for_path(this.soundFilePath);
+            let player = new Gio.Subprocess({
+                argv: ['paplay', file.get_path()],
```
rcalixte, line 82: Just a heads up that I believe this doesn't support .mp3 files. It's up to you if you want to implement this differently or to update the README so that users are aware.

SpencerLommel: I attempted to use GSound and ran into some issues so I went with this route. My reasoning was that all of the linux mint system sounds are ogg sounds so that's what I wanted to primarily support

password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/settings-schema.json
```diff
@@ -27,5 +34,28 @@
...
+    },
+    "play-sound": {
+        "type": "checkbox",
+        "default": true,
+        "description": "Play sound on copy",
+        "tooltip": "Toggles whether or not a sound is played when the password is copied to the Clipboard"
+    },
+    "sound-file": {
+        "type": "soundfilechooser",
+        "default": "/usr/share/mint-artwork/sounds/map.oga",
```
rcalixte, line 56: This will only be available on systems with the mint-artwork package installed, which will mostly just be Mint systems. It might be best to default to one of the freedesktop sounds so that it will work as many places as possible by default, i.e. files under `/usr/share/sounds/freedesktop/stereo/`.

password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/applet.js
rcalixte: Ah, that ties into the other comment. 😅

password-generator@spencerlommel.com/files/password-generator@spencerlommel.com/settings-schema.json
rcalixte: One option you have is including the sound file you want with the applet itself and then updating the path. (If you search this applets repository, you'll see other .oga and .mp3 files throughout.)

SpencerLommel: I'll have to do the latter option and just include it in the package.

rcalixte: Shout if you need a hand with the proper path and whatnot.

rcalixte: Just a note that I don't have map.oga on my system so it may be better to include the audio file that you prefer to be the default so that it works universally out of the box. I'll go ahead and merge this for now but be mindful that bug is still out there.

rcalixte (approved):

PR #5996, opened 2024-05-08
Make cinnamon-spices-makepot faster

Gr3q:
Was too slow processing each language 1 by 1 (a few seconds...), so now it uses all threads by default.

Made sure the process outputs are grouped and printed properly, just not in outputted stderr now, but stdout. I can print them to stderr again if you want it.


rcalixte:
> just not in outputted stderr now, but stdout. I can print them to stderr again if you want it.

Ah yes, we do want it in stderr since it should run in a terminal (and notify the person running it) but also for future automation purposes (when it's automated... someday 😅).

Everything else looks great and knocks another to-do off my list! Thank you!

Gr3q:
sure, I'll do that this evening

Gr3q:
Was a bit more complicated than I though it would be, but it's done.

rcalixte:
> Was a bit more complicated than I though it would be, but it's done.

Agreed on the complication. I'll test in a bit. Thanks again!

Gr3q:
Well yeah I didn't test it with all...I'll see what it does then.

rcalixte:
My main concern in trying to work with this is that it is significantly more complex to troubleshoot and maintain. I see the value in this when there are a multitude of translations such as the weather applet that you maintain but for the gain of a few seconds, we would be introducing a significant amount of complexity (concurrency in Python isn't pleasant).

Gr3q:
The concurrency part is the simple bit, keeping the order of stderr/stdout for each command to rplay it is the complicated part unfortunately. If that is not needed it's minimal changes.

But I understand what you mean.

Gr3q:
I had some time to look at this today, and I know why it broke. I'm also extremely concerned the usages of `os.chdir` in this script, especially when a subprocess can be ran in the current cwd isolated.

rcalixte:
> I had some time to look at this today, and I know why it broke. I'm extremely concerned the usages of `os.chdir` in this script, especially when a subprocess can be ran in the correct cwd isolated.

It makes it tricky indeed to implement threading but either way, you would still be dealing with paths and whatnot right? That is the core challenge but if you think it can be done in a cleaner way, go for it.

Gr3q:
It works now, `make_pot` doesn't need to know if it's running with `--all` either. Well, you decide if it's worth it, I tried to add useful comments and stuff so it's easier to review

rcalixte:
> It works now, `make_pot` doesn't need to know if it's running with `--all` either. Well, you decide if it's worth it, I tried to add useful comments and stuff so it's easier to review

This work is absolutely not in vain. I think this is a welcome change and I plan on porting it to the other repositories once we get it where we want. That is why it needs to be maintainable and portable. Your contributions are always appreciated.

rcalixte:
@Gr3q I just pushed up the revised changes to incorporate your improvements but also support readability, maintainability, and portability. The timed executions are on the same order as well. Hopefully you're happy with the final result. I think this is ready to merge and be ported to the other Spices.

rcalixte (approved):

PR #6088, opened 2024-06-15
c-eyes@anaximeno: Version 2.1.3

anaximeno:
- Add information about the mouse click effects split in settings
- Backport some fixes to the 5.4 and 4.2 versions of the applet
- Update translations

### DIFF
--- a/c-eyes@anaximeno/files/c-eyes@anaximeno/6.2/applet.js
+++ b/c-eyes@anaximeno/files/c-eyes@anaximeno/6.2/applet.js
@@ -41,6 +41,11 @@ function _(text) {
 }
 
 
+// Mark this for translation since it won't be
+// marked by the translation tool
+const _DEFAULT_TOOLTIP = _("Hey, I saw that!");
+
+
 class Eye extends Applet.Applet {
 	constructor(metadata, orientation, panelHeight, instanceId, areaWidth) {
 		super(orientation, panelHeight, instanceId);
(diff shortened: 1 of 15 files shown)


c-eyes@anaximeno/files/c-eyes@anaximeno/6.2/applet.js
```diff
@@ -41,6 +41,11 @@ function _(text) {
 }
 
 
+// Mark this for translation since it won't be
+// marked by the translation tool
+const _DEFAULT_TOOLTIP = _("Hey, I saw that!");
```
rcalixte, line 46: Maybe we can configure better string detection in the settings file. This is a fine workaround for now though.

rcalixte (approved):

c-eyes@anaximeno/files/c-eyes@anaximeno/6.2/applet.js
anaximeno: Yes, I was thinking about it too, because if you set the line for translation on the pot file it will disappear when running the makepot script. I was wondering if it was possible to add a marker above the text on the pot file signaling for it not to be removed even if the text is not found on the codes.

rcalixte: We don't need to fix it right away but long-term, I think we can fix the detection rather than encourage manual string placements.

PR #6103, opened 2024-06-20
Weather, Add line break {br} for Override label and Override tooltip

aperkins81:
Fixes \n, \\\\n or %n not inserting a line break in the Override label and Override tooltip.
Screenshot from 2024-06-20 14-52-40


rcalixte:
cc @Gr3q

Gr3q (approved):
While Shift+Enter or pasting finished text in with line breaks works, I can understand that those are not particularly or convenient methods.

rcalixte (approved):

PR #6131, opened 2024-06-28
Updated terminal app.

### LINKED ISSUE #6126: [qredshift@quintao] v2.0.0 doesn't work
### Applet version/Build date

Version 2.0.0 (2024-06-28 08:30:35)

### Cinnamon version

Cinnamon 6.0.4

### Distribution

Mint 21.3

Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller

qredshift@quintao @raphaelquintao

### What happened?

I've removed redshift, also entirely removed the old version of this applet (even config files).
Then I've installed the v2.0.0 of this applet and set all settings.
Nothing happens when I activate/deactivate this applet.

My `~/.xsession-errors` contains:
```
(cinnamon:2292): GLib-CRITICAL **: 14:39:07.442: g_child_watch_add_full: assertion 'pid > 0' failed
Gjs-Message: 14:39:07.445: JS LOG: [LookingGlass/warning] 
QRedshift: Created

(cinnamon:2292): GLib-CRITICAL **: 14:39:07.472: g_child_watch_add_full: assertion 'pid > 0' failed

(cinnamon:2292): Gjs-WARNING **: 14:39:07.483: Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to use TextDecoder.
(Note that array.toString() may have been called implicitly.)
0 _parse() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":103:12]
1 spawn_command_line_async_promise/<() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":125:31]
2 spawn_command_line_async_promise() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":110:11]
3 QRedshift() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":320:15]
4 main() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":1073:11]
5 createApplet() ["/usr/share/cinnamon/js/ui/appletManager.js":597:24]
6 addAppletToPanels() ["/usr/share/cinnamon/js/ui/appletManager.js":372:33]
7 finishExtensionLoad() ["/usr/share/cinnamon/js/ui/appletManager.js":98:13]
8 _init/<() ["/usr/share/cinnamon/js/ui/extension.js":286:31]

Gjs-Message: 14:39:07.488: JS LOG: [LookingGlass/warning] 
QRedshift: ADDED TO PANEL
Gjs-Message: 14:39:07.495: JS LOG: [LookingGlass/info] Loaded applet qredshift@quintao in 128 ms
Gjs-Message: 14:39:07.495: JS LOG: [LookingGlass/warning] 
ASYNC LOADING: qredshift 0.10

Gjs-Message: 14:39:07.495: JS LOG: [LookingGlass/warning] 
UPDATE: true

(cinnamon:2292): GLib-CRITICAL **: 14:39:07.519: g_child_watch_add_full: assertion 'pid > 0' failed
```

Later:
```
(cinnamon:2292): GLib-CRITICAL **: 14:39:19.753: g_child_watch_add_full: assertion 'pid > 0' failed
Gjs-Message: 14:39:19.820: JS LOG: [LookingGlass/warning] 
UPDATE: true

(cinnamon:2292): GLib-CRITICAL **: 14:39:19.835: g_child_watch_add_full: assertion 'pid > 0' failed
Gjs-Message: 14:39:19.843: JS WARNING: [/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function 696]: Too many arguments to method Clutter.InputDevice.ungrab: expected 0, got 1
Gjs-Message: 14:39:20.464: JS LOG: [LookingGlass/warning] 
UPDATE: true
```

I hope you'll can solve this issue. Your applet saves my eyes! Thank you!

### Other information

I'm using X11, not wayland.

raphaelquintao:
Fixes #6126


rcalixte:
@raphaelquintao Apologies, but we're going to revert the earlier pull request to revisit this binary. There are a few issues, ranging from the source not being available, to the fact that it will not be cross-architecture compatible. Would it be possible to include the source and then have it compile locally on each installation? (Check to see if the binary exists and if it does not, compile from the bundled source.) Thanks for your contribution! Hopefully we can get this resolved quickly.

raphaelquintao:
It will probably work on anything that use X, its almost like the old redshift.
Compile that in every machine i think its not a good idea, people will need to install gcc and some dev libs. It will be a bad user experience.
The problem we had today, was that i was setting the gamma ramps on all outputs including the disconnected ones, and for some reason my hardware accept that and some other don't, this version is updated.

rcalixte:
> @rcalixte It will probably work on anything that use X, its almost like the old redshift.

Right, but as a binary, it will only work for one architecture, which we can only assume to be x86_64 since we can't verify it with the source. Cinnamon is used cross-architecture.



> Compile that in every machine i think its not a good idea, people will need to install gcc and some dev libs. It will be a bad user experience.

What can we do to optimize this process? Shipping a binary in a Spice is not ideal. Can we keep the existing system/repository dependencies otherwise?

raphaelquintao:
We can ship the binary for different architecture, not problem about that.
The reason behind the idea of have our own binary is for future wayland support.
Redshift in no longer maintained and limits the applet possibilities.

rcalixte:
> We can ship the binary for different architecture, not problem about that (I hadn't thought about that, good point).

This is still not a scalable solution and it doesn't address the problem of the available source.


> The reason behind the idea of have our own binary is for future wayland support.

There is already a Wayland protocol for that and if there is going to be something implemented for it, it should be done within Cinnamon.



> Redshift in no longer maintained and limits the applet possibilities.

Who can maintain a binary without the source?

raphaelquintao:
What you means with a binary without the source?

rcalixte:
> What you means with a binary without the source?

Like this pull request. It was originally to update a binary to fix an issue. There is no source for it though. How would anyone else fix this issue?

raphaelquintao:
The code is open source [https://github.com/raphaelquintao/QRedshift](https://github.com/raphaelquintao/QRedshift).

raphaelquintao:
I mentioned that in the [Version 2.0.0 pull request](https://github.com/linuxmint/cinnamon-spices-applets/pull/6124) , its also on the readme.

rcalixte:
> The code is open source https://github.com/raphaelquintao/QRedshift.

There isn't a connection from the repository to the binary since the code and compilation are not included here.



> I mentioned that in the [Version 2.0.0 pull request](https://github.com/linuxmint/cinnamon-spices-applets/pull/6124) , its also on the readme.

This should not have been merged without further discussion. This is why it was reverted.

raphaelquintao:
OK. So keep the old version. I will not support that anymore.

rcalixte:
> OK. So keep the old version. I will not support that anymore.

The new version looked phenomenal. We need to implement a better way to either provide the binary or continue using the previous dependencies that are provided by distributions.

PR #6145, opened 2024-07-01
Set executable permissions (required for Cinnamon 6.2+)

rcalixte:
Ref: linuxmint/cinnamon#12174


fredcw:
@rcalixte Omg, I can't believe all these applets were effectively broken all this time and nobody knew. I assumed that because xlets could be downloaded as zip files and extracted manually from the spices website, that this would have been done by at least a few people and any problems with file permissions would have already been detected long ago. Thanks for doing this, it must have been pretty painstaking. I would have done it myself if I had known that existing file permissions were going to be a problem.

rcalixte:
> @rcalixte Omg, I can't believe all these applets were effectively broken all this time and nobody knew. I assumed that because xlets could be downloaded as zip files and extracted manually from the spices website, that this would have been done by at least a few people and any problems with file permissions would have already been detected long ago. Thanks for doing this, it must have been pretty painstaking. I would have done it myself if I had known that existing file permissions were going to be a problem.

It was on my radar for a while but not urgent until Cinnamon was tagged (and therefore likely to end up in Arch-based distros). There aren't that many nightly repos or PPAs that I'm aware of either which is why it wasn't urgent. There are also a decent amount of people who don't use any Spices. And it's no problem! I'm glad to have it done. It may not be perfect but nothing should be broken on the latest updates. Shout if anything is off! Thanks for the original commit that made this possible!

PR #6191, opened 2024-07-07
workspace-grid@hernejj: Added the ability to change the color of the grid squares for the user interface in the applet

mitchellkolb:
Author: Jason J. Herne (hernejj@gmail.com) 

My changes only effects the workspace-grid@hernejj applet in particular the 5.6 version. I adjusted 3 files and made the default color still be what is it is now which is blue, but the user can change what color the grid squares are by going into the applet settings page and pasting the hex color value in for what color they want.

@hernejj, please review the changes. Let me know if there are any improvements or adjustments I need to make.

Here are some images of how the applet looks/functions with my adjustments:
original-grid-applet
color-settings
red-grid-applet

### DIFF
--- a/workspace-grid@hernejj/files/workspace-grid@hernejj/5.6/BarIndicatorStyle.js
+++ b/workspace-grid@hernejj/files/workspace-grid@hernejj/5.6/BarIndicatorStyle.js
@@ -2,15 +2,16 @@ const Lang = imports.lang;
 const St = imports.gi.St;
 const Clutter = imports.gi.Clutter;
 
-function BarIndicatorStyle(applet, cols, rows, height) {
-    this._init(applet, cols, rows, height);
+function BarIndicatorStyle(applet, cols, rows, height, uiColor) {
+    this._init(applet, cols, rows, height, uiColor);
 }
 
 BarIndicatorStyle.prototype = {
 
-    _init: function (applet, cols, rows, height) {
+    _init: function (applet, cols, rows, height, uiColor) {
         this.applet = applet;
         this.button = [];
+        this.uiColor = uiColor;
         this.update_grid(cols, rows, height);
         this.switch_id = global.window_manager.connect('switch-workspace', Lang.bind(this, this.update));
         this.scroll_id = this.applet.actor.connect('scroll-event', Lang.bind(this, this.onMouseScroll));
@@ -153,10 +154,12 @@ BarIndicatorStyle.prototype = {
             if (i == active_ws) {
                 this.button[i].get_child().set_text((i + 1).toString());
                 this.button[i].add_style_pseudo_class('outlined');
+                this.button[i].style = `background-color: ${this.uiColor};`;
             }
             else {
                 this.button[i].get_child().set_text((i + 1).toString());
                 this.button[i].remove_style_pseudo_class('outlined');
+                this.button[i].style = `border: 1px solid ${this.uiColor}; opacity: 1;`;
             }
         }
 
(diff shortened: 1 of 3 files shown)


rcalixte:
> My changes only effects the workspace-grid@hernejj applet in particular the 5.6 version.

This affects every version after Cinnamon 5.6 until now (until a new verison folder is created).




> but the user can change what color the grid squares are by going into the applet settings page and pasting the hex color value in for what color they want.

Cinnamon already has settings for a color chooser that also supports alpha. Why not use that?



> Here are some images of how the applet looks/functions with my adjustments:

Have you tested the other grid layouts? Something like a 2x2 or 3x3 configuration.

mitchellkolb:
> > but the user can change what color the grid squares are by going into the applet settings page and pasting the hex color value in for what color they want.

> Cinnamon already has settings for a color chooser that also supports alpha. Why not use that? 

I  don't think I know what you mean with "settings for a color chooser in cinnamon". I took a look in the system settings in the applet, theme, and color categories and don't see an option that relates to changing the color of applets, apart from just changing my icon theme, which changes a lot more than what I'm looking for.


I think fundamentally, what I'm trying to achieve is the ability to adjust the color of this specific applet separately from the color dictated by my selected icon theme in system settings. I initially thought it would be easier to give this applet its own color adjustment option because that's what I'm looking to solve. However, if there is a setting that allows users to adjust the visual appearance of any applet, and I can apply that to this one, then it would achieve my goal. I would love some more details if that's a thing. 


> This should also be an option that can be toggled as plenty of themes supply settings for this applet. This is a great option but it should be able to be turned off as well.

> Have you tested the other grid layouts? Something like a 2x2 or 3x3 configuration.

If the color setting mentioned above doesn't work as expected and I continue with this pull request, I will need to commit additional changes. I believe your idea of defaulting to the user's theme and allowing custom colors if enabled in the settings is excellent. I will also need to adjust it to work with the grid-layout appearance, as it currently only supports the bar-indicator layout.


> Shout if you need a hand and we can walk through this!

Thank you for this, as you can probably see this is my first contribution to cinnamon applets and I appreciate your comments.

rcalixte:
> I don't think I know what you mean with "settings for a color chooser in cinnamon". I took a look in the system settings in the applet, theme, and color categories and don't see an option that relates to changing the color of applets, apart from just changing my icon theme, which changes a lot more than what I'm looking for.
> 
> I think fundamentally, what I'm trying to achieve is the ability to adjust the color of this specific applet separately from the color dictated by my selected icon theme in system settings. I initially thought it would be easier to give this applet its own color adjustment option because that's what I'm looking to solve. However, if there is a setting that allows users to adjust the visual appearance of any applet, and I can apply that to this one, then it would achieve my goal. I would love some more details if that's a thing.

Not a setting for the applet itself, that would still need an implementation similar to what you've already accomplished. The method of input shouldn't be a manual hex that doesn't support alpha. [Here](https://github.com/linuxmint/cinnamon-spices-applets/blob/2bd410f/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/5.4/settings-schema.json#L245-L255) is an example from another applet. The user interface is a color chooser and the values can be [accessed via properties](https://github.com/linuxmint/cinnamon-spices-applets/blob/2bd410f/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/5.4/applet.js#L516-L521). You can have a boolean toggle as a dependency that determines whether or not the color chooser is available and displayed, another [example from that same applet](https://github.com/linuxmint/cinnamon-spices-applets/blob/2bd410f/batterymonitor%40pdcurtis/files/batterymonitor%40pdcurtis/5.4/settings-schema.json#L217).

PR #6198, opened 2024-07-10
Announcement

raphaelquintao:


claudiux:
The `install.sh` script leaves an annoying archive file in `~/.local/share/cinnamon/applets`.

raphaelquintao:
I fixed that [here](https://github.com/raphaelquintao/QRedshiftCinnamon/blob/master/install.sh#L66).
You can test.

claudiux:
That works well now. Thank you!

fredcw:
@raphaelquintao Can't your QRedshift program be rewritten in python?

raphaelquintao:
Maybe, i need to research about this. 
However, the native and best way to do that is to use the native Linux library. The way I wrote it doesn't rely on any library.
The architecture problem i will fix, i will provide a binary for all architectures supported by gcc and the xorg library. 
I just need a little time to set up the environment to compile the binaries.

clefebvre:
We need to chat about this. If this is no longer "actively maintained" it will get deleted in cinnamon-spices.

clefebvre:
I understand it's a restriction to not being able to ship binaries or have this work out of the box, but the binaries represent a security issue for users. It creates a precedent which is dangerous if generalized to spices maintainers.

raphaelquintao:
I initially created this applet for my personal use and i used it a lot before deciding to share it here for free, simply for fun. To my surprise, many people have enjoyed it, and I have made an effort to keep it running and updated for almost five years.

The decision to remove the dependency on the old Redshift was made to make the applet more standalone. Redshift has many features that are not necessary for this applet, and its core functionality is easy to implement. Therefore, I developed a minimalist and also opensource C application to integrate into the applet, making it almost dependency-free.

However, if you choose to remove it, I completely understand. I will continue using it and developing it in my own way.

rcalixte:
@raphaelquintao You might be interested in [this](https://github.com/linuxmint/cinnamon-spices-applets/blob/b2442635ca8d95a5abb0feaccc8b80a79e0daef5/auto-dark-light%40gihaume/files/auto-dark-light%40gihaume/lib/time_change_listener/time_change_listener.js) example code from another currently proposed applet. This includes the code, build, and compilation steps in an automated fashion. It may take a little bit to implement but would this be a good workaround for you? Take a look at the overall structure of the applet as well (hopefully it will be merged soon so that it is easier to review) [here](https://github.com/linuxmint/cinnamon-spices-applets/tree/b2442635ca8d95a5abb0feaccc8b80a79e0daef5/auto-dark-light%40gihaume/files/auto-dark-light%40gihaume/lib/time_change_listener) as it suggests one possible directory structure as well.

PR #6200, opened 2024-07-10
[CustomApplicationsMenu@LLOBERA] store preferences in applet settings

alexkuz:
Use applet settings system to store user preferences.

Currently, settings are stored in `applications.json`, that gets overridden every time the applet gets updates.
This PR should fix it.

image


rcalixte:
cc @Nicolas01

CustomApplicationsMenu@LLOBERA/files/CustomApplicationsMenu@LLOBERA/TextWidget.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/python3
+
+import random
+from JsonSettingsWidgets import *
+from gi.repository import Gio, Gtk
```
rcalixte, line 5:
```suggestion

import gi
gi.require_version("Gtk", "3.0")
from gi.repository import Gio, Gtk
```

fredcw:
This is an improvement but as you're editing the json file, ~/.xsession-errors is spammed a *lot*, especially if there's already an invalid command.

Edit: Also, the readme.md file will need updating since the instructions would be wrong.

PR #6220, opened 2024-07-20
Cinnamenu@json: Set clip_to_allocation to true

### LINKED ISSUE #6219: Cinnamenu: Buttons that have been scrolled outside the menu can be activated
### Applet version/Build date

5.4.13

### Cinnamon version

6.2.2

### Distribution

Mint 22 BETA

NVIDIA GeForce GTX and xserver-xorg-video-nouveau

Cinnamenu@json @fredcw

### What happened?

When the menu is open, if you scroll down the application grid/list then click above the menu, it will launch the app that would be there if the menu was that large, despite the cursor not being within the menu. Hovering the cursor above the menu will also show tooltips if they are turned on.
As for the category list, if there are enough categories and you scroll down until some are outside the menu, clicking/hovering the cursor above the menu will switch the application grid/list to that category.

The expected behaviour is that hovering anywhere outside the menu should show nothing and clicking anywhere outside the menu should close it.

### Other information

Possibly related to a similar bug that occurs with the default menu applet: https://github.com/linuxmint/mint22-beta/issues/198

rcalixte:
Closes #6219
Ref: linuxmint/cinnamon#11123

cc @fredcw


fredcw:
Looks good. I can't find any side effects.

PR #6226, opened 2024-07-21
ddcci-multi-monitor@tim-we: bugfix

BigIskander:
ddcci-multi-monitor@tim-we is not working after last update, gives an error Glib not found. Fixed this error.


rcalixte (approved):

rcalixte:
Thank you for the fix!

PR #6240, opened 2024-07-24
Update recorder.sh

Hyakki09:
This change makes it work with pipewire.

### DIFF
--- a/ScreenShot+RecordDesktop@tech71/files/ScreenShot+RecordDesktop@tech71/recorder.sh
+++ b/ScreenShot+RecordDesktop@tech71/files/ScreenShot+RecordDesktop@tech71/recorder.sh
@@ -40,9 +40,9 @@ record_window() {
 
 record_audio() {
   parameters+=(
-    -f alsa
+    -f pulse
     -ac 2
-    -i pulse
+    -i default
   )
 }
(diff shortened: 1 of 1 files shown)


rcalixte:
@Hyakki09 Have you tested this on older versions where Pipewire is not the default and therefore possibly not installed?

Hyakki09:
> @Hyakki09 Have you tested this on older versions where Pipewire is not the default and therefore possibly not installed?

Oopps My bad, didn't test it on older version without pipewire.

rcalixte:
> Oopps My bad, didn't test it on older version without pipewire.

It's fine. If it doesn't work. We can make two versioned folders and that will resolve the problem. It's just good to know what we need to do so that we don't break the applet for users currently using it.

Hyakki09:
> > Oopps My bad, didn't test it on older version without pipewire.
> 
> It's fine. If it doesn't work. We can make two versioned folders and that will resolve the problem. It's just good to know what we need to do so that we don't break the applet for users currently using it.

Yes, and this makes it work with linux mint 22.

rcalixte:
> Yes, and this makes it work with linux mint 22.

Have you had a chance to check against previous versions?

PR #6272, opened 2024-07-31
[betterlock] Correct spelling in betterlock settings

rlatosky:
Fixed spelling mistake in betterlock settings. Tested changes on Mint 22 in a VM.

### DIFF
--- a/betterlock/files/betterlock/po/betterlock.pot
+++ b/betterlock/files/betterlock/po/betterlock.pot
@@ -50,6 +50,9 @@ msgstr ""
 msgid "Show notifications"
 msgstr ""
 
+msgid "Silence system notification sound"
+msgstr ""
+
 msgid "Show caps lock indicator"
 msgstr ""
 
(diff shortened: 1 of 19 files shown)


rcalixte:
@rlatosky Did you test this with other languages? This change will break translations for this string.

rlatosky:
@rcalixte You're right, I forgot about that. I grabbed the latest version from this repo, and tested using a different language, but, as far as I can tell, the setting "Silence system notification sound" doesn't seem to be translated by any of the .po translation files.

betterlock/files/betterlock/po/betterlock.pot
```diff
@@ -50,6 +50,9 @@ msgstr ""
 msgid "Show notifications"
 msgstr ""
 
+msgid "Silence system notification sound"
+msgstr ""
+
```
rcalixte, line 55: There is a script precisely so that these files do not need to be modified manually. If you look in the README, it is explained there. For now, I'm going to run `./cinnamon-spices-makepot betterlock` so that this can be merged.

rlatosky: Got it, I'll keep that in mind for the future. Thank you!

PR #6273, opened 2024-07-31
feeds@jonbrettdev.wordpress.com: Fix resizing when adding new feeds

philipp-meier:
When adding a new feed, the method [set_size_request](https://docs.gtk.org/gtk3/method.Widget.set_size_request.html) is called.
This method increases the height of the dialog, so that the user can see the new line that he has created.

Unfortunately, setting the size request will force the user to leave the window at least as large as the size request (see [docs](https://docs.gtk.org/gtk3/method.Widget.set_size_request.html)).
This is a problem, if you already have too many feeds and adding another one increases the dialog to a height that is not supported by your monitor.

If that's the case, the user can't add feeds anymore.

Can we use [set_default_size](https://docs.gtk.org/gtk4/method.Window.set_default_size.html) instead?
This wouldn't remove the resize-functionality and just add another row with scroll-behavior.

---
And thanks for this applet! :)

@jake1164

### DIFF
--- a/feeds@jonbrettdev.wordpress.com/files/feeds@jonbrettdev.wordpress.com/manage_feeds.py
+++ b/feeds@jonbrettdev.wordpress.com/files/feeds@jonbrettdev.wordpress.com/manage_feeds.py
@@ -320,7 +320,7 @@ class MainWindow(Gtk.Window):
         """ Adds a new row to the bottom of the array / Grid """        
         self.config.feeds.append([ConfigFileManager.get_new_id(), True, "http://", "", True, 5, False, False])        
         self.treeview.set_cursor(len(self.config.feeds) - 1, self.treeview.get_column(0), True)
-        self.set_size_request(-1, 150 + len(self.config.feeds) * 20 )
+        self.gtk_window_set_default_size(-1, 150 + len(self.config.feeds) * 20 )
         
 
     def save_clicked(self, button):


philipp-meier:
I just noticed that the applet does not support having more feeds that exceed the available screen height.
If that's the case,  the FeedDisplayMenuItems can't be expanded anymore.

By adding a vertical scrollbar with the following commit to my fork, I was able to fix it for me:
[e3ed645f3f3ad401c2c5c398963304ef8b4b7051](https://github.com/philipp-meier/cinnamon-spices-applets/commit/e3ed645f3f3ad401c2c5c398963304ef8b4b7051)

If someone is interested, I can also add this to the PR or create a new PR for the additional scrollbar.
I am not sure whether the solution is sustainable though, since I do not have any experience with [GJS](https://gjs.guide/guides/) and just tried to get a working solution/workaround.

rcalixte:
cc @jake1164 if you have any thoughts or feedback on this

fredcw:
Since @jake1164 is not replying, maybe this applet should be considered abandoned. @philipp-meier maybe you could become it's maintainer?

philipp-meier:
@fredcw I haven't done this before, but I would definitely do it if it was abandoned.

rcalixte:
> @fredcw I haven't done this before, but I would definitely do it if it was abandoned.

In the metadata, update the author field to your GitHub username and set the current author field as "original_author" and push it here. Feel free to look at other applets if you need an example.

philipp-meier:
@rcalixte Alright. I updated the files. I added jonbrett as original author and ensured jake1164 is in the original-authors list. I hope that's correct.

PR #6297, opened 2024-08-09
Cinnamenu@json: Update .po files

fredcw:
@rcalixte Thanks for updating the po files. I usually remove the comments just so it saves a bit of space though it's not probably not important.

### DIFF
--- a/Cinnamenu@json/files/Cinnamenu@json/po/Cinnamenu@json.pot
+++ b/Cinnamenu@json/files/Cinnamenu@json/po/Cinnamenu@json.pot
@@ -17,899 +17,533 @@ msgstr ""
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
-#: 5.8/applet.js:59 3.2/applet.js:157 4.0/applet.js:60
 msgid "Initializing"
 msgstr ""
 
-#: 5.8/applet.js:75 4.0/applet.js:78
 msgid "Find files..."
 msgstr ""
 
-#: 5.8/applet.js:283 4.0/applet.js:281
 msgid "Please wait..."
 msgstr ""
 
-#: 5.8/applet.js:909 5.8/applet.js:1124 4.0/applet.js:884
 msgid "Applications"
 msgstr ""
 
-#: 5.8/applet.js:913 4.0/applet.js:888
 msgid "Documents"
 msgstr ""
 
-#: 5.8/applet.js:917 4.0/applet.js:892
 msgid "Videos"
 msgstr ""
 
-#: 5.8/applet.js:921 4.0/applet.js:896
 msgid "Images"
 msgstr ""
 
-#: 5.8/applet.js:925 4.0/applet.js:900
 msgid "Music"
 msgstr ""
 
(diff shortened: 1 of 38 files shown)


rcalixte:
> I usually remove the comments just so it saves a bit of space though it's probably not important.

The comments don't end up in the compiled translations (so the size issue is largely only related to cloning the repository and even then, it's on the order of bytes) and they serve as a context aid for translators: image

On top of that, this fights against the automation that will eventually be in place that will revert these removals.



What is the net gain of this?

fredcw:
>the size issue is largely only related to cloning the repository and even then, it's on the order of bytes

It adds to the applet download size and the disk space required for the applet, 3.2MB compared to 1.9MB.

>they serve as a context aid for translators

That's the intention, yes, but it's hard to say whether translators find them useful or even look at them at all. I guess we'd have to ask them. Comments that just say, for instance, applet.js are probably not that useful but in cinnamenu they're probably more useful since they come from different files: categories.js, contextMenu.js, etc.

I also remove the list of internet search providers and the list of wikipedia languages derived from settings-schema.json as these shouldn't be translated anyway.

I'm happy to keep the comments if translators find them useful but I do need to remove the list of wikipedia languages because I think the list needs to be the same as the one used by wikipedia itself, ie. the language name written in it's own language.

rcalixte:
> It adds to the applet download size and the disk space required for the applet, 3.2MB compared to 1.9MB.

On a 1 Mbps connection, this still amounts to seconds at most. Effectively, this will only be noticeable to users on 2G connections. (3G is around 7 Mbps under normal conditions.) Bandwidth/hosting costs are also negligible. In all, I'm not seeing the need to over-optimize this.


> That's the intention, yes, but it's hard to say whether translators find them useful or even look at them at all.

Remove them because they _might_ not be useful... but for what gain? What are the pros and cons here? (And many translators use Poedit (what I used for the screenshot) which displays this by default.)


> Comments that just say, for instance, applet.js are probably not that useful but in cinnamenu they're probably more useful since they come from different files: categories.js, contextMenu.js, etc.

The comments also include line numbers, adding context if needed. The entire goal is to aid the translators already volunteering their time.


> I'm happy to keep the comments if translators find them useful but I do need to remove the list of wikipedia languages because I think the list needs to be the same as the one used by wikipedia itself, ie. the language name written in it's own language.

I was conflicted on this but doesn't it ultimately make sense for the languages to be translated for the end user who wants to adjust their settings? When I go to a website with language options, they are often displayed as translated, sometimes not. Ultimately, maybe we need to find a way for specific entries in the settings to be excluded. Manually managing these files will ultimately be more laborious than it is worth. That's the whole point of striving to have it be automated.

fredcw:
@rcalixte I'm happy with this PR as it is. Thanks.

erwn16:
**Comments in a po-file will lead to better translations**

I translate Cinnamenu@json and Hideable applets into German. Translating the entries of the user interfaces is in the great majority of all cases determined by other entries of a particular user interface (the context), the chosen wording and other aspects. Comments in a po-file, which inform about a multiple use of a translation, facilitate extremely the translation process. I was in this described situation while translating Cinnamenu@json. This is for me the main reason to keep them in a po-file because they support better translations.

PR #6302, opened 2024-08-11
quick-screenshot-sharing@Odyssey: Initial release

Odyssey:
This applet allows capturing a screen region using gnome-screenshot and immediately uploading it to a file hosting service and getting the direct access link copied straight to your clipboard, so you can share it with others, while also saving a history of your uploaded pictures and their links in a log file in the user's home directory. It's _slightly_ more minimalistic than similar programs such as flameshot or sharex.

The [uploader service](https://kappa.lol/uploaders) is widely used in combination with some extensions for Twitch streamers and has now been active for about 10 months. While it _could_ go down in a distant future, it could be fun while it lasts and even funnier to figure out another accountless, free of api keys, unrestricted service down the line.


quick-screenshot-sharing@Odyssey/files/quick-screenshot-sharing@Odyssey/applet.js
```diff
@@ -0,0 +1,124 @@
...
+    return myApplet;
+}
+
+// Checks if all the required packages required for the applet to function properly are installed
+function passesPackageCheck() {
+    let packages = ["xdg-user-dirs", "gnome-screenshot", "curl", "xclip", "jq"];
+    for (let pkg of packages) {
+        let [success, pid, stdin, stdout, stderr] = GLib.spawn_async_with_pipes(
+            null,
+            ['sh', '-c', `dpkg -l | grep -w ${pkg}`],
```
rcalixte, line 84: This will not work for non-Debian-based users. (The other package mentions and commands also are only applicable to Debian-based distributions.)

Odyssey: would you rather have me use something like ```which ${pkg}``` , being present in a bunch of distros by default (deb, fedora, arch...) or get the user distro (eg reading /etc/os-release) and use the main way from their main package manager (pacman, dnf, etc)? 

It's my first applet and I feel like some obscure distro is going to fail the check regardless. do I just log a generic warning in those cases, like "if the applet doesn't function properly you might need to install x y z"?

rcalixte: It's ultimately your call. Take a look at other applets that have implemented similar things. Some have a README with details, some check for binaries, etc. etc.

Odyssey:
should be ready now

rcalixte:
@Odyssey I'm not sure what to do here... this functionality is already built into apps like Flameshot and Shutter but neither works without credentials like this service does. If it's okay with you, I'm thinking we can wait to see what additional community feedback looks like.

Either way, thank you for your contribution here and your translations!

Odyssey:
No problem, I'll leave it open and you do as you see fit. If you feel like closing it now or eventually, go ahead.

Personally, I just wanted something that did just the basic job quick and easy without throwing too many possibilities at my screen, and opted to share it because other existing screenshot applets only have local functionality.

Odyssey:
Unnecessary as an applet as it's fairly easy to achieve the same goal with some scripting involving existing apps as you mentioned. [Here](https://gist.github.com/Odyssey/6c06bb06ddf19d8d864aac357bc90ac2) is a brief example of uploading to an alternative host anonymously using flameshot while keeping full gui capabilities. Closing.

PR #6303, opened 2024-08-12
Remove Cinnamenu & CinnVIIStarkMenu applets

fredcw:
@clefebvre If I'm not allowed to make changes to my applets as I see fit, then I do not want my applets on Cinnamon. Please merge PR #6297 or remove my applets from Cinnamon. Thanks.


rcalixte:
@fredcw It would be a shame to lose you over what feels like a trivial disagreement. You've been a phenomenal contributor and generally great to work with.

That being said, as you are not the original author of these, your proper course of action if you would like to continue down this path is to remove yourself as the maintainer.

PR #6357, opened 2024-08-31
wine-utils@elblake: Initial release

elblake:
Hello,

I've made a small applet to make it easier to access the various utilities of Wine.

Thanks,
Edward Blake


wine-utils@elblake/files/wine-utils@elblake/applet.js
```diff
@@ -0,0 +1,450 @@
...
+                Lang.bind(this, this.use_cmd));
+        }
+
+        if (this.showAdminProgramsMenu) {
+            this.menu_tools(_('Configuration'), this.menu_to_binds(
+                this.admin_programs_list));
+        }
+
+        if (this.showProgramsMenu) {
+            this.menu_programs = this.menu_tools(_('Programs'), []);
```
rcalixte, line 216: Just a note that these functions will not function without the proper translation support (importing Gettext). For the applet code, you can look at other applets for an example. For the bash scripts, there are a decent volume of examples if you look at the [cinnamon-spices-actions](https://github.com/linuxmint/cinnamon-spices-actions) repository. After both of these are done, there is a script in the root of the repository that can be used to generate the translation template file that translators can use: `./cinnamon-spices-makepot wine-utils@elblake`

This looks like it will at least work for English-speaking users as-is. Would you rather see this merged now and implement the proper translation support after the fact or should this wait for the translation support before you want to have it merged?

elblake: Ah I didn't know about the Gettext part, I'll add that in right away. I'll make the needed changes for the shell scripts as well too.

elblake: I think I've made the needed changes.

rcalixte (approved):

wine-utils@elblake/files/wine-utils@elblake/applet.js
rcalixte: Well done!

elblake:
@rcalixte Thank you!

PR #6374, opened 2024-09-04
vocab@rizanb: Initial release

rizanB:
screenshot

This applet shows a new vocabulary on the panel every time the user clicks on it, or on reboots. Simple, but very useful to learn new words on the fly.

**Features**
- Displays a random word on the panel

Please let me know if you have any concerns or feedback.

Regards,
Rizan B.


claudiux:
A database with just 300 words and their definitions, in English with no possibility of translation, hard-coded in applet.js.

The `extension.js` file is useless.

This applet should be reworked.

PR #6376, opened 2024-09-05
Add eu.po

alexgabi:
Add basque language (eu)


cinnamon-timer@jake1164/files/cinnamon-timer@jake1164/po/eu.mo
rcalixte, line 1: This file should be removed.

rcalixte (changes requested):

rcalixte (approved):

Issue #5519, opened 2024-02-16
Applet Bug app-launcher@mchilli


vilqvilq:
### Applet version/Build date

 "last-edited": 1707164733

### Cinnamon version

5.2.7

### Distribution

Mint 20.3

### Graphics hardware *and* driver used

intel HD Graphics P530 // i915

### Applet name and maintainer

app-launcher@mchilli @mchilli

### What happened?

More operations makes applet slower and slower in time... [even 5min]

1. from CinnVIIStarkMenu

a] Moving apps and dropping to applet

2. in Applet Editor

a] Moving apps/groups
b] Editing apps/group

after 10-15 operations cinnamon needs restart and cycle repeats...

### Other information

Looking Glass shows no info

mchilli:
Hey and I'm sorry you're having trouble with my applet. Maybe it's because the applet is always updated when you change the order or edit an entry. Unfortunately I don't have the time at the moment, so it may take some time to find a solution.
I hope you can use it anyway.

PR #6387, opened 2024-09-08
app-launcher@mchilli: refactored code

mchilli:
- the widget will now be displayed more natively in the settings window
- this should also improve the handling when maximizing the window
- added a save and chancel button to reduce the effort of rebuilding the applet
- this should fix:#5519
- use a custom icon button ref:linuxmint/cinnamon#12270
- update the screenshot
- update the translation files


app-launcher@mchilli/files/app-launcher@mchilli/dialogs.py
rcalixte, line 1: @mchilli Have you tested the translation support in this file?

rcalixte (approved):

app-launcher@mchilli/files/app-launcher@mchilli/dialogs.py
mchilli: It worked for me for German. I didn't test every single translation, sorry. Hopefully it works.

rcalixte: > It worked for me for German.

This is actually what I meant so it's good to know.

> But the strings in the glade file are not used by the script. Maybe I should delete them.

You can set them as translatable and they should be detected. I thought I saw that the strings were detected too. If there are any that should be translated and were missed, we should try to fix it.

PR #6424, opened 2024-09-15
power-profiles@rcalixte: Update power profile icons

anaximeno:
cc @rcalixte 

Got them from the Papirus icon theme, they look better on a smaller scale (and bigger ones as well) than the previous ones.

imagem


anaximeno:
@rcalixte on a side note, may I also suggest that you send a notification when the power profile is not found installed on the system? Probably on the section below:

https://github.com/linuxmint/cinnamon-spices-applets/blob/c47e25cae5b4f652118ca6dcc8b7015b835873ac/power-profiles%40rcalixte/files/power-profiles%40rcalixte/5.4/applet.js#L58-L66

I only found out why it was working because I looked into the logs, and that's probably not something a regular user would do, I also noted it is on the website README afterward, but that's also not something some/most regular users would look for I believe.

rcalixte (approved):

rcalixte:
> Got them from the Papirus icon theme, they look better on a smaller scale (and bigger ones as well) than the previous ones.

The previous icons were a hack job by me but usable because there was room to adjust should there ever be more than three states for Power Profiles in the future.



> may I also suggest that you send a notification when the `power-profiles-daemon` is not found installed on the system?

I can look into this. It will have to check the path instead of the system packages due to the variance in distros. My only hesitation is that the README also includes the information about conflicts which is also extremely important if users are migrating from something else prior to using this. Maybe I would have to include a button on the notification to open the README too. I'll open an issue for now.

rcalixte:
The icons feel a little bit thick but I'll give them time.

PR #6434, opened 2024-09-18
ddcci-multi-monitor@tim-we: Retry setting brightness and raise timeouts

qwertz19281:
Add a retry mechanism to ddcci-multi-monitor@tim-we for setting the brightness and raise some timeouts.

For updating brightness, retry up to 16 times and increase timeout to 5000ms.

For querying, increase timeout to 500ms.

This should improve the DDC/CI brightness applet functionality with monitors which are slow/unreliable reading/updating DDC/CI brightness.

The way it is done (with the send/sent_brightness) also minimizes cascading of updates if the brightness is changed quickly in succession.

cc @GingerGigiCat @tim-we

### DIFF
--- a/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
+++ b/ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
@@ -38,6 +38,8 @@ class Monitor {
         this.index = index;
         this.name = name;
         this.brightness = 50;
+        this.send_brightness = null;
+        this.sent_brightness = null;
         this.bus = bus;
         this.menuLabel = null;
         this.menuSlider = null;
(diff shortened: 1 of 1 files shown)


ddcci-multi-monitor@tim-we/files/ddcci-multi-monitor@tim-we/applet.js
```diff
@@ -82,18 +84,44 @@ class Monitor {
...
         this.updateMenu();
         this.#promises = this.#promises.then(() => {
             return new Promise((resolve, reject) => {
-                Util.spawnCommandLineAsync(
-                    `ddcutil --bus=${this.bus} setvcp 10 ${this.brightness}`,
-                    resolve,
-                    reject
-                );
+                setTimeout(resolve, 5000);
+                log(`trySetBrightness start`);
```
rcalixte, line 92: Do you need these `log` calls? They seem like debug values.

qwertz19281: Shall i remove them?

GingerGigiCat:
The changes make sense and would probably fix some strange edge cases, so thanks for adding. 
And of course, the extension still works normally with your changes so it can be merged.

rcalixte (approved):

PR #6438, opened 2024-09-24
[download-and-upload-speed@cardsurf] v1.6.6 - Binary or decimal units can now be selected

claudiux:
@cardsurf 

Binary or decimal units can now be selected in settings.

image

image

image

image

'Decimal' is the default.

Also, units are now translatable. French translation is provided.

I hope you'll like these changes.

Claudiux

N.B. I can do the same for disk-read-and-write-speed@cardsurf if you wish.


rcalixte:
@claudiux Maybe this could have waited a little while longer to get some feedback from the author?

claudiux:
@rcalixte The author has not posted anything to github since 2019.

rcalixte:
> @rcalixte The author has not posted anything to github since 2019.

You might mean 2021 here.

Do you think this needs to have the maintainer status updated then? Has there been lingering issues that have gone unresolved? Sometimes an applet is simple enough that it does not need regular maintenance. And activity as GitHub's profile pages is not always an accurate indicator since it does not reflect involvement in Issues, Discussions, or other Pull Requests.

PR #6474, opened 2024-10-08
Create is

sveinki:
Adding initial translation for Icelandic


gpaste-reloaded@feuerfuchs.eu/files/gpaste-reloaded@feuerfuchs.eu/po/is
rcalixte, line 1: Can you rename this to is.po?

rcalixte (changes requested):

PR #6493, opened 2024-10-13
[pin-unpin-panel@anaximeno] User can set panel pinned/unpinned at startup

claudiux:
@anaximeno 

Hi,
I often forget to unpin a panel before closing my session and this annoys me.

So I'm proposing here that the user should be able to set the panel state when the applet starts up. Three states are available:

- Leave the panel as it is.
- Pin the panel.
- Unpin the panel.

I hope you'll like this.

Regards
claudiux


anaximeno:
Hello @claudiux, looks good to me.

Thanks for your contribution!

rcalixte (approved):

PR #6511, opened 2024-10-19
Increase refreshInterval max value to 3600 seconds

dylanparker:
Every 60s is too frequent for many uses of bash-sensors. Increasing the max value allows for more options.


rcalixte (approved):

dylanparker:
Thank you!

PR #6525, opened 2024-10-22
Better to keep the original name untranslated.

Wibol:


rcalixte (approved):

rcalixte:
cc @haggen88 so everyone is on the same page

cc @Gr3q should these names be consistent in the applet code itself?

Gr3q:
> cc @Gr3q should these names be consistent in the applet code itself?

Yes they should, I'll add it to the next release.

PR #6527, opened 2024-10-24
profile-switcher@pixunil add german translation

Joshiy13:
Add german translation to profile-switcher@pixunil


Joshiy13:
Checks failing, will fix

Joshiy13:
Checks are passing now

rcalixte (approved):

PR #6573, opened 2024-11-14
sound150@claudiux: Play volume change sound after changing volume via dragging slider

timfeierabend:
fixes https://github.com/linuxmint/cinnamon/issues/9472

Users migrating from other operating systems might expect that changing the volume slider would play an audible tone to indicate how loud the volume is. Right now, the only ways to get this tone to play from the applet are to use the keyboard or mouse scroll wheel. This adds another option for users who prefer to use mouse/touch screen or don't have a mouse wheel to hear how loud the volume is.

@claudiux


claudiux:
@timfeierabend 
Hi,

Thank you for this contribution!

The sound is only emitted at the beginning and end of the drag, but not during the drag. Is this really what you want?

timfeierabend:
Yes - the sound should be emitted only at the end of the drag. I based the behavior off the volume slider in Windows.

claudiux (approved):

PR #6606, opened 2024-11-22
[MuteToggler@jebeaudet.com] Added new parameters for soundcard and input choice

lrademacher:
I am using a USB microphone (which is a soundcard by it's own).
The command 'amixer set Capture toggle' does not work in this environment.
In my case i need to select the soundcard as well and the control is not called 'Capture' but 'Mic'.
This extension let's you configure two strings that let you choose the device by a search string as it is displayed in 'cat /proc/asound/cards' and name control (Capture, Mic, ...).

Works fine for me, could also work fine for others...

### DIFF
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -152,6 +152,7 @@ MyApplet.prototype = {
         }
 
         VERBOSE = this.verbose;
+        this.evaluate_soundcard();
     },
 
     refresh_loop: function() {
--- a/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
+++ b/MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
@@ -16,7 +16,9 @@ const APPNAME = "MuteToggler";
 const UUID = APPNAME + "@jebeaudet.com";
 const HOME_DIR = GLib.get_home_dir();
 
-var VERBOSE = true; // VERBOSE value will be changed according to this applet settings.
+var VERBOSE = false; // VERBOSE value will be changed according to this applet settings.
+
+const POTENTIAL_INPUTS = ["'Capture'", "'Mic'"];
 
 /**
  * Usage of log and logError:
@@ -41,6 +43,15 @@ function _(str) {
   return gettext(str);
 }
 
+function findFirstMatch(searchStrings, checkString) {
+  for (let i = 0; i < searchStrings.length; i++) { 
+    if (checkString.includes(searchStrings[i])) { 
+      return searchStrings[i]; 
+    }  
+  } 
+  return null; // Return null if no match is found 
+}
+
 MyApplet.prototype = {
     __proto__: Applet.IconApplet.prototype,
 
@@ -58,11 +69,6 @@ MyApplet.prototype = {
                 this.settings.bindProperty(Settings.BindingDirection.IN,
                                      "soundcard",
                                      "soundcard",
-                                     this.on_settings_changed,
-                                     null);
-                this.settings.bindProperty(Settings.BindingDirection.IN,
-                                     "input",
-                                     "input",
                                      this.on_settings_changed,
                                      null);
                 this.settings.bindProperty(Settings.BindingDirection.IN,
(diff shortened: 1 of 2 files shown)


rcalixte:
cc @jebeaudet

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
```diff
@@ -192,12 +213,51 @@ MyApplet.prototype = {
         }
     },
 
+    evaluate_soundcard: function() {
+      // only use specific soundcard if searchstring is not empty
+      if (this.soundcard.trim() !== '')
```
claudiux, line 217: Use `if (this.soundcard.trim().length > 0)` instead of `if this.soundcard.trim() !== '')`

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
```diff
@@ -192,12 +213,51 @@ MyApplet.prototype = {
...
+      {
+        // per default use first soundcard 
+        this.soundcard_id = "0";
+  
+        let soundcard_list_cmd = "cat /proc/asound/cards"; 
+        let [res, stdout] = GLib.spawn_command_line_sync(soundcard_list_cmd);
```
claudiux, line 223: avoid using synchronous functions; prefer asynchronous functions.

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
```diff
@@ -192,12 +213,51 @@ MyApplet.prototype = {
...
+        }
+        log("Use specific soundcard id: '" + this.soundcard_id + "'");
+
+        this.amixer = this.amixer + " -c " + this.soundcard_id;
+      }
+    },
+
+    evaluate_input: function() {
+      this.input = POTENTIAL_INPUTS[0];
+      let [res, stdout] = GLib.spawn_command_line_sync(this.amixer);
```
claudiux, line 243: avoid using synchronous functions; prefer asynchronous functions.

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
```diff
@@ -192,12 +213,51 @@ MyApplet.prototype = {
...
+      {
+        // per default use first soundcard 
+        this.soundcard_id = "0";
+  
+        let soundcard_list_cmd = "cat /proc/asound/cards"; 
+        let [res, stdout] = GLib.spawn_command_line_sync(soundcard_list_cmd); 
+
+        if (res) {
+          // Split the result into lines 
+          let lines = stdout.toString().split('\n');
```
claudiux, line 227: Avoid `stdout.toString()`.

Use `to_string(stdout)` instead with:
```
const to_string = function(data) {
  const extraByteMap = [ 1, 1, 1, 1, 2, 2, 3, 0 ];
  var count = data.length;
  var str = "";

  for (var index = 0;index < count;)
  {
    var ch = data[index++];
    if (ch & 0x80)
    {
      var extra = extraByteMap[(ch >> 3) & 0x07];
      if (!(ch & 0x40) || !extra || ((index + extra) > count))
        return null;

      ch = ch & (0x3F >> extra);
      for (;extra > 0;extra -= 1)
      {
        var chx = data[index++];
        if ((chx & 0xC0) != 0x80)
          return null;

        ch = (ch << 6) | (chx & 0x3F);
      }
    }

    str += String.fromCharCode(ch);
  }

  return str;
}
```

claudiux (changes requested):
Please review some parts of the code (see comments).

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
lrademacher: If using asynchronous functions: How can we be sure that evaluate_soundcard was already executed when executing evaluate_input?

claudiux: Welcome in the world of async, await, promise etc.

https://www.w3schools.com/js/js_async.asp

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function

lrademacher: now i did this via spawn_async. Any benefits of spawnCommandLineAsyncIO?

claudiux: Advantages, not really, but I think it's easier to use.

MuteToggler@jebeaudet.com/files/MuteToggler@jebeaudet.com/applet.js
claudiux: You forgot to declare to_string, didn't you?

lrademacher: Well using Util.spawn_async gives stdout as string already.
And in line 13 there is already a to_string function imported, that is used in another place of the applet.

claudiux (approved):

claudiux:
Tested. It works well adding `this.evaluate_soundcard();` at the end of on_settings_changed().

lrademacher:
> Tested. It works well adding `this.evaluate_soundcard();` at the end of on_settings_changed().

Great. Do I understand correctly, that this fixes your previous comment?

I think, we should maybe put the whole cmd-line evaluation process into a separate function and call it in on_settings_changed. 
Something like this:

```
evaluate_cmd_line: async function() {
  await this.evaluate_soundcard(); 
  await this.evaluate_input();

  const parameters = ["", " -D pulse"];
  for (let param of parameters) {
      let cmd = this.amixer + param + " scontrols";
      log("Test mixer command '" + cmd + "'");
      // TODO: Make it async.
      let [res, stdout] = GLib.spawn_command_line_sync(cmd);
      if (res && to_string(stdout).indexOf(this.input) != -1) {
          this.amixer += param;
          log("Use mixer command '" + this.amixer + "'");
          break;
      }
  }
}
```
Should I test and add a new pull request?

claudiux:
@lrademacher It was an oversight on my part. Please create a new PR.

PR #6607, opened 2024-11-22
Calender@ccprog version 1.3.0

### LINKED ISSUE #6587: [calendar@ccprog] - Add control over when to call Enrico service

calendar@ccprog @ccprog

Hello,

I appreciate the addition of non-working days to the calendar, however I noticed that currently it pings the Enrico service (e.g. https://kayaposoft.com/enrico/json/v2.0?action=getHolidaysForYear&year=2024&country=fra&holidayType=public_holiday) everytime the applet starts/restarts.
This is not so nice privacy-wise, and I believe not so useful since the data is not likely to be updated so often. It would be great to be able to set when Enrico should be called. Options I can think about:
- new geographic zone set (load data for new country code)
- new date set (if year is modified - load data for new year)
- monthly automatic update (or any other more appropriate period)
- now (manual trigger)

ccprog:
- Thanks for multiple translations: blogdron for Russian, qadzek for Dutch, Odyssey for Catalan
- Multiple fixes merged from upstream calendar@cinnamon
- Included custom tooltip formatting from upstream calendar@cinnamon
- Bugfix: correctly check cache whenever data are needed
- Support for holidays in Cyprus, Montenegro and El Salvador, and for regional holidays in Belgium.  
  **Users from Belgium will need to check their regional settings!**

Resolves #6587

### DIFF
--- a/calendar@ccprog/CHANGELOG.md
+++ b/calendar@ccprog/CHANGELOG.md
@@ -1,4 +1,12 @@
-###1.2.2
+### 1.3.0
+- Thanks for multiple translations: blogdron for Russian, qadzek for Dutch, Odyssey for Catalan
+- Multiple fixes merged from upstream calendar@cinnamon
+- Included custom tooltip formatting from upstream calendar@cinnamon
+- Bugfix: correctly check cache whenever data are needed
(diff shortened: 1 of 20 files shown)


rcalixte (approved):

ccprog:
That's what I do, why do you think I did something differently?

PR #6616, opened 2024-11-24
[test-spice] Add applet reload after successful copy

timfeierabend:
**Change:** 
Add automatic reload of applet after copying to DEVTEST path

**Rationale:** 
Currently, when a developer needs to test their applet, they need to run `./test-spice uuid`, followed up by explicitly entering the applet settings, and triggering a reload of the applet from the hamburger menu. In my tests, removing/re-adding the applet from the panel didn't trigger a reload in Cinnamon.

Now, when a developer runs `./test-spice uuid`, the script will tell Cinnamon to reload the applet using same method as the applet settings, making it easier for the developer to test their changes.

@rcalixte

### DIFF
--- a/test-spice
+++ b/test-spice
@@ -53,15 +53,7 @@ def reload_xlet(uuid):
     """
     Reloads the Spice via dbus-send.
     """
-    args = ['dbus-send', 
-        '--dest=org.Cinnamon',
-        '--print-reply',
-        '--type=method_call',
-        '/org/Cinnamon',
-        'org.Cinnamon.ReloadXlet',
-        f'string:{uuid}', 
-        'string:APPLET'
-        ]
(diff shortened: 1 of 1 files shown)


test-spice
```diff
@@ -42,10 +44,29 @@ def copy_xlet(uuid):
...
+    """
+    args = ['dbus-send', 
+        '--dest=org.Cinnamon',
+        '--print-reply',
+        '--type=method_call',
+        '/org/Cinnamon',
+        'org.Cinnamon.ReloadXlet',
+        f'string:{uuid}', 
+        'string:APPLET'
+        ]
```
rcalixte, line 64:
```suggestion
    args = ['/usr/bin/cinnamon-dbus-command', 'ReloadXlet', uuid, 'APPLET']
```

rcalixte (approved):

PR #6651, opened 2024-11-29
[MuteToggler@jebeaudet.com] Separate cmd line evaluation and executed on settings change

lrademacher:
This is a follow-up of #6606
The current solution is not clean. on_settings_change() only evaluates the soundcard id, but not the rest of the amixer cli.
This fixes it and also ensures that on_settings_change() is not executed twice on _init().


rcalixte:
cc @jebeaudet

rcalixte (approved):

PR #6670, opened 2024-12-06
cinnamonvantage@garlayntoji: Fixed notification

Garlayntoji:
I finally fixed the notification system in cinnamonvantage: when switching to hybrid graphics, a notification was supposed to show up and say that a system restart is required. It is now fixed.


Garlayntoji:
I have multiple commits attached to this PR because of the multiple syncs I had to perform since my last contribution. Sorry about that.

rcalixte (approved):

Issue #6613, opened 2024-11-24
power-profiles@rcalixte: Add support for TuneD


rcalixte:
### Applet name and maintainer

power-profiles@rcalixte @rcalixte

### What would you like to see?

Implement proper support for [TuneD](https://github.com/redhat-performance/tuned) via DBus.

cc @hansingt

hansingt:
I'm not sure, whether I get this issue right, but as I've already mentioned, TuneD provides a compatibility API for applications using PPD. 

Thus, in first place, it would be enough to make it work,even though PPD is not installed and maybe check whether the required APIs are available instead, if that is possible.

In future releases, direct support for TuneD might be a good idea though.

PR #6698, opened 2024-12-20
power-profiles@rcalixte: Support TuneD PP API by only checking for powerprofilesctl if DBus is not found

hansingt:
Red Hat based distributions like Fedora started using [TuneD](https://github.com/redhat-performance/tuned) instead of power-profiles-daemon. See #6613.

The TuneD daemon supports a  power-profiles compatibility API, but the current impelementation of the power-profiles applet explicitly checks whether the `powerprofilesctl` command is installed.

This PR changes the logic to first check, whether the corresponding DBus is available and only, if it is not available checks, whether the powerprofilesctl command is available or not.

This allows for other daemon providing a compatible DBus system to work with the power-profiles applet.

### DIFF
--- a/power-profiles@rcalixte/files/power-profiles@rcalixte/5.4/applet.js
+++ b/power-profiles@rcalixte/files/power-profiles@rcalixte/5.4/applet.js
@@ -54,44 +54,18 @@ PowerProfilesApplet.prototype = {
     this._profilesProxy = null;
     this._proxyId = 0;
 
-    if (!GLib.find_program_in_path("powerprofilesctl")) {
-      let source = new MessageTray.Source(this.metadata.name);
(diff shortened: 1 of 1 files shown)


rcalixte (approved):

rcalixte:
This will work in the short-term but just to be clear, the reason that I'd opened the issue and assigned it to myself was because I was working on it. There are still some changes that need to be implemented but they are less urgent. We also need to test further to make sure that all situations are handled properly.

hansingt:
> This will work in the short-term but just to be clear, the reason that I'd opened the issue and assigned it to myself was because I was working on it. There are still some changes that need to be implemented but they are less urgent. We also need to test further to make sure that all situations are handled properly.

Yes. I totally agree. I don't know for myself, how long TuneD will maintain the compatibility API. Tje intention of my change was just, to make it usable again wie before the update. I agree, that propper support sjould be implemented and needs testing.

PR #6712, opened 2024-12-27
[gpaste-reloaded@feuerfuchs.eu] Add option to hide tray icon

benmpl:
Add option to hide tray icon while applet is not being used.
It would be shown when panel edit mode is on, or with key bindings.
Also added the respective .pot entry and included portuguese and spanish translations.


rcalixte:
cc @feuerfuchs

rcalixte (approved):

rcalixte:
@benmpl Just a note, there is a script in the base of the repository named `cinnamon-spices-makepot` that is designed so that the template is updated automatically and new strings are applied to existing translations that can then be modified by translators. Please try to avoid manually editing the template file to prevent inconsistent changes in the future.

benmpl:
@rcalixte My bad for missing that, I'll be sure to pay more attention in the future. Thanks for the heads up!

PR #6718, opened 2024-12-29
bing-wallpaper@starcross.dev: Add force refresh menu.

nabetaro:
I would like to make this wonderful applet even better.

It seems that once you change the background manually, you can't immediately change it to a bing background. So I would like to add the ability to refresh it immediately.

Please let me know if there is a better way to do this in implementing the feature.

### DIFF
--- a/bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
+++ b/bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
@@ -5,6 +5,8 @@ const Gio = imports.gi.Gio;
 const GLib = imports.gi.GLib;
 const Mainloop = imports.mainloop;
 const Lang = imports.lang;
+const PopupMenu = imports.ui.popupMenu;
+const St = imports.gi.St;
 
 const logging = false;
 
@@ -45,9 +47,18 @@ BingWallpaperApplet.prototype = {
         this.wallpaperPath = `${this.wallpaperDir}/BingWallpaper.jpg`;
         this.metaDataPath = `${this.wallpaperDir}/meta.json`;
 
+        let refreshBg = new PopupMenu.PopupIconMenuItem(_("Refresh Now"), "view-refresh", St.IconType.SYMBOLIC);
+        refreshBg.connect('activate', Lang.bind(this, function() {
+            this.forceReload = true;
+            this._refresh()
+        }));
+        this._applet_context_menu.addMenuItem(refreshBg);
+
         // Begin refresh loop
         this._refresh();
     },
+
+    forceReload: false,
 
     _refresh: function () {
         log(`Beginning refresh`);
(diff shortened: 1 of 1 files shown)


rcalixte:
cc @Starcross

bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
```diff
@@ -45,10 +47,19 @@ BingWallpaperApplet.prototype = {
         this.wallpaperPath = `${this.wallpaperDir}/BingWallpaper.jpg`;
         this.metaDataPath = `${this.wallpaperDir}/meta.json`;
 
+        let refreshBg = new PopupMenu.PopupIconMenuItem(_("Refresh Now"), "view-refresh", St.IconType.SYMBOLIC);
+        refreshBg.connect('activate', Lang.bind(this, function() {
+            this.forceReload = true;
```
Starcross, line 52: Instead of using a `forceReload` flag, I think it may be enough to simply make a call to `_downloadMetaData` here. 
The only difference this makes to the behaviour I can think of is that the timer will not be reset, but I don't think this matters. The idea of a refresh 'now' could be understood as an additional one anyway.

bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
```diff
@@ -140,7 +151,7 @@ BingWallpaperApplet.prototype = {
 
     _downloadMetaData: function () {
         const process_result = data => {
-            
+
```
Starcross, line 154: Do we need these extra line spaces? I appreciate the formatting is not consistent anyway!

Starcross (commented):
Thanks for your PR, I like this idea! I've left a few comments...

bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
rcalixte: This actually just looks like whitespace was trimmed, possibly by some auto-formatting. GitHub has a feature where you can view diffs with whitespace changes hidden as well. This doesn't seem like a bad thing since it is minor here.

bing-wallpaper@starcross.dev/files/bing-wallpaper@starcross.dev/applet.js
nabetaro: @Starcross 
Thank you for your suggestion.

I stopped using flags.
how do you feel about it?

Starcross: Looks good! Working for me

Starcross:
LGTM @rcalixte 

Thanks

rcalixte (approved):

nabetaro:
@Starcross @rcalixte 

Thanks!!

PR #6719, opened 2024-12-30
added click-clock a clock that you need to click to show the time

wernerstucky:
A clock that helps with focus.
You need to click the icon to show the time, then it shows for 20 seconds and goes back to an icon.


rcalixte (approved):

rcalixte:
@wernerstucky You might want to take a look at other applets to see how to implement translation support. In addition, I would expect a common request to be a setting to configure the duration. You might want to take a look at that as well.

PR #6732, opened 2025-01-03
split used+cache up into used+shared+buffer+cache

binarynoise:
@OrcusCZ didn't implement my feature request, so I finally did it myself.

As 4 shades of blue wouldn't be distinguishable, I added violet and yellowish as new  color defaults (kinda inspired by htop, which I took my personal palette from).


rcalixte:
@binarynoise I wonder if everyone that uses this applet wants the same change. Is there any chance you can implement it as an option that defaults to the current way things are with an option to switch to the new format that you propose here?

binarynoise:
Easiest way to keep the old behaviour would be to use the same colors for used, shared and buffer. I could change the default colors accordingly.
I'm not experienced enough to add that kind of switch, I was happy I could just go ahead and duplicate the existing settings...
On the other hand, at least there is one person that did :+1: on the original issue, so it's too odd.

PR #6742, opened 2025-01-05
Delete batterypower@joka42 directory

ojtovsky:


rcalixte:
@ojtovsky You are not the listed author for this and have provided no useful information about why this should be removed. Can you elaborate?

PR #6760, opened 2025-01-13
cheaty@centurix: fixed typo in regex sheet.json

TheKessler0:
fixed an unescaped "\b" in the anchors sections


rcalixte:
cc @Centurix This looks like a bug fix so we'll push it through. Shout if there's anything else we should do.

rcalixte (approved):

PR #6765, opened 2025-01-15
[gpaste-reloaded@feuerfuchs] Fix label on ModalDialog

benmpl:
As per title:

Before:
before

After:
after


rcalixte:
cc @Feuerfuchs

rcalixte (approved):

PR #6782, opened 2025-01-20
[radio@driglu4it]: fix icons very large on new cinnamon

jonath92:
Without the fix, it looks like that on Linux Mint 22.1

image


rcalixte:
> Without the fix, it looks like that on Linux Mint 22.1

Have you tested this on older versions of Cinnamon?

jonath92:
No I haven't but I am optimistic that it works because in the past the "icon-size:16px" has been included in the mint-y cinnamon.css (see: https://github.com/linuxmint/mint-themes/commit/78965ce070d8b7e02feaa0e1c546e84a8a0228b8#diff-91f9c5ba3758104f896731e0373b935687baa6a5ba0ff731513cc68cd28b42eaL307) and for mint-x it seems to be still included. There is a small change it get broken 3rd party themes but that than needs to be fixed by the theme author (e.g. by using "important"). However I tested with Adapta-Nokto and there it still works.

rcalixte:
Hmm. It would be nice to know that this doesn't break anyone with Mint who hasn't upgraded or anyone not running Mint but still using Cinnamon. Would it make sense to move this to a versioned subdirectory?

jonath92:
yes, makes sense. I have added a new version

PR #6793, opened 2025-01-22
Add configurable keybinding for desaturate-all

K9-guardian:
I have this set in my local version of desaturate-all (in `~/.local/share/cinnamon/applets`). For some reason, when trying to use the `./test-spice` script, the settings menu would not open on the developer copy, and I'm not exactly sure why. However, this works for me locally so it might be an issue with me having both the developer copy and the main copy?


rcalixte:
cc @hkoosha

PR #6798, opened 2025-01-23
NightLightSwitch@claudiux - v1.0.0

claudiux:
Related to https://github.com/orgs/linuxmint/discussions/758#discussioncomment-11930364


DerDreschner:
This applet isn't listed in the bug report template, field `Applet name and maintainer`. Had to choose something else instead @claudiux . See #6800

Issue #6810, opened 2025-01-26
Applet Feature Request


rdlmda:
### Applet name and maintainer

mailnagapplet@ozderya.net @hyOzd

### What would you like to see?

I wrote a pull request to expand #5230 

- The limit is now configurable on the applet settings page.
- Limit now applies to sender name, sender address and subject.

dudumaroja:
can i abuse this sugestion thread? 

- it whould be nice to pin some stuff in the main applet menu like  choose your prefered output or input devices if you have many (so you dont have to open another submenu to change things,) shortcuts for custom with custom commands etc.

rdlmda:
@dudumaroja You can. As in "being able to". But you shouldn't. And it's quite pointless.

- It's a closed issue, marked as complete.
- No one is gonna see it besides you and me. Neither of us are repo mantainers.
- Even if anyone else saw it, what's your request got to do with the `mailnagapplet@ozderya.net` applet (which this feature request / commit is all about)?

Issue #5230, opened 2023-12-14
mailnagapplet@ozderya.net: Feature Request


rdlmda:
### Applet name and maintainer

mailnagapplet@ozderya.net @hyOzd

### What would you like to see?

Currently, there's no limit for the length of the message's sender and subject when displayed on the menu (see below):

![image](https://github.com/linuxmint/cinnamon-spices-applets/assets/4602525/f2a4ff1f-987b-43d5-a8c4-fb1e748f707a)

I suggest a configurable option to limit their length.

rdlmda:
This is just a suggestion. Values are hardcoded and the coding style is probably all wrong.

rdlmda:
Reopening for improvements:

- The limit is now configurable on the applet settings page.
- Limit now applies to sender name, sender address and subject.

rdlmda:
@rcalixte

PR #6811, opened 2025-01-26
Configurable char limit for mail sender and subject

rdlmda:
- The limit is now configurable on the applet settings page.
- Limit now applies to sender name, sender address and subject.

See #6810 and #5230 

Comments and suggestions are greatly appreciated.


rcalixte:
cc @hyOzd

rcalixte (approved):

PR #6924, opened 2025-02-25
mic-level@zamszowy: switch from pulseaudio to pipewire

zamszowy:
Support pipewire server instead of pulseaudio.


zamszowy:
> @zamszowy Just a note that you could use versioning for this applet so that folks on older versions of Cinnamon who might still be using PulseAudio would be unaffected.

I think that to be truly compatible, I would need to add some kind of detection mechanism, since for people like me, using Cinnamon with other distribution than Mint, pulseaudio could be in use with any Cinnamon version.

But I think that I'm the only one using this applet so far. 😄 If anyone will raise an issue with it not working with pulseaudio, I will consider such an improvement.

PR #6935, opened 2025-02-27
apt-updates@zamszowy: initial release

zamszowy:
Add applet that shows tray icons for pending updates in APT-based systems.


apt-updates@zamszowy/files/apt-updates@zamszowy/updates.sh
```diff
@@ -0,0 +1,23 @@
...
+check)
+    apt list --upgradable 2>/dev/null | tail -n +2 &> "$DIR"/updates
+    wc -l <"$DIR"/updates
+    ;;
+view)
+    /usr/bin/cjs "$DIR"/info-window.js "$(wc -l <"$DIR"/updates) updates" "$DIR"/updates
+    ;;
+command)
+    readonly cmd=$2
+    gnome-terminal --wait --  /bin/bash -c "echo \"Executing $cmd\"; $cmd; echo -en \"\nDone - press enter to exit\"; read"
```
rcalixte, line 18:
```suggestion
    gnome-terminal --wait --  /usr/bin/bash -c "echo \"Executing $cmd\"; $cmd; echo -en \"\nDone - press enter to exit\"; read"
```

rcalixte (approved):
Technically, `apt` isn't really meant to be used in scripts like this. There are services like PackageKit that can handle this and independently of a specific package manager as well.

zamszowy:
> Technically, `apt` isn't really meant to be used in scripts like this. There are services like PackageKit that can handle this and independently of a specific package manager as well.

Thanks, didn't know about PackageKit. I'll rework the applet to use `pkcon` instead of `apt`.

PR #6964, opened 2025-03-09
Cinnamenu@json: Update to 5.5.1

fredcw:


Cinnamenu@json/files/Cinnamenu@json/po/ar.po
```diff
@@ -15,900 +15,538 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
 "X-Generator: Poedit 3.0.1\n"
 
-#: 5.8/applet.js:59 3.2/applet.js:157 4.0/applet.js:60
```
rcalixte, line 18: @fredcw I really don't want to have to rehash this. Can you just run the scripts and leave the files unmodified? In the future this will all be automated away anyways. I still don't understand why this is a point of contention.

Outside of these, a number non-executable files were made executable. Can you fix the permissions on those as well?

rcalixte (changes requested):

PR #6987, opened 2025-03-21
turn-off-monitor@zablotski: Add Russian translation

Krivetochka:


rcalixte:
@Krivetochka Is this still in progress or are you comfortable with it as is?

Krivetochka:
I tested it and it seemed good to me.

Krivetochka:
I didn't translate some things because they were already translated by default, so I can call this translation finished.

rcalixte (approved):

PR #7050, opened 2025-04-07
turn-off-monitor@zablotski: Add optional feature of locking cinnamon session

D0han:
This PR adds feature of locking cinnamon session along with switching screens off. Configurable with default toggle state off, so nothing suddenly changes for anyone after update.
@zablotski


zablotski (approved):
Thank you for extending this @D0han!

rcalixte (approved):

PR #7058, opened 2025-04-10
Update it.po - Add Italian translations

smalldeenee:
Translated the remaining empty strings.

@Driglu4it
@jonath92 
First time committing something so I am following the guidelines by mentioning the author/mantainer.


rcalixte:
> First time committing something so I am following the guidelines by mentioning the author/mantainer.

Just a note that it is recommended to use a tool like Poedit instead of manually modifying the translation files, especially if you are new to the process.

jonath92:
Why do you think that I am the maintainer? I don't care about translations. Software translations are in my opinion a very bad thing. They add complexity to the code and make it more difficult for users to interact with developers. In my opinion, translation harms the software. Users should use the English version in 99% of cases. The only exceptions are cases where a translation is essential for the application

rcalixte (approved):

Issue #7080, opened 2025-04-18
cpu-monitor-text@gnemonix : Improve percentage display so applet width does not change


DoneWorkin:
### Applet name and maintainer

cpu-monitor-text@gnemonix @gnemonix

### What would you like to see?

When the CPU percentage changes number of digits the width of the applet changes and moves things around.  
Improve the functionality by padding the percentage with Figure Space characters.
[](https://en.wikipedia.org/wiki/Figure_space)

Suggested code change to _pad function as follows (will attempt to create pull request for this)

````
    _pad: function(percent) {
        let str = percent.toString();
        let str_length = this.max_percentage.toString().length;
        
        while(str.length < str_length) {
            str = "\u2007" + str;
        }
        return (str);
    },
````

PR #7129, opened 2025-04-27
cpu-monitor-text@gnemonix: Adjust minimum width of applet

fisherman734:
Since Issue #7080 has been fixed the width of the applet is too small to show the label (with the added whitespace). The percentage character gets (partially) cut off at the right border.

This adjusts the applet width by taking into account the length of the "Text" and "Max Percentage" fields. The  (hardcoded) single whitespace between the "Text" and "Max Percentage" fields has been removed, if needed it can be added to the end of the "Text" field.

@gnemonix


fredcw:
I can't reproduce the issue. What cinnamon theme are you using?

fisherman734:
Mint-L-Dark-Teal

PR #7160, opened 2025-05-01
Add applet: AlwaysOn@krivetochka V1.0.0

Krivetochka:
Clicking on the applet changes the system settings responsible for actions when the system is idle. When you click on the applet again, the settings change back.


claudiux:
@Krivetochka Please rewrite `applet.js` using `Gio.Settings` instead of all these `spawnxxx` functions.

Krivetochka:
> @Krivetochka Please rewrite `applet.js` using `Gio.Settings` instead of all these `spawnxxx` functions.

I'll do it, thanks for the advice.

PR #7161, opened 2025-05-02
gpaste-reloaded@feuerfuchs: Add scrollbar feature

GabrielCamara3526:
- Add a vertical scrollbar to @Feuerfuchs's clipboard applet so the user can scroll using the mouse wheel
- Increase Gpaste's max history size so each item is visible in the applet


fredcw:
This doesn't work.

GabrielCamara3526:
Hello, @fredcw ! I really appreciate your feedback. I built this solution as it really saved me, and later decided to push the lever and add this request. Unfortunately, i was having a few different issues on my desktop that demanded attention, so i thought of waiting patiently while having in mind the stipulated time break for reviews and a proper response. Today i closed this pull request since i believe @Feuerfuchs (the original spice author) is currently unavailable on Github, which could require further contact to work on. Sorry for taking your time.

fredcw:
@GabrielCamara3526 From the readme:

>If the author fails to respond in a reasonable time, we will assume the applet is abandoned (as mentioned above) and the pull request will be merged assuming it meets all other requirements.

@Feuerfuchs hasn't been active on github for a long time but in this, this PR was probably just overlooked.

PR #7217, opened 2025-05-12
auto-dark-light@gihaume: Refactored for JSDoc type checking via TS compiler

guillaume-mueller:
Why was every file set as executable ?


rcalixte:
> Why was every file set as executable ?

Any chance it was done by your editor or a local script? Do you want to push the fix or do you want me to push it?

guillaume-mueller:
OK so it happens because I test the changes on the actual installed files in `~/.local/share/cinnamon/applets`, because I can quickly relaunch Cinnamon with a keyboard shortcut, and then copy the files on the clone folder of this repository.

So I have those questions:
1. Why are the files of my applet in `~/.local/share/cinnamon/applets` unnecessarily set as executable while they weren't in this repository ?
2. Is my workflow wrong and should try another one ?

rcalixte (approved):

rcalixte:
> * Why are the files of my applet in `~/.local/share/cinnamon/applets` unnecessarily set as executable while they weren't in this repository ?

I believe this has been fixed in the latest versions of Cinnamon. You can mitigate this by using the `./test-spice` script in the root of the repository.

guillaume-mueller:
> You can mitigate this by using the `./test-spice` script in the root of the repository.

But from my tests with `./test-spice`, the files are copied somewhere else and I still have to copy them back to the original repository, right ?

rcalixte:
> But from my tests with `./test-spice`, I think the files are copied somewhere else and I still have to copy them back to the original repository, right ? It would be nice if the tested spice instance would use the repository files directly.

You can iterate locally and then use the script to copy/install the spice. If it's easier to develop on the test, you can copy it back manually. Most Spices don't tend to have that many files which makes it manageable. Maybe we should consider additional flags or options?

guillaume-mueller:
> This is what it does? It is just a local script that copies what is in your directory at the time of the copy. The assumption is that you have a single checkout from GitHub right?

Yes but I prefer to work in testing the code in near real-time, so a one time copy needing a few steps is ridiculously slow compared to editing the actual files and relaunching Cinnamon or even removing then adding it from the Applets menu/window.

> Maybe we should consider additional flags or options?

I tried to create a symbolic link of the `files` folder of an applet from the repository to `~/.local/share/cinnamon/applets` and it seems to work well. An obstacle though is that an applet maintainer certainly uses its own applet, so it name-clashes with the real one.

PR #7243, opened 2025-05-14
sysmon@rohan: Add a simple CPU and memory usage monitor

222rohan:
This PR adds a new applet: System Monitor (sysmon@rohan)

Features
- A simple system monitor which displays only CPU and memory usage.
- Clean Design with icons

Settings:
- Users can toggle the visibility of CPU and RAM usage through the settings.
- Update interval for data refresh can be customized.
- Display Memory in GB or % of use.

Notes:
- All strings are translatable.
- The applet ID and my GitHub username differ due to local configuration, but I am the author and maintainer of this submission. Original repository : https://github.com/222rohan/sysmon-cinnamon


rcalixte:
Do you want to run a linter over some of the files here? At least applet.js and settings-schema.json could benefit from this and it would help readability quite a bit. After that, you should run `./cinnamon-spices-makepot` in the root of the repository to create the translation template file correctly. Ask if you need a hand with anything!

222rohan:
Thank you for the help! - cleaned up the code with a linter and ran `./cinnamon-spices-makepot` . Let me know if there’s anything else I should fix.

rcalixte (approved):

rcalixte:
> Thank you for the help! - cleaned up the code with a linter and ran `./cinnamon-spices-makepot` . Let me know if there’s anything else I should fix.

This will work. Thanks for taking care of it. In the short-term, the screenshot will need a proper hyperlink so that it will work on both GitHub and the Spices website. In the longer-term, you may want to review some of the GLib calls and the GTop usage. They can be problematic.

PR #7260, opened 2025-05-18
 [notifications-enhanced@hilyxx] - Fix error with translations

Hilyxx:
- Fix error with translations
- Update settings-schema ( forgot to change it in my pull request)

### DIFF
--- a/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/po/fr.po
+++ b/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/po/fr.po
@@ -112,7 +112,7 @@ msgstr ""
 
 #. settings-schema.json->showDisturbIcon->description
 #, fuzzy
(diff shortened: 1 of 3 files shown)


Hilyxx:
I forgot to update compile-schema, so inevitably some of the translation descriptions were wrong.

rcalixte (approved):

rcalixte:
> @rcalixte I forgot to update settings-schema file, so inevitably some of the translated descriptions were wrong.

I would encourage you to use the included translation template script with the repository. In the future, there will be automation that does this so it would help minimize the onboarding when that time comes. You can selectively include or exclude files if you only want certain files pushed upstream too.

Hilyxx:
@rcalixte  I have used cinnamon-spices-makepot . But my applet was tested in my session before being modified on github and I forgot to correct some files. No excuses.

"In the future, there will be automation that does this so it would help minimize the onboarding when that time comes"

Good news.

PR #7313, opened 2025-06-09
show-hide-applets@mohammad-sn: Add an option to hide only elements until the next separator element

jogoe222:
I added this because someone on the Cinnamon Spices site had commented: 

> However, since it's function is based on what's on the Left-side of it's placement, it has rather limited use-scenarios, kinda requires us to layout the taskbar a specific way. It hides/expands icons to the Left of it, so if you use a weather-applet, you have to carefully consider where things are placed if you don't want that hidden too.

I felt the same, so I added this setting, so that the user can have more control over how much is hidden (however this means that if this setting is enabled the separator elements cannot be hidden anymore). This way elements on the left of the applet can still be shown if there's a separator element inbetween.

### DIFF
--- a/show-hide-applets@mohammad-sn/files/show-hide-applets@mohammad-sn/settings-schema.json
+++ b/show-hide-applets@mohammad-sn/files/show-hide-applets@mohammad-sn/settings-schema.json
@@ -71,6 +71,11 @@
     "dependency" : "autohiders",
     "description" : "Reshowers' autohide delay: ",
     "tooltip" : ""
+ },
+  "hideuntilseparator" : {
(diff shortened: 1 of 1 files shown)


rcalixte:
cc @mohammad-sn

claudiux:
@mohammad-sn: 0 contribution since 2019.

claudiux:
@jogoe222 
The `settings-schema.json` file is missing.

jogoe222:
@claudiux Oh damn, forgot to add it but it's there now!

claudiux (approved):

PR #7330, opened 2025-06-13
batterymonitor@pdcurtis: battery percent calculation, fix for ageing batteries

jtMUMT:
This applet previously relied on the 'capacity' file to obtain percent charge remaning. However, with my 2012 macbook pro whose battery is not capable of reaching the design spec for full charge, this capacity is incorrect. Instead, the real effective charge percentage should be calculated using the 'charge_now' and 'charge_full' files. Matches linux mint cinnamon default battery applet % calculation now.

@pdcurtis
Linux Mint 22.1
Cinnamon 6.4.8


rcalixte:
This should probably be an option rather than globally overriding the calculation for everyone.

PR #7374, opened 2025-06-25
mailnagapplet@ozderya.net: Add option to hide tray icon when there are no unread emails

victor-marino:
Mentioning @hyOzd as he's listed as the spice author [here](https://cinnamon-spices.linuxmint.com/applets/view/244), as well as @clefebvre since I can see he's responsible for much of this code.

This small PR adds the option to have the tray icon only show up when there are no unread messages, otherwise it removes itself from the panel.

I find this really handy to de-clutter the panel, as showing a persistent icon saying you have zero emails isn't very useful in my opinion. Also, having it appear only when there's a new email makes it stand out more, which I find works even better as a notification tool.

I've set its default value to `false` so that it doesn't confuse people when they update and the icon suddenly disappear.

I have also translated the strings to Spanish since that's my first language, and also corrected a small typo (`léido -> leído`).

Apologies for the polluted diff... VS Code reformatted everything on save and changed the indents, new lines, etc., so now everything shows up as modified. Hope it's not an issue 🙏

### DIFF
--- a/mailnagapplet@ozderya.net/files/mailnagapplet@ozderya.net/applet.js
+++ b/mailnagapplet@ozderya.net/files/mailnagapplet@ozderya.net/applet.js
@@ -32,7 +32,7 @@ const UUID = "mailnagapplet@ozderya.net";
 Gettext.bindtextdomain(UUID, GLib.get_home_dir() + "/.local/share/locale")
 
 function _(str) {
-  return Gettext.dgettext(UUID, str);
+	return Gettext.dgettext(UUID, str);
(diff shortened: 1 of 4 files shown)


fredcw:
> VS Code reformatted everything on save and changed the indents, new lines, etc., so now everything shows up as modified. Hope it's not an issue

Yeah it probably is. It makes it hard to see what's actually changed. Best to change VS Code settings or manually copy only the lines you've changed back into the original file. Meld is useful for that.

PR #7387, opened 2025-07-01
[Sensors@claudiux] v5.6.1: Fixes bad data in json file

claudiux:
@binarynoise This PR should solve your problem.


binarynoise:
Yes, that works, thank you!

Issue #2701, opened 2019-11-25
UUID: gputemperature@silentage.com


Padre2:
`
 * Applet version/Build date (toVersion 2.0.0 (2019-10-25 12:22:15)
 * Cinnamon version 4.4
 * Distribution - (Mint 17.2, Arch, Fedora 25, etc...)
 * Graphics hardware AMD

**Notify author of applet**
 gputemperature@silentage.com

**Issue**
Warning:  This applet contains function calls that will possibly cause Cinnamon to crash or freeze.

Some times will not load, but if loaded may cause cinnamon to crash.

Padre2:
@silentage.com
Now on cinnamon version 4.8 and have updated the metadata.json for each version, but still get warnings and crashes, if loaded, and the .Xsessionerrors file is flooded with pages  of warnings:
`Cinnamon warning: Log level 128: posix_spawn avoided (fd close requested)`

Issue #2491, opened 2019-07-29
[gputemperature@silentage.com] Cinnamon freezing every 4-5 second while using applet


mubat:
* Applet version/Build date: v.2.0.0
 * Cinnamon version: 4.0.10
 * Distribution: Mint 19.1
 * Graphics hardware: Nvidia GeForce GTX1060 with Nvidia drivers 430.26
 * [.xsession-errors](https://github.com/linuxmint/cinnamon-spices-applets/files/3440978/xsession-errors.txt)

**Issue**
All system freezing every 4-5 seconds when I use this applet. If I disable it, system continue freezing till restart. But after restart [if it not using] cinnamon works fine.

My-Random-Thoughts:
Same issue on Cinnamon 3.8.8 with Ubuntu 19.04.
A ALT+F2 then "r" fixes the problem, mostly.

luisboch:
Please, fix this issue, or, remove this applet, because it causes a lot of trouble when using it.

claudiux:
This applet has no author assigned to it.

Pxtl:
It took me far too long to figure out what was causing the performance issue on my PC, especially since the behavior is subtle-enough - it doesn't effect the mouse or the audio, so you can miss it for some time until you play like a game or a stream.  Please remove this applet from library until it can be fixed.

PR #7437, opened 2025-07-24
add warning to gputemperature@silentage.com

Pxtl:
The gputemperature applet has bad behavior, but I don't know if there are people who find it useful.  Added warning to description.

#2491 #2701

### DIFF
--- a/gputemperature@silentage.com/files/gputemperature@silentage.com/metadata.json
+++ b/gputemperature@silentage.com/files/gputemperature@silentage.com/metadata.json
@@ -1,6 +1,6 @@
 {
    "uuid": "gputemperature@silentage.com", 
    "name": "GPU Temperature", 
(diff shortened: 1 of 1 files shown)


rcalixte:
This change will break translations as-is. Can it be applied to the README instead? Are there specific reasons the issues are happening? Should it just be removed?

Pxtl:
Imho the readme isn't enough because who checks the readme before enabling a spice applet?  I just searched through the bar for a gpu temperature, found one, enabled it, and then spent days trying to figure out what the heck was happening to my computer not realizing the GPU temp thing was causing it.

Removing it would be my preference but I'm new to Cinnamon and didn't know if that was really an option.

I have no idea about root cause or what percentage of users are facing this issue, I just know it happened to me and to enough people that others are mentioning the problem in various posts online (or else I never would've been able to figure out what was causing the problem - who would ever guess the temperature monitor was causing such a problem?)

Pxtl:
Oh... has that always been there?  Maybe this should be a bit louder....

image

PR #7455, opened 2025-07-28
multicore-sys-monitor@ccadeptic23: Update .po files

mikehaertl:
Result from `./cinnamon-spices-makepot multicore-sys-monitor@ccadeptic23`.

I think this should be run from time to time to remove unused translations, right?


rcalixte:
> I think this should be run from time to time to remove unused translations, right?

The goal is that it will be automated in the future.

PR #7475, opened 2025-08-04
multicore-sys-monitor@ccadeptic23: optimize the bash script get-cpu-data3.sh

LeosBitto:
The script get-cpu-data3.sh was starting many awk processes, which made it slow and unnecessarily increased the CPU load.


claudiux (approved):

claudiux:
@LeosBitto Many thanks to you!

PR #7479, opened 2025-08-04
proxy-mode@yrtimid: new proxy mode switcher applet

yrtimiD:
new applet, allows easy switch of proxy modes directly from the panel


yrtimiD:
@claudiux , looks like you are most active recently, can you, please, take a look or advise who can?

claudiux:
@yrtimiD On Cinnamon 6.4, the schema is `system.proxy`, not `org.gnome.system.proxy`.
Please find out when this change was made and then adapt your code accordingly.

yrtimiD:
> @yrtimiD On Cinnamon 6.4, the schema is `system.proxy`, not `org.gnome.system.proxy`

hm, that's interesting, I'm right now on 6.4.10 and applet somehow works. anyway, will change. thanks!

**update**: tried, and looks like the `org.gnome.` namespace prefix is correct.
even gsettings complains if used without
image

also, dconf shows `org.gnome.system.proxy" under "/system/proxy" folder
image

so, I believe, folder is what confused you, no?

claudiux:
You're right, sure.

claudiux (approved):

yrtimiD:
@claudiux , who can run [.github/codeowners_creator.py](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/codeowners_creator.py) to update CODEOWNERS file? Or is it must be a part of "new applet PR"? (then we might want to update contribution guide)

rcalixte:
> @claudiux , who can run [.github/codeowners_creator.py](https://github.com/linuxmint/cinnamon-spices-applets/blob/master/.github/codeowners_creator.py) to update CODEOWNERS file? Or is it must be a part of "new applet PR"? (then we might want to update contribution guide)

I normally do it every so often. The plan is to have it be fully automated eventually, along with some of the translation scripts as well.

PR #7482, opened 2025-08-05
multicore-sys-monitor@ccadeptic23: optimize the bash script get-cpu-data3.sh even more

LeosBitto:
After removig many calls to awk I have removed calls to all other external commands as well, so now this script should be as fast as possible (in bash).


claudiux:
Thank you very much @LeosBitto! I'll test it soon.

PR #7495, opened 2025-08-07
notifications-enhanced@hilyxx: add filters to hide from list, or force a notif as urgent

igorsantos07:
Also:
- critical notification icon now "shakes" the bell's clapper, for a tiny animation to call out some more attention. Tried to make it entirely orange, but I guess it isn't that easy :sweat_smile: 
- includes pt_BR translation, together with some fixes / polishing to existing sentences.

> not from the author, which should be @Hilyxx :nerd_face:

### DIFF
--- a/notifications-enhanced@hilyxx/README.md
+++ b/notifications-enhanced@hilyxx/README.md
@@ -10,5 +10,6 @@ This applet is a fork of the original Notification applet developed for the Cinn
 * Tooltip now displays the number of notifications.
 * New keyboard shortcut to enable or disable notifications.
 * Middle-click on a notification to dismiss it.
+* Filters to hide irrelevant notifications from the list, or force a special app to show as urgent.
 
 ⚠️ To avoid code conflicts (as both applets share parts of the same code), please remove the vanilla Notification applet before installing Notifications Enhanced, then restart Cinnamon.
--- a/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/applet.js
+++ b/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/applet.js
@@ -58,6 +58,8 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
         this.settings.bind("ignoreTransientNotifications", "ignoreTransientNotifications");
         this.settings.bind("showEmptyTray", "showEmptyTray", this._show_hide_tray);
         this.settings.bind("showDisturbIcon", "showDisturbIcon", this._show_disturb_icon);
+        this.settings.bind("blockList", "blockList");
+        this.settings.bind("urgentList", "urgentList");
         this.settings.bind("keyOpen", "keyOpen", this._setKeybinding);
         this.settings.bind("keyClear", "keyClear", this._setKeybinding);
         this.settings.bind("keyMute", "keyMute", this._setKeybinding);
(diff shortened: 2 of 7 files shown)


Hilyxx:
That could be nice, but there are a few issues:
 
**For the "hide notifications from the list" filter:**
- Notifications don’t show up in the list as expected, but they still pop up on screen, so I don’t really see the point here.
- For some apps like flatpaks, the notification still shows up in the list. I tried with the "Fragments" app, for example.
    
**For the urgent notifications filter:**

- The icon animation should probably keep playing until the user clicks it and checks the urgent notification.
- The urgent notification doesn’t show up in the notification list.

There’s also the issue with the app name that needs to be entered in the text field. Sometimes the name shown in the menu doesn’t match the executable name. For example, for gnome-music, you launch it with 'gnome-music', but in the menu it just shows up as 'Music'. Same goes for other apps using names like 'org.gnome.something', or mintinstall.
That adds some complexity for users who aren’t familiar with that kind of setup.

Thanks again for your PR. It’s an interesting idea and I’ll definitely think about it, especially the animated icon, which is pretty cool. But I’m going to decline it for now. I’d rather not add more complexity to the code at this point, considering the issues mentioned above.

igorsantos07:
There is, indeed, a very odd inconsistency with the notifications shown or not on the list. In my use case, I mostly have notifications from the following apps:
- Flameshot: screenshot helper, which tells me when the screenshot failed or not
- blueman: for some reason, every now and then my keyboard disconnects and reconnects, but it's never really an issue for me, and I definitely don't need the history hahah
- Chrome: which never shows anything on that list, I have no clue why
- Slack: as you may guess, very important
- some other applet that warns about applet updates

The thing is: I got so scalded from the repeated and useless notifications from flameshot and blueman, cluttering the notifications widget, that I tend to miss Slack notifications mixed in that mess.
The idea is not to completely hide notifications (as this would be a task for System > Notifications, not an applet), but only hide what shows on the list, considering it a "history of things you didn't see".

That's why I came with the filtering options. Does it make sense?
It would be akin to smartphone settings that allow _us_ to define whether a specific type of notification is high or low priority, given it looks like most apps don't care for defining that (e.g. I noticed Telegram and Slack notifications both show with `is_chat: false`).

This is how my notification list usually looks like, when I'm looking for something I missed when I was away:
image

----------

On the glaring icon issue, I'm a bit confused, as I only changed the position of the bell clapper. I didn't change the icon size (which can be seen in the diff tab), so I guess this would be an existing issue, not introduced by my PR?

-------

And yep, I totally agree with the issue of not knowing the exact name to enter on that list... But I didn't come up with a better idea either. Do you think `*` placeholders (turned into a simple regex) could help?

igorsantos07:
I just copied the source from my development folder directly into `~/.local/yadda-yadda` and... it's working fine with the forced-urgent notification :thinking: 

image

Hilyxx:
@igorsantos07 or @rcalixte, please close this pull request since it won’t be merged. It adds unnecessary complexity to an issue that should be handled through cinnamon-settings rather than this applet.
Also, there’s already an option in the Flameshot settings to disable notifications.

igorsantos07:
Oh, cool. That's how open source software works, right?

You decide _"that PR adds unnecessary complexity"_ because it's actually not a feature you're interested in, so you leave other interested users stuck with their own problems, since they're not yours anyway. _"Because it should be handled by the system"_, even though extensions are made exactly because the system doesn't handle a _lot_ of things we think it should handle.

You say I could disable Flameshot notifications, as if this applet was the only place where notifications show up — as you said with _"that should be handled through cinnamon-settings"_, notifications go through the system, and I DO want them to show up, just don't want them in a history list, which _accidentally_ is exactly what this extension's interface is all about.

Does it help with my argument as to why this feature is important, and why I took my time to code it, and why I spent more time explaining why it is important and how it works with, surprisingly, 40 lines or so of "unnecessary complexity"?

> I understand that, by this time, the so-called "owner" of this extension doesn't care for whatever I have coded, my ideas or suggestions. Nonetheless, I feel it's important to show how rude and egotistical that behavior is. This is not the first time I see such ignorance around this repo... Coincidentally, both around problems related to excessive notifications that no one cares to clean, even though it's public knowledge we live in a world riddled with unnecessary notifications.

PR #7501, opened 2025-08-08
multicore-sys-monitor@ccadeptic23 - v2.3.5

claudiux:
Speed improvement.


multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/scripts/get-cpu-data3.sh
```diff
@@ -3,22 +3,21 @@
...
 sleepDurationSeconds=1
 
-declare -A prevstat
-
-while IFS=' ' read -r cpu rest
-do
-    [[ "$cpu" =~ ^cpu ]] && prevstat["$cpu"]="$rest"
-done < /proc/stat
+previousDate=$(date +%s%N | cut -b1-13)
+previousStats=$(cat /proc/stat)
```
LeosBitto, line 7: This reverts my last changes - why? Did they not work properly? Additionally, date and cut are called here just to fill an unused variable previousDate, later the same is done with another unused variable currentDate - that seems like a bug which slows down execution of this script (and increases CPU load).

LeosBitto: My version of this script is in another file "get-cpu-data3 (copie).sh" - that's probably a mistake.

claudiux: @LeosBitto Please refer to version 2.6.0, which uses a new script `get-cpu-raw-data.sh` and processes these values in `applet.js` instead of the bash script.

PR #7528, opened 2025-08-14
multicore-sys-monitor@ccadeptic23: Update German translation

mikehaertl:


multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/po/de.po
```diff
@@ -599,6 +597,11 @@ msgid ""
 "left of the circle.\n"
 "None: No Swap bar."
 msgstr ""
+"100%: Das Swap-Diagramm ist ein Balken hinter dem Kreis, der die gesamte "
+"Breite des Diagramms einnimmt\n"
+"25%: Das Swap-Diagramm ist ein Balken links vom Kreis, der 25% der Breite "
```
mikehaertl, line 602: @claudiux Please note that this is wrong in the original. The option is "25%" but the hint uses "10%".
I've already corrected this here.

PR #7529, opened 2025-08-14
Add envycontrol applet

zamszowy:
Initial commit with the applet for https://github.com/bayasdev/envycontrol


rcalixte (approved):
@zamszowy Overall, great work as always. Two suggestions:
1. For the visual strings, you should consider making them translatable (and adding a translation template).
2. For the `/bin/bash` invocations, it should be `/usr/bin/bash` and the header in the script should probably be `/usr/bin/env bash` instead of `/bin/bash`.

Neither of these are blockers per se but they would make the applet a bit of a better fit for more people.

Issue #7542, opened 2025-08-17
apt-updates@zamszowy: FR: support hold packages


yrtimiD:
### Applet name and maintainer

apt-updates@zamszowy @zamszowy

### What would you like to see?

Thank you for useful applet, I'd be happy to see support for hold packages, i.e. packages explicitly set to ignore updates with `apt-mark hold <package_name>`
currently `apt list --upgradable` doesn't respects hold packages, so "has updates" status icon will always be on.

minor recommendation: instead of running extra "sleep" process you can use native `Mainloop.timeout_add_seconds()` function or, even better, [`util.setInterval()`](https://github.com/linuxmint/cinnamon-spices-applets/blob/1c5db553b9f9006aa17eeeef47593e529b6a5b7e/.typescript-declarations/cinnamon/misc/util.d.ts#L193)

zamszowy:
@yrtimiD functionality is now moved into updates-notifier@zamszowy applet, which should now skip hold packages, now that it uses `pkcon` instead of `apt`.
Thanks also for code suggestions - applied.

yrtimiD:
checked today, hold packages are shows as "Blocked" by `pkcon get-updates`, so check script still counts them
```
$ pkcon get-updates
Getting updates               [=========================]         
Loading cache                 [=========================]         
Finished                      [=========================]         
Blocked     	libnvidia-compute-535-535.230.02-0ubuntu0.20.04.1.i386 (ubuntu-focal-security-restricted)	NVIDIA libcompute package
Blocked     	libnvidia-decode-535-535.230.02-0ubuntu0.20.04.1.i386 (ubuntu-focal-security-restricted)	NVIDIA Video Decoding runtime libraries
Blocked     	libnvidia-encode-535-535.230.02-0ubuntu0.20.04.1.i386 (ubuntu-focal-security-restricted)	NVENC Video Encoding runtime library
Blocked     	libnvidia-fbc1-535-535.230.02-0ubuntu0.20.04.1.i386 (ubuntu-focal-security-restricted)	NVIDIA OpenGL-based Framebuffer Capture runtime library
```

zamszowy:
@yrtimiD, #7648 should fixed this, although I couldn't reproduce it on my Debian machine - held packages were filtered by  the `pkcon` itself.

PR #7552, opened 2025-08-22
apt-updates@zamszowy: rework into updates-notifier@zamszowy

zamszowy:
Rework apt-updates@zamszowy into applet that no longer depends on `apt`, but is using high-level abstraction layer via `pkcon` from `packagekit` to check/update system packages.


zamszowy:
implements #7542

rcalixte (approved):

PR #7599, opened 2025-08-26
Update CODEOWNERS and templates

yrtimiD:
many new applets changed
@rcalixte, please click the button ;)


yrtimiD:
right, there is also "bug_report_creator"... my bad. next time will not miss it.

rcalixte:
> right, there is also "bug_report_creator"... my bad. next time will not miss it.

I would say not to worry about it too much. 😄 

I tend to run them whenever but it's a stopgap until the automation is done. I try to avoid too much churn unless something actually comes up.

PR #7620, opened 2025-08-29
mic-level@zamszowy: don't measure when option disabled

zamszowy:
This update features the following:
 - add translation support
 - fixes running measurement even when not set to do so on applet startup
 - add timeout to rec command to avoid rare situation when it hangs


mic-level@zamszowy/files/mic-level@zamszowy/po/ca.po
```diff
@@ -1,14 +1,14 @@
-# SOME DESCRIPTIVE TITLE.
+# MICROPHONE LEVEL MONITOR
 # This file is put in the public domain.
-# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+# zamszowy, 2024
 #
 #, fuzzy
-msgid ""
+sgid ""
```
rcalixte:
```suggestion
msgid ""
```

zamszowy: Forgot to run validation. :sweat_smile:  Regenerated with `poedit`.

PR #7625, opened 2025-08-30
mouse-keys-shortcut@GabrielCamara3526: Improves sliders

claudiux:
@GabrielCamara3526
Your applet is a real gem!

This PR improves the sliders:

- Setting a mark for each default value.
- Translating messages, adding current value and unit.
- Making magnetic each default value.

I hope you'll like this PR.


GabrielCamara3526:
Wouldn't be possible without your inspiration and support. I'm glad you like it!

PR #7630, opened 2025-08-30
test-spice: add watch option

yrtimiD:
adds -w / --watch option to the test-spice tool allowing faster checking of applet changes
if watch is requested, next will happen:
1. applet is validated and installed (as in default mode)
2. applet uuid folder is monitored with inotifywait tool (standard inotify-tools package must be installed)
3. on each file change under uuid folder validate, copy and reload are performed (no more than once in a 1 sec)

### DIFF
--- a/test-spice
+++ b/test-spice
@@ -6,6 +6,7 @@ import os
 import sys
 import shutil
 import subprocess
+import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


yrtimiD:
I'm not very familiar with python, will appreciate any corrections

PR #7666, opened 2025-09-04
updates-notifier: fix TextDecoder error on old cjs

yrtimiD:
looks like on cjs@4.4.0 TextDecoder is not yet available, but old-good String(byte[]) constructor works fine.

Implemented fallback to String ctor if TextDecoder is not available. Guess for such a simple output as pkcon produces String constructor should be good enough.

Bonus: moved updates count title calculation into the info-window.js and slightly simplified "view" script.

owner: @zamszowy

### DIFF
--- a/updates-notifier@zamszowy/files/updates-notifier@zamszowy/info-window.js
+++ b/updates-notifier@zamszowy/files/updates-notifier@zamszowy/info-window.js
@@ -176,22 +176,25 @@ function showDetails(item) {
 Gtk.init(null);
 
 // main window
-let win = new Gtk.Window({ title: ARGV[0] });
+let win = new Gtk.Window({ title: "Updates" });
 win.set_default_size(640, 640);
 
 let scroll = new Gtk.ScrolledWindow();
 let listbox = new Gtk.ListBox();
 
-// parse updates from ARGV[1]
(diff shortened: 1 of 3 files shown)


rcalixte (approved):

updates-notifier@zamszowy/files/updates-notifier@zamszowy/info-window.js
```diff
@@ -176,22 +176,25 @@ function showDetails(item) {
 Gtk.init(null);
 
 // main window
-let win = new Gtk.Window({ title: ARGV[0] });
+let win = new Gtk.Window({ title: "Updates" });
```
rcalixte, line 179: This should probably be translated.

rcalixte:
Finger slipped on the button, please disregard and sorry for the noise.

updates-notifier@zamszowy/files/updates-notifier@zamszowy/info-window.js
yrtimiD: Actually there are more untranslated strings in this specific window. Maybe in the next versions me or the author will address them.

rcalixte (approved):

PR #7689, opened 2025-09-07
powerman@dr.drummie: Add new applet

drdrummie:
powerman@dr.drummie: Add new applet

Upgraded version of the original Cinnamon power applet with added automation. For laptops, it automatically switches brightness and power profiles when changing between AC and battery, includes low‑battery protection, customizable settings, and offers an option to replace the default power applet or hide this one.
While partially based on existing code, it introduces additional functionality and customization features.

All usage instructions are provided in the README.md.


fredcw:
@drdrummie Since this applet and power@cinnamon.org applet are both single instance and ~conflict with each other~ cannot both run at the same time, do you think it would be possible to put the conflicting code into cinnamon's JS code so that multiple power applets could access the same info at the same time. That way someone could have multiple power applets on different monitors and both power@cinnamon.org and this applet could be made multi instance? I don't know if it's possible or even a good idea, I just wondering if you have any thoughts about that.

fredcw:
@claudiux Sorry, I didn't mean to imply that this applet conflicts with power@cinnamon.org in that it causes a problem or that it shouldn't be merged, I just meant that both applets cannot run at the same time. From the readme, It seems that this applet just replaces the power@cinnamon.org applet.

Unfortunately, I can't test this applet because I don't have a laptop.

drdrummie:
@fredcw 
I believe they can run at the same time, you should be able to hide powerman if don't want it into system tray. 
I saw some battery applets but didn't see the option to automatically switch power mode or brightness. 
I've tried my own implementation, but power applet seemed as the right way to go.. Also didn't found that automation option in Mint so tried to made it with this applet.. 
I've tested it on 3 different laptops, haven't see the issues, still - my own are older ones so I can't really tell how it behaves on modern systems..

claudiux:
It is possible to notify the users that the `power@cinnamon.org` applet is also in use and give them the option to disable it. See the code from `sound150@claudiux`, which detects `sound@cinnamon.org`.

drdrummie:
Sure, no problem, I'll look into it..
Applet already handles removal of original power applet, so I need to see about notification..

claudiux (approved):

PR #7713, opened 2025-09-11
timer@Severga: Improve shutdown

claudiux:
@Severga 

This simple change makes shutting down the computer more efficient.

With the old method, my computer no longer shuts down.


Severga:
Ok! 👍

PR #7716, opened 2025-09-11
quick-settings@celiopy: add applet

celiopy:


fredcw:
This is a great applet. I found a couple of errors though. I have Manjaro cinnamon so powerprofilesctl is not installed by default and it's a desktop PC so I don't need the powerprofiles anyway. It's not a problem but some people may think that the button doesn't work.

Clicking on the "prevent sleep" button, I get an error in ~/.xsession-errors

```
(cinnamon:1922997): Cjs-CRITICAL **: 20:41:04.410: JS ERROR: TypeError: this.preventSleepToggle.button is undefined
_togglePreventSleep/<@/home/fred/.local/share/cinnamon/applets/quick-settings@celiopy/applet.js:732:21
asyncCallback@resource:///org/cinnamon/cjs/modules/core/overrides/Gio.js:118:22
```

Also, I use larger font so the "Prevent Sleep" button label is cut short "Prevent S..."

celiopy:
Hey, thank you for trying it out!
Been a while since I last update it.

I will make sure it hides the power toggle if power profiles is not installed. (As well of other toggles that isn't useful, the idea is to make it modular)

I will have to look into why this error happens when you click prevent sleep, seems like it does not define the button, not the action...

The trimming is intended, so if the text gets too big, the box itself doesn't grow. So it trims the text. If you change text scaling factor (or use the attached "slider" in the menu, you can see when it trims), but as you font size is already larger it trims from the get go. I might increase the width limit so it doesn't trim early. But the main idea is so it doesn't affect the look of the menu.

PR #7717, opened 2025-09-11
signout@kayfo: adds new applet

kayfo:
This applet adds an easy and obvious sign out button on the panel. Useful in public computing environments like libraries, schools, and computer labs.

### DIFF
--- a/signout@kayfo/files/signout@kayfo/po/signout@kayfo.pot
+++ b/signout@kayfo/files/signout@kayfo/po/signout@kayfo.pot
@@ -1,0 +1,34 @@
+# SIGNOUT@KAYFO
+# This file is put in the public domain.
+# kayfo, 2025
+#
+#, fuzzy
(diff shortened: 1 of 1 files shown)


kayfo:
Thanks for the feedback! 
- stylesheet and settings schema deleted
- Hide configuration added
- metadata description enhanced with vertical panel disclaimer and more specific use case.

This applet is intended for public computing environments like public libraries where users are not tech savvy at all and would not recognize a typical sign out icon. The "Sign out" label must remain for this use case. I'm putting together tools and a configuration guide that makes it easier for low-tech public libraries to deploy secure public computer kiosks. Having the applet as an official spice would make this more easily repeatable.

fredcw:
Works fine. It automatically can't be added to vertical panels (the add to panel button is greyed out in applet settings).

rcalixte (approved):

PR #7718, opened 2025-09-11
Add back better-backgrounds@simonmicro applet with new author

fredcw:
@simonmicro has given permission on his github page: https://github.com/simonmicro/better-backgrounds


fredcw:
@rcalixte The number of stars (score) on the spices website has reset back to zero. I checked what it was before is was removed and it was 20 stars. Is there any way to put that back? The comments are still the same so I don't know why the score has changed.

rcalixte:
> @rcalixte The number of stars (score) on the spices website has reset back to zero. I checked what it was before is was removed and it was 20 stars. Is there any way to put that back?

I would guess something in a database but that's just my speculation. There are comments but the Spice ID was reset. My guess it's just data in the database. Maybe @clefebvre has an idea?

PR #7727, opened 2025-09-12
New applets to complement Maximus extension

hanspr:
Maximus extensión allows to fully maximize any window without the title bar.

Under this situation the user loses information about the windows name and the close, maximize, minimize buttons.

maximus-title : displays a window title so a user can identify any active window including the one maximized

maximus-buttons : adds virtual window buttons on the status bar to allow the user to interact as with the regular title bar buttons.

### DIFF
--- a/maximus-buttons@hanspr/files/maximus-buttons@hanspr/applet.js
+++ b/maximus-buttons@hanspr/files/maximus-buttons@hanspr/applet.js
@@ -39,7 +39,7 @@ class MyApplet extends Applet.TextIconApplet {
         this.button = []
         this.createButtons(this.buttons_style)
         this.on_panel_edit_mode_changed
-        this._windowChange(global.display.focus_window)
+        this._showButtons(global.display.focus_window)
     }
 
     bindSettings() {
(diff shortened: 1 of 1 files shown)


hanspr:
Pardon my ignorance, I have read the readme file a couple of times and I can't figure out, if I'm following the correct process to add a new applet.

I would like to know, if:

* you have a predefined cycle to review new applet requests?. If so, when would the next revision cycle be?
* I'm aware that after my first pull request I've modified the applet, so does this pushes the reviews in someway forward?. If so for how long should the pull request be without new commits, so is considered for review?
* I could confidently say that the applet is stable and functional, at least in my environment. So is there a tag, or version number to add, to notify that the applet at least from my point of view is ready for a review?
* When you review an new applet, do you notify if you consider it of low value to add it?. So I could simply close this request? (I would not feel offended, I understand that not everything that we want to publish brings a real value to Cinnamon desktop).

Thank you for your consideration and hope you could solve some of my doubts about this process, possibly document in the readme file for others.

fredcw:
@rcalixte @claudiux

PR #7751, opened 2025-09-15
show-my-ip@triyadi.my.id: new applet

uwaiscode:
Simple panel applet that shows IPv4 addresses for every network interface.
Tested on Cinnamon 6.x.


fredcw:
Doesn't work. It just says ": undefined" on the panel. 

`output` is empty after `let output = String.fromCharCode.apply(null, out).trim();`

PR #7806, opened 2025-09-23
updates-notifier@zamszowy: switch to DBus monitoring of updates

zamszowy:
Instead of parsing pkcon output, listen to PackageKit DBus signals
to get notified about new updates.
Parsing "pkcon get-updates"" output to get input for "pkcon
get-update-details" is unreliable and will not work for all the
packages, especially the ones with complex naming/versioning.

Since with Dbus we now have more control over returned details of the
update, we can make the info window more user-friendly with different
formatting and extra description.


rcalixte:
One additional implementation detail: you can use the user's default terminal rather than relying on something being hard-coded. Look at how Nemo implements the "Open In Terminal" functionality for an idea.

zamszowy:
> One additional implementation detail: you can use the user's default terminal rather than relying on something being hard-coded. Look at how Nemo implements the "Open In Terminal" functionality for an idea.

Thanks for the hint, I'll add that in the next version.

PR #7810, opened 2025-09-24
updates-notifier@zamszowy: add option to show also firmware updates

zamszowy:


rcalixte:
I _think_ fwupd has a DBus interface too. 😅

Also ,great job on this applet. There's definitely a few open issues asking precisely for this. I'll go through the backlog and close them out some time. Thanks!

zamszowy:
> I _think_ fwupd has a DBus interface too. 😅

It has and I tried to use it, but it seems that it's not intended for monitoring of the available updates. But I'll check it more thoroughly some time later to be sure.

PR #7817, opened 2025-09-26
Exit v1.8.3: Adds the option "Screen On/Off with Lock feature" …

claudiux:
…to the menu and its shortcut to the settings


dragonleopardpig:
Downloaded the applet (version 2.0.0), with Screen Lock option checked, Mouse click on icon will Turn Off monitor + Screen Lock. However, with keyboard shortcut: Turn Off Monitor + NO Screen Lock. Can help look into that? Running on Linux Mint 22.2. Thank you.

PR #7820, opened 2025-09-27
darkMode: (Theme does not update after resuming from suspend (fix)

lobster4all:
@linuxedo

This PR fixes an issue where the dark/light theme is not updated if the scheduled time for the change occurs while the computer is suspended or off.

**Problem:**
The applet uses a `Mainloop.timeout` to schedule the next theme change. If the system is suspended, this timer does not account for the elapsed time, and the theme change is missed upon resume.

**Solution:**
This change adds a listener to the `org.freedesktop.login1` D-Bus service for the `PrepareForSleep` signal. When the system resumes from sleep (`PrepareForSleep` emits `false`), it triggers the `change_mode_automatically()` function, ensuring the correct theme is applied based on the current time.

This makes the applet's automatic mode more reliable for users who frequently suspend their machines.


fredcw:
@slgobinath

PR #7850, opened 2025-10-04
`ddcci-monitor-control@andr35`: Allow 0% brightess/contrast

FreshwaterShrimp:
Let's the user set brightness/contrast to 0%, previously 1% was the minimum and setting it to 0% would have no effect.


fredcw:
@andr35

andr35:
👍

PR #7871, opened 2025-10-11
cinnamon-timer@jake1164: Update outdated URL

MaxF1:
Fix outdated link to g_date_time_format() in GLib documentation.


rcalixte:
> Fix outdated link to g_date_time_format() in GLib documentation.

This was well-intentioned but please reference the README for the proper scripts to use to automate the management of the translation files. Modifying them by hand is problematic. In this case, you did not modify the source strings and as a result, this could have broken the very translations you meant to fix.

PR #7884, opened 2025-10-15
power-profiles@rcalixte: Add scroll wheel action support

anaximeno:
Allow to change profiles by using scroll wheel action on the panel icon.

cc @rcalixte


rcalixte:
My initial reaction is that I don't know if this is necessary.

anaximeno:
Just a quality of life improvement, I've been cycling between power modes frequently depending if I am connected or not with my laptop to the power supply, so I was thinking that it would be useful if I could just scroll to the next profile instead of having to open the menu, though, understandable if you don't see it as necessary overall.

I also tend to use more the mouse or touchpad than keyboard shortcuts for panel stuff, specially to avoid increasing the amount of shortcuts I have to memorize in the system.

rcalixte (approved):

PR #7912, opened 2025-10-24
Add Chinese translation

gov360:
Sensors@claudiux/files/Sensors@claudiux/po/cn.po


gov360:
files/SpicesUpdate@claudiux/po/zh_CN.po

claudiux (approved):

PR #7916, opened 2025-10-26
HomeAssistant@MatthewGodding - v1.0.0: New applet

matthewgodding:
This new applet switches https://www.home-assistant.io/ entities such as lights. I personally have found this useful to quickly switch my hue play monitor light on and off. I plan to expand with more features however I've been using it for a few weeks and have found no issues so wish to publish to a larger audience.


HomeAssistant@MatthewGodding/files/HomeAssistant@MatthewGodding/applet.js
```diff
@@ -0,0 +1,380 @@
...
+              try { this._logToFile && this._logToFile('_addConfigureMenu: configureItem added'); } catch (e) {}
+              try { log('_addConfigureMenu: configureItem added'); } catch (e) {}
+         } catch (e) {
+              // fall back silently if menu creation fails on some environments
+         }
+     },
+
+         // Request helper: use curl via GLib.spawn_command_line_sync and return
+         // callback(err, parsedResponseOrBody)
+         _request: function(method, path, body, callback) {
```
Centurix, line 195: There's a library called Soup in `gi` that can handle requests. Should be examples in other applets that use `imports.gi.Soup`, might save rolling your own implementation.

Centurix:
I like this idea. Tested it and it works. Nice!

My 2 cents on this as I'd like to see it working and in mainline:

* Support more than one entity ID
* Config moved into a `settings-schema.json` configuration window rather than editing `metadata.json`
* Move config out of `metadata.json` because the data will be wiped when the applet is updated.

HomeAssistant@MatthewGodding/files/HomeAssistant@MatthewGodding/applet.js
```diff
@@ -0,0 +1,380 @@
...
+          // Settings.AppletSettings will load persisted values. Remove hard-coded
+          // tokens/IDs to avoid accidental leaks.
+
+          this.set_applet_icon_symbolic_name("network-server");
+          this.set_applet_tooltip("Home Assistant: " + this.entityId);
+
+          this._state = null;
+
+          // Network session removed: use curl fallback only (works reliably
+          // across environments). Any libsoup-specific code was removed.
```
Centurix, line 41: libsoup is a dependency of Cinnamon, shouldn't be any concern here

matthewgodding:
> I like this idea. Tested it and it works. Nice!
> 
> My 2 cents on this as I'd like to see it working and in mainline:
> 
> * Support more than one entity ID
> * Config moved into a `settings-schema.json` configuration window rather than editing `metadata.json`
> * Move config out of `metadata.json` because the data will be wiped when the applet is updated.

Thanks!

Yes, agree with these points.

Will need to think about the best approach for multiple entities, would it be multiple buttons, a list to select from etc? From a home assistant perspective you could use a scene or automation to switch multiple things.

It would also be nice in the future to give the Home Assistant URL and token then pull back a list of devices and entities to select from.

PR #7924, opened 2025-10-29
sound150 v13.4.0: Improved tooltip duration

### LINKED ISSUE #7922: [sound150@claudiux] Tooltip disappears after a very short time
### Applet version/Build date

13.3.4 (2025-10-28 19:06:18)

### Cinnamon version

6.4.8

### Distribution

Linux Mint 22.2 Cinnamon

_No response_

sound150@claudiux @claudiux

### What happened?

When moving the mouse cursor over the applet, the tooltip is shown, but it disappears after what feels one second.

IMHO the tooltip should stay on until the cursor leaves the applet or one of the menus is shown.

I believe this behaviour has been introduced with the latest update to v13.3.4 (but I might be wrong).

### Other information

_No response_

claudiux:
Fixes #7922.

### DIFF
--- a/sound150@claudiux/files/sound150@claudiux/6.4/applet.js
+++ b/sound150@claudiux/files/sound150@claudiux/6.4/applet.js
@@ -46,6 +46,8 @@ const {
     timeout_add_seconds,
     setTimeout,
     clearTimeout,
+    setInterval,
+    clearInterval,
     source_remove,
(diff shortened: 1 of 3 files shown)


sphh:
Excellent! Now I get all the time to read the tooltip, because it only closes when I move the cursor away or click on the icon. Thanks.

PR #7929, opened 2025-10-30
simple-timer@naheller v1.0.0: New applet

naheller:
This is a small timer applet that focuses on basic functionality and a simple interface. While there are other timer applets in the Cinnamon ecosystem, I found them to be complicated with extra features and confusing layouts.

I hope this might be useful for others who desire a minimal alternative.

timer-screenshot

PR #7941, opened 2025-11-03
simple-timer@naheller v1.0.0: New applet

naheller:
This is a small timer applet that focuses on basic functionality and a simple interface. While there are other timer applets in the Cinnamon ecosystem, I found them to be complicated with extra features.

I hope this might be useful for others who desire a minimal alternative.

timer-screenshot


naheller:
@claudiux Thanks for your review on my PR #7929, which I closed due to a mistake I made with the git history.

I've opened this new PR with the file structure changes you called for. Please let me know if anything looks out of place.

PR #7964, opened 2025-11-12
sound150@claudiux - update es.po

haggen88:


rcalixte:
This looks like it landed in the wrong folder. Do you want to move it into the subdirectory? (I forgot I put the check in place as well. 😅)

PR #7967, opened 2025-11-13
Direct v1.3.0: Adds Favorite Apps Section

claudiux:
As requested by @bach1590 on https://cinnamon-spices.linuxmint.com/applets/view/422#count.
A very good idea, indeed!
The Favorite Apps Section can be set at the bottom of the _Favorites_ tab of this applet settings.


bach1590:
Thank you so much for this and i upvoted too.  Much appreciated.

PR #7981, opened 2025-11-17
weather@mockturtl: Fix log spam, convert hourly summary into a tooltip.

mtwebster:
The summary label was disabled in https://github.com/linuxmint/cinnamon-spices-applets/commit/c57b27467bcd3943b39fcafc540a6ab26ec054c0 but not completely removed, and was still throwing allocation warnings during updates.

(cinnamon:177153): St-CRITICAL **: 16:17:23.909: st_widget_get_theme_node called on the widget [0x5a935feae560 StLabel.hourly-data ("...")] which is not in the stage.
== Stack trace for context 0x5a935e0f5130 ==
...
...
...
...

This removes the unused actor entirely, along with all references to it, and adds a tooltip for the hourly icon that will display the weather summary on hover.

_________
edit: Gave up with the build, using the build.sh script, or simply `npx webpack`, seems to ignore the es2017 in the tsconfig and wants to make a lot of compatibility changes (let/const -> var, etc...). Simply running `tsc` in the src/3.8 folder does not seem  to do this (or would not, if the build didn't fail for other reasons).

Since my changes were pretty isolated it was simple enough to pick just those for the commit here.

I don't really care about adding the summary as a tooltip, I just wanted to fix the log spam, and felt bad about simply discarding that info. This seemed a simple alternative.

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
@@ -18808,6 +18808,7 @@ class UIForecasts {
 const { PolicyType } = imports.gi.Gtk;
 const { ScrollDirection } = imports.gi.Clutter;
 const { addTween } = imports.ui.tweener;
+const { Tooltip: uiHourlyForecasts_Tooltip } = imports.ui.tooltips;
 const { BoxLayout: uiHourlyForecasts_BoxLayout, Side, ScrollView, Icon: uiHourlyForecasts_Icon, Align: uiHourlyForecasts_Align } = imports.gi.St;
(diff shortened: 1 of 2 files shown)


fredcw:
Ping @Gr3q

Gr3q (approved):
as for build, you need to do from the weather@mockturtl folder

1. `npm i` once
2. `npm watch` while developing
3. I used `./test.sh` that extracts it to the correct folder and restarts only the applet.

`build*.sh` and `install.sh` should be deleted in the future, they were created when I was not using webpack for building. the build guide should be changed to reflect that or deleted.

Edit: It really should be done when I relinquish my maintainer status (which I should've done already)

PR #7999, opened 2025-11-25
sshlauncher@sumo: Add setting to toggle SSH connection notifications

ChrisB85:


rcalixte:
@ChrisB85 If you used a third-party tool for the translations, it would be best to leave them to formal translators. There is also tooling in the repository to regenerate the translation template if any strings are updated so that translation files do not need to be manually curated.

cc @sumo

ChrisB85:
@rcalixte Yes, I used AI to create the translations (which, based on my experience, gives good results). Should I revert the commit with the translations in order for this PR to be merged?

fredcw:
AI translations are pretty good, especially if you give it the context. I think this is a tech we should be taking advantage of And even if every translation isn't 100% perfect, it's still a lot better than nothing, and most languages won't get translated at all in most applets otherwise. It saves work for translators as well bc they only have to fix the odd translation that they think could be better worded.

PR #8004, opened 2025-11-27
 azan@fahri.nurul.id: Added an option for extra reminders before prayer time

defaultUser822:


fredcw:
I haven't tested but looks OK.

defaultUser822:
> I haven't tested but looks OK.

I did test it, and it was a bit of a pain to change the time system back and forth (for the purpose of testing). But it works. I will start testing on my machine as I do on a day to day basis. Just for to be very sure.

defaultUser822:
Update: Works just as intended

azan@fahri.nurul.id/files/azan@fahri.nurul.id/applet.js
```diff
@@ -296,6 +308,9 @@ AzanApplet.prototype = {
                         nearestPrayerId = prayerId;
                     }
 
+                    if (diffMinutes == extraReminderTime && extraReminderTime > 0) {
+                        Main.notify(_(extraReminderTime + " Minutes untill " + this._timeNames[nearestPrayerId]));
+                    }
```
rcalixte, line 312: A few things here:

* This string should be translatable since it is user facing.
* How does this condition work if the machine goes to hibernate? Will it send on resume if the limit hits during that time? Should it?
* Later in this file, some whitespace was modified that can probably be reverted while making other changes in this file. Please try to be mindful.

Thank you!

defaultUser822: Alright, I will make it translatable through editing the azan@fahri.nurul.id.pot file. I didn't know about that. 
About the hibernation thing, I haven't tested. How should I test it? As far as I know, Cinnamon doesn't restore sessions after hibernation. It shouldn't notify after hibernation.

defaultUser822: I updated the .pot file.

defaultUser822: I updated the .pot file (again), but I made sure that the references to the lines are correct. I tested the applet while sleeping -hibernation doesn't work- the reminders don't show up after sleep, which is intended behaviour.

defaultUser822:
Hello, it's been quite a while since I opened this PR (4 months). Are there any remaining problems to be resolved?

rcalixte:
cc @fahrinh

PR #8006, opened 2025-11-28
Fix path for httpd-vhosts.conf in applet.js

hadsonmoraes:
fix location extra/httpd-vhosts.conf


fredcw:
LGTM. opt/lampp/etc/extra/httpd-vhosts.conf is the proper location according to ChatGPT.

rcalixte (approved):

PR #8011, opened 2025-11-29
Update German translations for notifications

Kachelkaiser:


rcalixte:
@Kachelkaiser Take a look at some recent translation pull requests. You used the non-default branch and the incorrect path. Additionally, it is strongly suggested to use one of the translation tools like Poedit rather than manually editing the translation files. Ask if you need a hand. Thank you!

PR #8034, opened 2025-12-03
panel-copy@mostlynick3: Add new applet

mostlynick3:
Adds an applet that lets users copy their panels to new locations or monitors. Tested on Cinnamon 6.x


fredcw:
This doesn't check to see if an applet is multi instance before copying it.

The following default applets are single instance: inhibit, network, notifications, on-screen-keyboard, power, slideshow, sound, systray, xrandr.

I don't know why all these are single instance but presumably one or both instances will not work correctly. For instance, with the notifications applet, the notifications will only appear in one of the instances. Many 3rd party applets are also single instance.

mostlynick3:
Thanks for your comment, @fredcw - I haven't tested all single instance applets but haven't found any issues with network, power, sound, systray, or inhibit. These applets work identically across all panels with the applet.

My goal with this applet is to add identical panels to secondary monitors, so I can fully manage my system no matter which monitor's dock I'm using. Seeing as I don't have any issues with the single instance applets I use, I would prefer not to exclude them altogether. With your comment in mind, I believe it would then be most reasonable to add a config setting to enable or disable cloning single instance applets. What do you think?

fredcw:
I just tried copying a panel with your applet and cinnamon crashed.

I think if an applet is single instance, it's probably because there's a good reason and copying it anyway could cause system instability or crashes.

Some of the default applets may be easy to make multi-instance or not, I'm not sure. One thing I noticed is that with the inhibit applet, turning off/on power management in one instance does not flip the switch in the other instance. Also, with systray applet, I have clipIt app, but it only appears in one of the instances. The sound applet also disappeared from both panels.

I think it's too risky to copy all single instance applets.

mostlynick3:
I've now barred single instance applets from being copied, would very much appreciate it if you could give it another go @fredcw!

fredcw:
Yeah that works great except for some 3rd party applets e.g. sysmonitor@orcus because it's looking for metadata.json file in the versioned directory "3.2"

Also, I get this in ~/.xsession-errors

```(cinnamon:1922997): Cjs-WARNING **: 14:39:29.306: Some code called array.toString() on a Uint8Array instance. Previously this would have interpreted the bytes of the array as a string, but that is nonstandard. In the future this will return the bytes as comma-separated digits. For the time being, the old behavior has been preserved, but please fix your code anyway to use TextDecoder.
(Note that array.toString() may have been called implicitly.)
0 _isMultipleInstanceApplet() ["/usr/share/cinnamon/js/misc/fileUtils.js line 211 > Function":186:40]
```

maybe try JS's TextDecoder?

mostlynick3:
Thanks for your feedback @fredcw, I have decided to rework this applet into an extension with your input in mind. This will let users automatically set up cloning of panels to new monitors, instead of manually copying each time from an applet. I'll PR the extension to the extension repo once I'm satisfied with it

PR #7705, opened 2025-09-09
better-backgrounds@simonmicro Removed my own applet, as it become broken and is unmaintained.

simonmicro:
It were three lucky years, but since I have no time to fix it anymore...

_So Long, and Thanks for All the Fish!_

PR #8048, opened 2025-12-08
Remove better-backgrounds applet

fredcw:
Useful features are available in other applets and only two of the 5 sources still work.


Cobinja:
Registered code owner is @simonmicro

fredcw:
It's recently changed to me: #7705
I was hoping to fix some of the issues but it turns out that 3 of the sources no longer provide free pictures, and I could only find Bing and picsum.photos as reliable sources. And tbh, I don't really have the time to maintain this applet.

Edit: Actually, maybe I'll keep it for a while, the webpage rendering feature might be useful.

Cobinja:
Ah, sry, missed that. The Codeowners file (.github/CODEOWNERS) still says simonmicro

fredcw:
I changed my mind again, I don't have time to maintain this.

rcalixte:
> I changed my mind again, I don't have time to maintain this.

If you're not the originator, isn't the proper course of action to leave it unmaintained? I agree this is a bit of a saturated area but if it's still providing some value, that's probably okay?

fredcw:
@rcalixte Well the original author did delete it (https://github.com/linuxmint/cinnamon-spices-applets/pull/7705) but I added it back. There are actually a couple of features not found in other applets. I'll take another look at it. xD

PR #8054, opened 2025-12-10
Betterlock : use the user-defined symbolic icon size

pilatomic:
Hello !

This PR makes the BetterLock applet use the user-selected symbolic panel icon size, rather than a hard-coded value, so that the size of the applet is consistent with the other applets.

Tested on Linux Mint 22.2 (Cinnamon 6.4.8) : Applet loads and works fine at the correct size, and change size like the other applets when changing the symbolic icon size of the panel, and logs show no issue whatsoever.


pilatomic:
@Centurix

rcalixte (approved):

PR #8073, opened 2025-12-14
Betterlock : Add more settings and code cleanup

### LINKED ISSUE #1297: [betterlock]can you add "Hide icons if inactive" option ?
<UUID: betterlock>
```
 * Applet version/Build date (to be found in "About.." or in "metadata.json")
 * Cinnamon version (cinnamon 3.4.6)
 * Distribution - (Fedora 26)
 * Graphics hardware *and* driver used: Intel Corporate HD Graphics 5500
 * No author

There is option in GNOME variant of this applet. It is "Hide icons if inactive". It mean that if CapsLock key is inactive, then it's icon will not appear at all in panel. While if this key activated by user, then it's icon will appear in panel. The same for NumLock key. 

Please notice that this option differ from the already added option of "Chose which indicator to show".

Best.

### LINKED ISSUE #3921: betterlock: Feature Request - Readonly-Mode
Hi,
My english is poor. I hope you understand my request.
Your App is very nice.
Is it possible to add a Readonly-Mode?
When I click on the Icons in the Taskbar the Modus will be changed.
But want only to see the Status and don't want to change it by clicking.
It will be fine, when in the options was a Checkbox that will be turn off the status change by clicking the Icons with the Mouse.
I only want to see the Status.

The reason is, my taskbar pops up, when my Mouse is on the bottom of the screen.
When I click on a Button in a Window that is near the bottom of the screen, the Taskbar will pop up and I'm clicking on the app, not on the Button.
I've 2 displays and the space is always to less. ;)

thx n greets

pilatomic:
Good evening !

After my previous (tiny) PR on Betterlock, I felt that I could invest more time in this applet, since it really fits my use case (wireless keyboard with no indicators).

What this PR includes, in (mostly) chronological order : 
- Add a setting to react or not to click on indicator (closes #3921)
- Make Caps Lock and Num Lock indicator visible by default (as it is likely the most common use case). Scroll Lock does not seem mapped in most Keyboard maps anyway.
- Add a setting to hide inactive indicators (closes #1297)
- Add sections to the settings, to distinguish indicators settings from notification settings.
- Code cleanup : get rid of the _firstRun variable, move each action (notify, set visibility, set icon) to its own function, compare simple booleans instead of comparing the displayed item. I previously had added conditions to run each action only if it produced an effect (no need to show() an already visible indicator), but benchmarking showed absolutely no reduction in CPU usage, so I dropped it in favor of code simplicity.
- Also (I almost forgot to mention it), I added French translation for the new settings

I am not sure if I did the translations correctly, as the settings are not translated when I try the applet with "test-spice betterlock".
Is it expected ?

Other than that it works perfectly.

@Centurix, if you can have a look :-)

I understand the code cleanup part (with no other benefit than subjectively cleaner code) might be a bit controversial, so if we think that's unneeded, I can remove it, that's why I did not attempt to squash the commits together.

### DIFF
--- a/betterlock/files/betterlock/metadata.json
+++ b/betterlock/files/betterlock/metadata.json
@@ -1,5 +1,6 @@
 {
     "description": "Shows whether caps lock/num lock/scr lock is on, and displays a notification when these change.",
     "uuid": "betterlock",
-    "name": "Lock keys indicator with notifications"
+    "name": "Lock keys indicator with notifications",
+    "version": "1.1.0"
 }


pilatomic:
Found a small issue with the NumLock OSD message, converting to draft until I fix it.

Centurix:
No problems, looking good so far, I'll wait till this comes out of draft and take another look. Thanks!

claudiux:
@Centurix

rcalixte (approved):

Issue #6704, opened 2024-12-23
Closing the 'Add new entry' or 'Edit entry' windows in the QuickLauncher settings crashes the CinnVIIStarkMenu settings


RenegadeScooter:
### Applet version/Build date

1.32 (2024-07-20 08:13:57)

### Cinnamon version

6.2.9

### Distribution

Mint 22

### Graphics hardware *and* driver used

NVIDIA GeForce RTX 3060 Ti, nvidia-driver-550

### Applet name and maintainer

CinnVIIStarkMenu@NikoKrause @fredcw

### What happened?

See title. This bug makes it impossible to customize QuickLauncher Places or Applications.

This bug was mentioned in the comments on [this applet's Cinnamon Spices page](https://cinnamon-spices.linuxmint.com/applets/view/281) ~7 months ago, but apparently has still not been fixed yet.

fredcw:
Turned out to be packaging issue in arch based distros. No longer present in cinnamon 6.4 + later.

Issue can be closed.

PR #8081, opened 2025-12-15
CinnVIIStarkMenu@NikoKrause: fixed applet crashing when adding null icon quicklauncher items

NotSirius-A:
Hello, 

@fredcw 

Fixes an issue that crashes the applet:

Steps to reproduce:
- Add a new quicklauncher application entry, without specifying an icon
- Restart cinnamon/reboot etc.
- Applet crashes because with the error `Argument icon_name may not be null`

From what I understand this error is referring to an St.Icon argument `icon_name` on line 1075. Similar incorrect behavior occurs when a quicklauncher places (also null icon) entry is created, but this doesn't not result in a crash, just renders an error icon.

I was actually trying to fix issue #6704, but I can't reproduce this behavior, is it fixed?

### DIFF
--- a/CinnVIIStarkMenu@NikoKrause/files/CinnVIIStarkMenu@NikoKrause/4.2/applet.js
+++ b/CinnVIIStarkMenu@NikoKrause/files/CinnVIIStarkMenu@NikoKrause/4.2/applet.js
@@ -2257,7 +2257,7 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 
             // Quicklauncher places icon
             let quicklauncher_places_icon = this.quicklauncher_places[i].icon;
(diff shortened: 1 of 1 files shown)


fredcw:
Thanks. However, I think I prefer `if (!quicklauncher_apps_icon)`, it's less accurate but more concise I guess.

Regarding #6704, I can't reproduce it anymore either. I think it was due to a packaging error in arch and derived distros because it happened in Manjaro but not Mint.

NotSirius-A:
Sure, I've changed it to `if (!quicklauncher_apps_icon)`.

fredcw:
@rcalixte @claudiux This can be merged. Thanks.

PR #8102, opened 2025-12-20
app-drawer@mostlynick3: Fix multi-monitor app drawer and element bleed, add features requested in 8119

mostlynick3:
Fixes a bug where the app drawer opens in the middle of the virtual resolution split between multiple monitors (now it's confined to the screen that app drawer is opened from) and prevents elements out of view from being clickable


mostlynick3:
Updating PR to include feature request in https://github.com/linuxmint/cinnamon-spices-applets/issues/8119

claudiux (approved):

PR #8122, opened 2025-12-27
22.2-menu@luka: a fork of the old menu with some extra settings, Adding new applet

WerIstLuka:
this is the default cinnamon menu from before 22.3 with some extra settings
-toggle to show/hide buttons (lock screen, logout, shut down)
-toggle to show/hide application names in the bottom right when hovering over them
-toggle to show/hide application description
-toggle to lock the menu size
-toggle to make the menu infinitely resizable
-button to resize the menu to my preffered size
-button to restart cinnamon (needed for the lock menu size toggle)

these settings are in a new settings tab

this was forked from cinnamon 6.4.6

### DIFF
--- a/22.2-menu@luka/files/22.2-menu@luka/applet.js
+++ b/22.2-menu@luka/files/22.2-menu@luka/applet.js
@@ -2715,9 +2715,8 @@ class CinnamonMenuApplet extends Applet.TextIconApplet {
 	            this.menu.close();
 	            this._session.LogoutRemote(0);
 	        };
+            this.systemButtonsBox.add(button.actor, { y_align: St.Align.END, y_fill: false });
 	    };
-
-        this.systemButtonsBox.add(button.actor, { y_align: St.Align.END, y_fill: false });
 
         //Shutdown button
         if (this.show_quit) {


fredcw:
xD, looks like we had the same idea. https://github.com/linuxmint/cinnamon-spices-applets/pull/8120

I guess, I'll close mine as there's no point having two of them and your's has added options.

WerIstLuka:
@fredcw didnt see yours, i checked a few days ago and saw nothing so i did it myself
i looked at your pr and saw you have a fix for a scroll bug, i have so few categories that i never experienced that but i'll add your fix into this pr

if theres anything you want to add let me know

fredcw:
You might want to make it translatable as well. It's easy enough, just add the following to the top of applet.js:

```
const Gettext = imports.gettext;

Gettext.bindtextdomain(__meta.uuid, GLib.get_home_dir() + '/.local/share/locale');

function _(str) {
    let cinnamonTranslation = Gettext.gettext(str);
    if (cinnamonTranslation !== str) {
        return cinnamonTranslation;
    }
    return Gettext.dgettext('22.2-menu@luka', str);
}
```
then run `cinnamon-xlet-makepot 22.2-menu@luka` in applets directory containing the 22.2-menu@luka directory. That will add the `po` directory and the .pot file.

fredcw:
oops, sorry, I edited my comment and swapped the line

`return Gettext.dgettext('Cinnamenu@json', str);`

for

`return Gettext.dgettext('22.2-menu@luka', str);`

which is what it should be xD

mtwebster:
We don't need yet another fork do we? This is ridiculous.

WerIstLuka:
@mtweobster it also has extra options

anaximeno:
> @mtweobster it also has extra options

would be better contributing to the already merged menu to avoid duplication specially if the changes aren't that drastically different.

PR #8146, opened 2026-01-03
fixed few bugs to ui, logic, language updates and added new ui.

iMayhem:
- new black ui
- added option to change language in settings 
- made the compact mode default.
- a bug fixed related to a user mentioned some alignment issue in reddit - [](https://www.reddit.com/r/linuxmint/comments/1pzdhnm/comment/nx5259e/?context=1)
- added option so the user can see previous days network usage
- added usage history option in panel
- added option to reset today's stats 
- now user can manually turn off daily network usage logging in settings if they feel their system is lagging in older devices ( the logs are kept locally of how much internet is getting used ) they will not be sent over network to anyone
- fixed some issues related to ui, where internet speeds ( KB / MB ) where getting out of the panel.
- made the compact mode more compact so if anyone have lots of applet, then it will be easy to accomodate.
- added user guide for easy to understand how the app works.


claudiux:
Please use the same method as for all other applets to translate messages.

iMayhem:
@claudiux Hi, i have updated the code.

PR #8152, opened 2026-01-05
multicore-sys-monitor v3.2.2: Takes into account HiDPI display

claudiux:
As requested by @paranoidi on https://cinnamon-spices.linuxmint.com/applets/view/79#count


paranoidi:
@claudiux Sorry to report back, but this is even more broken now.

image

This is with "Limit height .." at general tab off. All of the panel icons get offset out of the screen.

With "Limit height .." on:

image

Then the CPU graph is broken. My preference is to have it always at full panel height, it was working fine before these recent changes.

I believe you're also involved maintaining "download-and-upload-speed" as well. You can see it being broken too. The texts do not align properly.

Edit: Hmm. Seems to align now after clearing custom CSS and adding it back.

claudiux:
About "download-and-upload-speed": I'm not the maintainer. I've created [Bps@claudiux](https://cinnamon-spices.linuxmint.com/applets/view/405) to replace it.

> Seems to align now after clearing custom CSS and adding it back.

How did you do it? What "custom CSS"?

paranoidi:
image

Just removed that one to see if it was the cause of the issue. And then added it back. Not sure why removing it and adding it back would have made any difference.

PR #8171, opened 2026-01-09
[timer@Severga] Bug fix

Severga:
Small display bug fix for "Set period" click while time is "00:00:00"


rcalixte:
Do you want to update the translation files as well?

Severga:
@rcalixte, there is no need. I didn't change any text, only pure code.

rcalixte:
> @rcalixte, there is no need. I didn't change any text, only pure code.

Right, I just saw it too. My mistake. I thought it was a new string but it matches an existing one.

PR #8195, opened 2026-01-12
Hot Fix Last Update Applet [fix bug 8190]

Skullckoor:
Hot Fix Last Update Applet [fix bug 8190]


hanspr:
@Skullckoor 

I don't think this is the right way to do this, you should send the PR on my my fork so we discus if it is the right solution. I'm about to send the proper solution to this in a moment.

https://github.com/hanspr/cinnamon-spices-applets

Skullckoor:
@hanspr Sorry, I've just never used Git.
I made a new fork.

PR #8196, opened 2026-01-12
fix regression

### LINKED ISSUE #8190: maximus-buttons@hanspr icon bug
### Applet version/Build date

0.0.3 (2026-01-12 16:52:57)

### Cinnamon version

6.4.8

### Distribution

Mint 22.2

_No response_

maximus-buttons@hanspr @hanspr

### What happened?

You're right. I used two panels, but that's not the cause of the problem.
I tried completely uninstalling and reinstalling the applet. Using just one panel results in the same problem.
TS window:
Image

About TS:
Image

Click to **Desktop** (**Рабочий стол**)
Image
The buttons turned off, but the icon remained.

When launching Brotato (screenshots above), things happen differently:
A single window opens. Its title is Godot, which has no icon (the default one is used).
Then (after the final launch), the title changes to Brotato and an icon is added.
However, the applet only picks up the icon after switching elsewhere and returning to the Brotato window.
After clicking on the desktop (Рабочий стол), the Brotato icon remains.

i.e. The window is the same, but it doesn't have its own icon when launched, loading it later.
The applet doesn't track changes to the **current** window's icon.
Overall, this isn't a big problem.
The only real problem is that when you select a desktop, the previous icon remains unchanged.

I think the icon needs the same mechanism as titles.

### Other information

_No response_

hanspr:
Fix #8190 

Regression introduced on previous PR
Fixed correct icon on multiple monitors
Removed warning undefined title window in maximu-title

### DIFF
--- a/maximus-title@hanspr/files/maximus-title@hanspr/applet.js
+++ b/maximus-title@hanspr/files/maximus-title@hanspr/applet.js
@@ -84,6 +84,9 @@ class MyApplet extends Applet.TextIconApplet {
         } catch (e) {
             return
         }
+        if (w == undefined) {
+            return
+        }
         if (w.get_monitor() != this.panel.monitorIndex) {
             let title = this._getTopWindowFromMonitor(w.get_monitor())
             this._setTitle(title, "")


rcalixte:
@hanspr When you submit pull requests, can you start using a different branch for each? Using the master branch is leading to a messy commit history. Take a look at other pull requests if you need to see an example.

hanspr:
Yes your right, I will in the future, sorry about that

PR #8198, opened 2026-01-12
Radio3.0 v3.10.2: Repairs Configuration menu for Cinnamon 6.6

claudiux:
as requested by @tonywhelan on https://cinnamon-spices.linuxmint.com/applets/view/360#count


tonywhelan:
Thanks for the super-fast fix!

PR #8212, opened 2026-01-16
Storage activity LED

lucutus8255:
It emulates on the screen  HDD / SDD LED activity for computers without one


claudiux:
File structure

A spice can contain many files, but it should have the following file structure:

    UUID/
    UUID/info.json
    UUID/screenshot.png
    UUID/README.md
    UUID/files/
    UUID/files/UUID
    UUID/files/UUID/metadata.json
    UUID/files/UUID/applet.js
    UUID/files/UUID/icon.png

There are two important directories:

    UUID/ is the root level directory, it includes files which are used by the website and on github.
    UUID/files/ represents the content of the ZIP archive which users can download from https://cinnamon-spices.linuxmint.com/ or which is sent to Cinnamon when installing the spice from System Settings. This is the content which is interpreted by Cinnamon itself.

As you can see, the content of the spice isn't placed inside UUID/files/ directly, but inside UUID/files/UUID/ instead. This guarantees files aren't extracted directly onto the file system, but placed in the proper UUID directory. The presence of this UUID directory, inside of files/ isn't actually needed by Cinnamon (as Cinnamon creates it if it's missing), but it is needed to guarantee a proper manual installation (i.e. when users download the ZIP from the Cinnamon Spices website).

Important note:

    The UUID/files/ directory has to be "empty", which means that it should contain ONLY the UUID directory. Else the spice won't be installable through System Settings.

At the root level:

    info.json contains information about the spice. For instance, this is the file which contains the github username of the spice's author.
    screenshot.png is a screenshot of the spice in action.
    README.md is optional and can be used to show instructions and information about the spice. It appears both in Github and on the website.

claudiux:
`info.json` is missing.

claudiux:
Same as #8216. `info.json` is still missing.

PR #8216, opened 2026-01-19
Add Mass Storage Activity LED applet

lucutus8255:


claudiux:
`info.json` is still missing.

claudiux:
```
[storage-act-led@mrbartblog] Error during validation: [storage-act-led@mrbartblog] Forbidden unicode characters in field 'author' in files/storage-act-led@mrbartblog/metadata.json
```

PR #8226, opened 2026-01-20
Fade Monitors: Applet for Fade Monitors script

hisovereign:


rcalixte:
@hisovereign Please follow the file layout for applets. Take a look at the other applets if you need an example.

PR #8243, opened 2026-01-26
Fix: add backward compatibility for older Cinnamon versions

lucutus8255:
- Added a helper function `to_string` to handle differences in how ByteArray is processed in older Cinnamon versions (e.g., 4.2 in Mint 19.2).
- This ensures the applet works on both newer systems (like Mint 22) and legacy ones by detecting the available conversion method.


claudiux:
@lucutus8255 
Please rebase then check file structure: https://github.com/linuxmint/cinnamon-spices-applets/blob/master/README.md#file-structure

PR #8255, opened 2026-01-30
directory-menu@torchipeppo: Add native version with Wayland support

anaximeno:
This adds a Native to Cinnamon version of the applet with the main goal of allowing Wayland support for the applet. In X11 sessions this native implementation is accessible and usable as well (the user must change to it in settings), but the default remains the previous Gtk based implementation now labeled as "classic". For Wayland sessions the native version is set to run instead, but this behavior can also be disabled from the settings.

ref: https://github.com/torchipeppo/directory-menu-for-cinnamon/issues/2 (I've discussed this with the author in their personal repo before sending this PR)

I'll also ping @torchipeppo here for their review.

#### Video example of how it should work

https://github.com/user-attachments/assets/197e0443-af82-4006-ba8d-1d9f9c2de6ae

### DIFF
--- a/directory-menu@torchipeppo/files/directory-menu@torchipeppo/metadata.json
+++ b/directory-menu@torchipeppo/files/directory-menu@torchipeppo/metadata.json
@@ -2,7 +2,7 @@
     "uuid": "directory-menu@torchipeppo",
     "name": "Directory Menu",
     "description": "Quickly navigate through directories on your system using a dropdown menu.",
-    "version": "0.9.0",
+    "version": "1.0.0",
     "author": "torchipeppo",
     "max-instances": "-1"
 }


torchipeppo:
Hi, thank you very much for your work.
I tried a quick test-spice on my Mint 21.3 machine to confirm the GTK behavior remains the same, and it does, but it looks like I can't open the settings.
It may still be something dumb on my end. I'll be looking into it some other time.

Apart from that, it's all great! I appreciate the fix to quit unmapped menu processes too, I hadn't noticed that in my earlier pass.

rcalixte:
> but it looks like I can't open the settings

Sometimes things can get wedged and require a Cinnamon restart. If that doesn't work, do you want to hold off on merging this?

anaximeno:
> Apart from that, it's all great! I appreciate the fix to quit unmapped menu processes too, I hadn't noticed that in my earlier pass.

I also commented a line regarding the `event.free()` I noticed it came from the xapps c code, but shouldn't be required on Python since it handles the free automatically with the garbage collector, and I think this might've been the reason I've been having for some time seeing stack traces about a python process from the applet popup_menu.py script appearing in the reports app.

image

It didn't happen anymore after that (I can send these fixes on a different PR if this is holded for now).

anaximeno:
Regarding the settings not opening issue, consider checking the logs with either the Cinnamon Looking Glass app or reading the `~/.xsession-errors` file, if you find something relevant, alert me and I'll fix it.

torchipeppo:
Hi, sorry for the delay.

I can't see any logs despite multiple Cinnamon restarts, but it seems like the issue only affects the test-spice version. Applying your changes to my development copy of the applet results in everything working correctly, including the settings.

Actually, even the test-spice of the current version of the applet doesn't show the settings either, so this is probably a glitch contained within devtest.

So, I'm optimistic that the merge will be okay. Let's go with it and see if anything happens.

torchipeppo:
Downloaded the update, can confirm everything is in order. Thanks a lot to @anaximeno for your work!

PR #8261, opened 2026-01-31
CPU/GPU/RAM Viewer new applet

LYK4S5:
A minimalist and high-performance applet to monitor CPU, GPU, and RAM usage in real-time.


LYK4S5:
Update RSSDock applet.js, fixed refresh time problem

PR #8270, opened 2026-02-02
Add Moderately Advanced RAM Monitor

notoyeah:
Hello, 
I would like to submit a new applet: **Moderately Advanced RAM Monitor**.

**Features:**
- Reads data directly from `/proc/meminfo` (no external processes).
- Visual stability: Optional Monospace font and custom padding to prevent panel "jumping".
- Blinking alerts for high usage or low availability.
- Customizable click action to launch a system monitor.
- Support for GiB (binary) and GB (decimal) units.

The applet includes full Polish and English translations.


mtwebster:
Wait I don't know that synchronously reading is a good idea here. 

```
            let [, contents] = file.load_contents(null);
```
You should use `load_contents_async()/_finish()`.

claudiux:
@mtwebster Is `GLib.file_get_contents(filename)` preferable, or should we use `load_contents_async()/_finish()`?

mtwebster:
They're pretty much the same, just one is a function that takes a path argument and the other is a Gio.File method - but if it doesn't end in _async, or have a callback argument, it's synchronous and can block the UI if the file is inaccessible (like being actively written to by another process).

We should ALWAYS try to use asynchronous IO - I thought this was already being checked for during PR validation.

notoyeah:
Thanks for the reply. This is my first applet ever. I'll read up on it and try to improve it.

PR #8274, opened 2026-02-03
Moderately Advanced RAM Monitor: Switch to async reading

notoyeah:
Following up on #8270.

As suggested by @mtwebster, I have updated the file reading logic 
to be asynchronous (`load_contents_async`). 

This ensures the applet is non-blocking. I've also refreshed 
the .pot and .po files to align with the code changes.


notoyeah:
I've just pushed the updates. The applet now uses load_contents_async as suggested, and I've also synchronized it with the recent translation updates and bumped the version. Ready for review!

notoyeah:
The new version doesn't work properly. I'm closing this pull request.

PR #8293, opened 2026-02-08
Adapat to new alocation API

Sunderland93:
Muffin part: https://github.com/linuxmint/muffin/pull/793

### DIFF
--- a/SpicesUpdate@claudiux/files/SpicesUpdate@claudiux/5.6/SU_messageTray.js
+++ b/SpicesUpdate@claudiux/files/SpicesUpdate@claudiux/5.6/SU_messageTray.js
@@ -853,7 +853,7 @@ Notification.prototype = {
         }
     },
 
-    _bannerBoxAllocate: function(actor, box) {
+    _bannerBoxAllocate: function(actor, box, flags) {
         let availWidth = box.x2 - box.x1;
 
         let [titleMinW, titleNatW] = this._titleLabel.get_preferred_width(-1);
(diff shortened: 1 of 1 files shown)


claudiux:
Perhaps check the version of Cinnamon before applying the new API?

fredcw:
@Sunderland93 This won't work unfortunately, you would need to version the applets. For example with sticky@scollins, you'd need to create a new directory `sticky@scollins/files/sticky@scollins/6.8` and copy all the files from `sticky@scollins/files/sticky@scollins/3.2` into `/6.8` and make the changes to the version in `/6.8`

An alternative way would be to use `version_exceeds()` in `util.js`

For example in `sticky@scollins/files/sticky@scollins/3.2/applet.js` 
```javascript
allocate(actor, box, flags) { 
    let childBox = new Clutter.ActorBox();

    childBox.x1 = box.x1;
    childBox.x2 = box.x2;
    childBox.y1 = box.y1;
    childBox.y2 = box.y2;

    const CINNAMON_VERSION = imports.misc.config.PACKAGE_VERSION;
    if (Util.version_exceeds(CINNAMON_VERSION, "6.8")) {
        this.textBox.allocate(childBox);
    } else {
        this.textBox.allocate(childBox, flags);
    }
}
```

The `allocate(actor, box, flags) {` line should be fine as `flags` would just be `undefined` in cinnamon 6.8
also, `Util.version_exceeds` should really be called `version_same_or_exceeds` or `version_at_least`

PR #8306, opened 2026-02-12
sound150 v14.1.0: Resize the menu by grabbing its edges with the mouse

claudiux:
As requested by @TheNort12 on https://cinnamon-spices.linuxmint.com/applets/view/306#count.

### DIFF
--- a/sound150@claudiux/files/sound150@claudiux/6.6/applet.js
+++ b/sound150@claudiux/files/sound150@claudiux/6.6/applet.js
@@ -222,11 +222,9 @@ class Sound150Applet extends Applet.TextIconApplet {
         this.alreadyCalledBysetAppletTooltip = false;
 
         this.real_ui_scale = 1.0;
-        this.menuWidth = this.userMenuWidth; //520;
         Util.spawnCommandLineAsyncIO(PATH2SCRIPTS + "/get-real-scale.py", (stdout, stderr, exitCode) => {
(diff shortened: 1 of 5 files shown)


fredcw:
@claudiux I don't think resizing the menu actor will work very well because cinnamon wants to size all the widgets itself, including the menu actor size itself. The only thing in the applet window that you can dynamically resize without cinnamon trying to determine it's own size for it is the size of the coverart picture. all other widgets and the menu actor size itself will then automatically resize around that. I don't know the structure of the applet but I think you need something like this:

```
_onBoxResized(width, height) {
        change_coverart_size(width, height);
        if (!this._resizer.resizingInProgress) {
           //~ this.settings.popup_width = width / this.real_ui_scale;
           //~ this.settings.popup_height = height / this.real_ui_scale;
           this.popup_width = Math.max( Math.round(width / this.real_ui_scale), 450 );
           this.popup_height = Math.round(height / this.real_ui_scale);
        }
    }
```
    
If you can make a function like change_coverart_size() that can dynamically resize the current cover art, I think that should work well.

claudiux:
@fredcw 
The goal is not really to be able to increase the size of the album cover, but rather to be able to expand the menu in order to display the counters properly.
I'll think about it, because rebuilding a new image with each mouse movement once the edge has been captured can have a significant computational cost.

PR #8312, opened 2026-02-13
weather@mockturtl: Fix touchpad hourly forecast scroll behaviour

### LINKED ISSUE #6484: Hourly forecast doesn't scroll correctly when using a touchpad
### Applet version/Build date

3.6.4

### Cinnamon version

6.2.9

### Distribution

Mint 22, LMDE 6

_No response_

weather@mockturtl @Gr3q

### What happened?

Using a laptop's touchpad to scroll the hourly forecast on the weather applet isn't handled correctly resulting in the forecast always trying to scroll all the way to the right. I'm not exactly sure how to describe it any better than that, so I've included videos to help demonstrate. I've tested this on an amd framework 13 and an old thinkpad t61, the same behavior occurred on both. (It also effected the thinkpad's trackpoint suggesting this might be a problem with anything other than a traditional up-down scroll wheel.)

### Other information

Video with the Framework: https://github.com/user-attachments/assets/1bf4b324-3e5a-4a7d-a676-6401ab5e5e00
Video with the ThinkPad: https://github.com/user-attachments/assets/ddb64885-e2f0-43ba-9ae9-e72cf2948fa0

### LINKED ISSUE #6706: Applet Bug: Using left/right click on mouse-wheel does not function for scroll-wheel left click.
### Applet version/Build date

3.6.4 (2024-11-23 18:11:41)

### Cinnamon version

6.0.4

### Distribution

NixOS 24.05.20241205.190c31a (Uakari) x86_64

NVIDIA GeForce GT 710

weather@mockturtl @Gr3q

### What happened?

While using the scroll-wheel up/down works to scroll the hourly forecast view left/right, when using the scroll-wheel left/right it only works for scrolling right, while using left it behaves in what looks like an "off-by-one" error, it jumps left but returns to the right. In fact if you click left repeatedly, the view will actually scroll to the right.

### Other information

_No response_

Matszwe02:
Fix scroll on hourly forecast by only handling up/down scroll direction, horizontal scroll is being handled automatically.

Fixes #6484 , fixes #6706

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
@@ -18897,7 +18897,7 @@ class UIHourlyForecasts {
             const adjustment = hScroll.get_adjustment();
             const direction = event.get_scroll_direction();
             const newVal = adjustment.get_value() +
-                (direction === ScrollDirection.UP ? -adjustment.step_increment : adjustment.step_increment);
+                ((direction === ScrollDirection.UP) ? -adjustment.step_increment : (direction === ScrollDirection.DOWN) ? adjustment.step_increment : 0);
             if (global.settings.get_boolean("desktop-effects-on-menus"))
                 addTween(adjustment, { value: newVal, time: 0.25 });
             else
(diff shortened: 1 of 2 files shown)


rcalixte:
cc @Gr3q

Matszwe02:
> Embarrassing, especially because I didn't catch this when I was checking the code for those issues

Yeah, looks like it was written when cinnamon scroll had only 2 directions.
Thanks for merging

PR #8320, opened 2026-02-15
CurtainExtension: Lightweight and Fast alternative to window manager on Cinnamon, initial release

AdvaitaSoni:
# Curtain Applet for Cinnamon
## Introduction
The **Curtain** applet for Cinnamon is designed to be a lightweight,fast and easy to install and configure alternative to popular window managers. Essentially, it is aimed to completely navigate workspaces and windows using only the keyboard.

## Checks done
- [x] passed the validation script 
- [x] tested locally
- [x] tested on applications and mentioned usage in README.md
- [x] provided icon 

## Preferences
Due to individual windows not following **Curtain** rules(please see README.md) ,it is advisable the following applications are preferred: 
- kitty over gnome-terminal

### DIFF
--- a/curtain@AdvaitaSoni/files/curtain@AdvaitaSoni/metadata.json
+++ b/curtain@AdvaitaSoni/files/curtain@AdvaitaSoni/metadata.json
@@ -5,6 +5,5 @@
     "description": "Tiles your windows using algorithm",
     "version": "0.0.1",
     "url": "https://github.com/AdvaitaSoni/curtain/tree/APPLET",
-    "author": "AdvaitaSoni",
-    "last-edited": 1760819814
+    "author": "AdvaitaSoni"
 }


AdvaitaSoni:
Hey @claudiux, please review the PR whenever you are free. Thanks!!

AdvaitaSoni:
Thanks! once again

claudiux:
@AdvaitaSoni 
Great applet!
Missing option: Do not create a new workspace when using \<Super>T.
I have 4 workspaces. When I am on the 4th workspace and use \<Super>T, I do not want to create a 5th, but switch to the 1st.

mtwebster:
This is an extension, not an applet, and should not have been approved.

AdvaitaSoni:
> This is an extension, not an applet, and should not have been approved.

@mtwebster Can you explain why it shouldn't be an applet? Originally i designed this to be an extension only but made it an applet because that way one can see the current and total workspaces

mtwebster:
You're using an applet simply as an entry point to replace/intercept core Cinnamon behavior. This can cause unintended side effects that result in confused users and confused bug reports, and is not stable from one Cinnamon release to the next.

We try to preserve third-party Applet and Desklet behavior across releases, even fixing them ourselves if there's something we just can't avoid. With extensions, we're not binding ourselves to this rule - when I'm making internal changes to Cinnamon I don't check if I'm breaking them.

One general issue I noticed - xlets should not override existing default keybindings (like super-l, -p, -s).

I'm not against this being an extension but I am against it being presented as an applet.

AdvaitaSoni:
> You're using an applet simply as an entry point to replace/intercept core Cinnamon behavior. This can cause unintended side effects that result in confused users and confused bug reports, and is not stable from one Cinnamon release to the next.
> 
> We try to preserve third-party Applet and Desklet behavior across releases, even fixing them ourselves if there's something we just can't avoid. With extensions, we're not binding ourselves to this rule - when I'm making internal changes to Cinnamon I don't check if I'm breaking them.
> 
> One general issue I noticed - xlets should not override existing default keybindings (like super-l, -p, -s).
> 
> I'm not against this being an extension but I am against it being presented as an applet.

Understood @mtwebster. It makes sense that it may break from one cinnamon version to another, which violates the stability offered by Cinnamon(especially for Linux Mint users).
However this applet mostly uses APIs that have been part of the documentation for a long time. Also while I can't promise that I will be able to look after the applet itself, i do plan to mention the compatibility issues and their solutions. 
And while I do overwrite some existing hotkeys, they are configurable by the user themselves and the whole point of applet is to use these hotkeys to do window management.

I have no problem with what you suggested however seeing the current and total workspaces on the panel is an essential experience to **Curtain** as a window manager. Here's some options i suggest: 

# Option 1 
We remove this applet altogether and only use it as an extension. In this case the users will not be able to see the current and total workspaces as well as whether the extension is enabled or not. 

# Option 2
We remove the core functionality of the applet but it still shows the current and total workspaces but not the status of the extension  

# Option 3 
We modify the panel from the extension itself. However this may cause even more confusion as user will think of it as an applet even thought it is an extension

# Option 4
We follow option 2 but control the "enabled" status of the applet from the extension itself. - I don't know if it is possible or not though

Personally I think Option 4 sounds to be much better if we can achieve it. What are your thoughts?

mtwebster:
Option 1 all the way

- Users have plenty of other options to see what workspace they're on.
- An extension is enabled/disabled in the Extension management settings. If you need to tell the user this fact or allow it to be toggled, send a notification.

> And while I do overwrite some existing hotkeys, they are configurable by the user themselves and the whole point of applet is to use these hotkeys to do window management.

Choose different defaults or leave them empty by default. Users can figure it out when they configure the extension.

> However this applet mostly uses APIs that have been part of the documentation for a long time. 

Unfortunately this is due to the fact the UI is written in javascript. We can't prevent you doing/using certain things programatically but that doesn't make it ok or right. [this garbage](https://github.com/linuxmint/cinnamon-spices-applets/pull/8320/changes#diff-6f1a7649ccb4eaa3829b7c0eac83f2257e0d7bb8672d7892e5b5a1173a95e3b8R328-R329
), for instance.

AdvaitaSoni:
Thanks for your feedback @mtwebster!!! will definitely open a PR for the extension. Deleting this branch for now.

Issue #8314, opened 2026-02-14
radio@driglu4it: Support UTF-8 characters


futurk:
### Applet version/Build date

2.2.4

### Cinnamon version

6.6.7

### Distribution

Mint 22.3

### Graphics hardware *and* driver used

Intel Corporation Raptor Lake-P [Iris Xe Graphics]

### Applet name and maintainer

radio@driglu4it

### What happened?

Awesome extension. But, song titles break -probably- because the encoding is set to ASCII.

### Other information

![Image](https://github.com/user-attachments/assets/d5afb7e1-8d1d-4321-af76-e0d56aa55b9b)

In this example, `ü` (UTF-8) renders as `Ã¼` (ASCII)

hthienloc:
@futurk I've worked on a solution for this and submitted a pull request. Once it is approved and finalized, the problem should be resolved. I'll keep you updated.

hthienloc:
Hey @futurk! Could you check your mpv version by running `mpv --version` in the terminal?

The maintainer noted that this encoding issue was fixed in mpv **v0.37.0**. If you're on an older version, upgrading mpv might already resolve it for you.

If you're already on v0.37.0+ and the issue persists, it would be a huge help if you could **leave a comment on PR #8355** confirming. That kind of feedback really helps the maintainer prioritize and merge the fix.

futurk:
> Hey [@futurk](https://github.com/futurk)! Could you check your mpv version by running `mpv --version` in the terminal?
> 
> The maintainer noted that this encoding issue was fixed in mpv **v0.37.0**. If you're on an older version, upgrading mpv might already resolve it for you.
> 
> If you're already on v0.37.0+ and the issue persists, it would be a huge help if you could **leave a comment on PR [#8355](https://github.com/linuxmint/cinnamon-spices-applets/pull/8355)** confirming. That kind of feedback really helps the maintainer prioritize and merge the fix.

Yes I am currently on 0.37.0 and the issue **persists**:

<img width="862" height="489" alt="Image" src="https://github.com/user-attachments/assets/865d1895-a7d6-4b18-8eab-c81d37eef871" />

at least for the letters `ü` and `ö`

```bash
$ mpv --version
mpv 0.37.0 Copyright © 2000-2023 mpv/MPlayer/mplayer2 projects
libplacebo version: v6.338.2
FFmpeg version: 6.1.1-3ubuntu5
FFmpeg library versions:
   libavutil       58.29.100
[...]

PR #8355, opened 2026-02-21
radio@driglu4it: support UTF‑8 metadata

hthienloc:
- Add `decodeMetadataString()` helper to repair titles that have been mis‑decoded as Latin‑1 by the mpris plugin.
- Apply the helper when processing xesam:title and in `getCurrentTitle()`.
- Extend Mpvock to accept an initial title; add three unit tests verifying broken titles are fixed and well‑formed titles stay untouched.

This resolves issue #8314 where non‑ASCII song names rendered as Ã¼ etc.

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/6.4/radio-applet.js
+++ b/radio@driglu4it/files/radio@driglu4it/6.4/radio-applet.js
@@ -1272,9 +1272,9 @@ function equalArrays(array, other, bitmask, customizer, equalFunc, stack) {
 
 
 /** Built-in value references. */
-var Uint8Array = _root.Uint8Array;
+var _Uint8Array_Uint8Array = _root.Uint8Array;
 
-/* harmony default export */ const _Uint8Array = (Uint8Array);
+/* harmony default export */ const _Uint8Array = (_Uint8Array_Uint8Array);
 
 ;// ./node_modules/lodash-es/_mapToArray.js
 /**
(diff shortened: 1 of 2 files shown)


fredcw:
@jonath92

jonath92:
@hthienloc thank you very much for the contribution but I am wondering if we really need this. AFAICS this is an issue in mpv which should be fixed in newer versions. I opened for this an issue at mpv 4 years ago: https://github.com/mpv-player/mpv/issues/8844. It should be fixed in mpv version v0.37.0 and above. I am also often listening to german radio stations and I know that I had this bug in the past but I haven't seen this issue any more since years and I am using mpv 0.37.0. 

But it could of course be that there is another issue which is not related to the issue I mentioned above. In that case I would definitely approve your pr. So can you please double check that the issue you fix is not already fixed by mpv v.037.0 or above?

hthienloc:
@jonath92 yes, the original mpv bug should be gone in 0.37+, but we should keep the decoder as a fallback for users stuck on older versions or for streams that are simply mis‑encoded.

jonath92:
@hthienloc sorry for not answering before. I disagree with the fallback approach because: 
- We engage users to use outdated versions of mpv which is usually not a good idea. I don't know why some people have an outdated version - maybe it is a problem in some package mangers / repositories but it is not the task of the applet to remedy this problem. I would even call it counterproductive, as it hides the actual problem: having outdated versions.
- Every line of code in the end also needs to be maintained. It might work right now but who knows if the code also works in 5 years? For example, I see that you used "escape" in your code which is according to mozilla deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape. So I think there is a change that the function gets removed at some time in future from the javascript runtime cinnamon is using. I don't know if this happens - maybe it also stays forever - but who knows? It would be annoying to break the app just because of a fallback

So imo it would be enough to add the information on the Readme

hthienloc:
@jonath92 Let me first check with @futurk what version of mpv they are using - that should clarify whether this is indeed an mpv issue or something else.

hthienloc:
@jonath92 I just heard back from @futurk — they are running mpv **0.37.0** and the issue still persists for characters like `ü` and `ö`. This confirms the bug is not solely an mpv version issue, so I think the fix in the PR is still relevant.

jonath92:
Ok than it is fine for me to merge it @rcalixte

Issue #8042, opened 2025-12-06
Applet Feature Request


Pol853:
### Applet name and maintainer

gpaste-reloaded@feuerfuchs.eu @feuerfuchs

### What would you like to see?

Please add an option to sort history from bottom to top, such that new items appear at the bottom of the list instead of at the top. This would make the applet easier to use with Cinnamon's default layout (where the panel is at the bottom of the screen) by reducing the distance the mouse pointer needs to travel to select the newest item.

hthienloc:
@Antics419 Happy to announce to you that this feature is merged.

PR #8357, opened 2026-02-21
[gpaste-reloaded@feuerfuchs.eu] Add option to reverse history order (newest at bottom)

hthienloc:
**What’s changed**

- added a new checkbox setting “Reverse history order (newest at bottom)” in settings‑schema.json
- bound setting in applet.js and implemented logic to flip the order of displayed entries during normal browsing and when searching
- introduced helper _onSettingsUpdated() to reload the menu on setting changes minor refactoring of applet.js for readability and a couple of style improvements

**Why**

Requested in https://github.com/linuxmint/cinnamon-spices-applets/issues/8042.

### DIFF
--- a/gpaste-reloaded@feuerfuchs.eu/files/gpaste-reloaded@feuerfuchs.eu/6.4/GPasteHistoryItem.js
+++ b/gpaste-reloaded@feuerfuchs.eu/files/gpaste-reloaded@feuerfuchs.eu/6.4/GPasteHistoryItem.js
@@ -105,16 +105,6 @@ GPasteHistoryItem.prototype = {
         }
     },
 
-    /*
-     * Refresh history item's content
-     */
-    refresh: function() {
-            this._applet.client.get_element_at_index(this._index, Lang.bind(this, function(client, result) {
-                let item = client.get_element_at_index_finish(result);
-                this._uuid = item.get_uuid();
-                this.label.set_text(item.get_value().replace(/[\t\n\r]/g, ''));
-            }));
-        },
     
     /*
      * Remove history item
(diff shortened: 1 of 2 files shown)


fredcw:
You don't need comments that say what you've changed e.g. `// refresh will handle highlighting and ordering`

You've swapped spaces for tabs in settings-schema.json

hthienloc:
@fredcw Can you check again?

hthienloc:
@fredcw Can you please take another look?

fredcw:
Sorry, I forgot about this.

There's an issue when the list is reversed and you delete an item, the list doesn't always update properly.

hthienloc:
I've fixed this issue. The `_onClientUpdate` function wasn't handling the `reverseHistory` setting correctly, so I updated it to trigger a full refresh when the list is reversed. I also added a safety check for history item indexing to prevent potential errors.

fredcw:
I measured the timing of function `refresh` in GPasteApplet class, and it's less than 2ms regardless of the value of the parameter `startID`. So I think we may as well remove the parameter `startID` and always refresh the whole list as this would simplify the program somewhat. `startID` can be assumed to be 0 and all calls to `refresh` can omit a parameter.

`_onClientUpdate` would then be just:
```
_onClientUpdate(client, action, target, position) {
        debugLog("[" + uuid + "] Client event: _onClientUpdate");

        this.refresh();
}
```

And with `this._historyItems[position].refresh();` removed from `_onClientUpdate`, there's no more calls to `refresh` function in `GPasteHistoryItem` class so that can be deleted as well.

fredcw:
Nice.

@rcalixte @claudiux LGTM

hthienloc:
My Mint got GNOMEed in the process so I couldn't test it properly.

fredcw:
>My Mint got GNOMEed

Unlucky!

But the applet is OK, I tested it. It works fine.

PR #8378, opened 2026-02-26
[radio@driglu4it] add feature to play last station with middle click

### LINKED ISSUE #8376: [radio@driglu4it] Remember last played station

radio@driglu4it @jonath92

Would it be possible to remember the station played last, so that when the `play` multimedia key is pressed, this station starts to play?

Also the applet's menu could then grow a `play` button, which starts playing this station. And a middle click on the applet would also start this station.

This avoids the need to open the menu, finding the radio station and clicking on it.

This radio station could be marked with the `stopped` icon similar to the `paused` icon displayed, when the playing of a station is paused …

What do you think?

jonath92:
Closes #8376

### DIFF
--- a/radio@driglu4it/files/radio@driglu4it/6.4/radio-applet.js
+++ b/radio@driglu4it/files/radio@driglu4it/6.4/radio-applet.js
@@ -4404,38 +4404,48 @@ function addDownloadingSongsChangeListener(callback) {
 
 
 
+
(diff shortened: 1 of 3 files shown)


jonath92:
I don't want to stress, but I'd appreciate it if the change could be merged. @claudiux can you do this?

jonath92:
thanks :-)

PR #8382, opened 2026-02-26
New HDD-LED applet to indicate drive r/w activity

tucsonst:
This is another HDD/SDD LED emulator based on the nice work by storage-act-led@mrbartblog. This applet is derived from storage-act-led@mrbartblog with some changes in how diskstats is parsed, look of the "LED" icons, and addition of a tooltip with r/w totals and a pop-up dialog with r/w counts for each disk in system.
No longer dependent on emoji fonts but does depend on zenity for the pop-up dialog.

### DIFF
--- a/orloj@faragofr/files/orloj@faragofr/applet.js
+++ b/orloj@faragofr/files/orloj@faragofr/applet.js
@@ -201,7 +201,10 @@ class OrlojApplet extends Applet.TextApplet {
         if (!this._state) return false;
         const [sx, sy] = event.get_coords();
         const [ax, ay] = actor.get_transformed_position();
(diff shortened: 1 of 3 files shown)


tucsonst:
Second commit is minor cleanup.

tucsonst:
Updated to async file operations.

tucsonst:
The Best Practices Scanner found 1 potential issue with shell_string_spawn. This can be disregarded since the shell command is generated internal to the script and does not incorporate user-supplied strings.

tucsonst:
Eliminated the external call to zenity, resolving the shell_string_spawn finding completely.

PR #8398, opened 2026-03-03
wireguard@nicoulaj.net : Show all lines of output in error popup

### LINKED ISSUE #8397: On error, the popup only shows the first line of log, hiding the actual error
### Applet version/Build date

1.0.3

### Cinnamon version

6.6.7

### Distribution

Mint 22.3

_No response_

wireguard@nicoulaj.net @nicoulaj

### What happened?

On error, the Wireguard applet only shows the 1st line of e

### Other information

On error, the popup only shows the first line of log, hiding the actual error 

Image

Actual complete log of the wg-quick command (with the server URL redacted) : 

# wg-quick up wg_perrier
[#] ip link add wg_perrier type wireguard
[#] wg setconf wg_perrier /dev/fd/63
Name or service not known: `XXXX.XXXX.fr:51820'
Configuration parsing error
[#] ip link delete dev wg_perrier

pilatomic:
This fixes #8397 by collecting all lines of log from wg-quick. 

Before : 

image

After : 

image

@nicoulaj : Can you have a look ?

### DIFF
--- a/wireguard@nicoulaj.net/files/wireguard@nicoulaj.net/applet.js
+++ b/wireguard@nicoulaj.net/files/wireguard@nicoulaj.net/applet.js
@@ -124,16 +124,13 @@ const WireGuardApplet = class WireGuardApplet extends Applet.IconApplet {
                 Gio.SubprocessFlags.STDOUT_PIPE | Gio.SubprocessFlags.STDERR_MERGE | GLib.SpawnFlags.SEARCH_PATH
             );
 
-            let out, self = this;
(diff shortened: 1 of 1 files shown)


pilatomic:
I don't think @nicoulaj is around anymore. What's the plan in that case ?

rcalixte:
> I don't think @nicoulaj is around anymore. What's the plan in that case ?

This is a small patch that fixes an actual bug so I've gone ahead and pushed it through. Long-term, if this is now abandoned, we may look for a new maintainer or mark it as orphaned.

PR #8432, opened 2026-03-15
xampp-panel@backids99: Add MySQL shortcuts

GabrielCamara3526:
- Add MySQL Server actions
- Add new applet structure of code
- Add switch option to settings schema
- Remove last edited metadata key


GabrielCamara3526:
I needed a way to access XAMPP's MySQL actions through the applet, so i thought of adding them along this applet's new decoration body. It took me some time to understand the warning above, but after some tinkering i found and applied this new fix that is working for me

Issue #8386, opened 2026-02-28
Thumbnails persistence


gbit00:
### Applet version/Build date

2.4.6

### Cinnamon version

6.6.7

### Distribution

Mint 22.3 - Cinnamon 64-bit

### Applet name and maintainer

CassiaWindowList@klangman @klangman

### What happened?

 Thanks again, this applet is really the most useful and crucial window-sanity tool there is.
One BIG issue though: On the newest version of cinnamon/mint/etc, recently a bug has started where window thumbnails are very often not closing. Moving the mouse along the window list might populate 7 or 8 window thumnails, all over-lapping and difficult to get to disappear, except by clicking the thumbnail window, which always closes the thumnail but obviouslly switches to that window. I tried a million different settings, themes, and even reset all cinnamon settings, yet the issue persists, even with just the text thumbnail.  I also tried other window lists and confirmed thumbnail are working normally on other applets.  Right now, the only solution I can use is to disable thumbnails completely.
Any ideas?

### Other information

<img width="2558" height="1440" alt="Image" src="https://github.com/user-attachments/assets/bc26be98-9271-4a01-ba31-8d888314821c" />

klangman:
I think the latest fix for the Cassia issues under Cinnamon 6.6 has caused this one to appear. I'll be working on it.

nynjapirate:
> I think the latest fix for the Cassia issues under Cinnamon 6.6 has caused this one to appear. I'll be working on it.

One possibly helpful bit of info: I discovered that when cinnamon-settings efffects are disabled the issue goes away.  After re-enabling cinnamon-settings effects, even with all the options disabled, the problem re-appears.  That's all I got but if I discover anything else I will post it here.  Thanks for looking in to this!

klangman:
I have a a fix in the works.

klangman:
I pushed the fix to my Cassia [repo](https://github.com/klangman/CassiaWindowList). I will push the changes to Cinnamon Spices in a few days I hope.

nynjapirate:
> I pushed the fix to my Cassia [repo](https://github.com/klangman/CassiaWindowList). I will push the changes to Cinnamon Spices in a few days I hope.

So far, so good! Works like a charm!

klangman:
> So far, so good! Works like a charm!

Thanks for helping with the testing. Let me know if you spot any issues so I can fix it before pushing this on all users.

PR #8448, opened 2026-03-22
[CassiaWindowList@klangman] V2.4.7 Fix Thumbnail menu issues

klangman:
- Simplify the Thumbnail open/close logic to fix several issues, primarily the issue where Thumbnail menus would not go away, leaving more than one open Thumbnail menu
- Added an option to remove the hotkey combos from the tooltip text for window list buttons
- Improved the animation for removing a window list button
- Removed more deprecated Tweener uses and replaced them with ease calls (only two Tweeners left now)
- Moved to common code for Cinnamon 5.4+ and will stop adding new features to the Cinnamon 4 version after this update


klangman:
Fixes #8386

rcalixte (approved):

PR #8474, opened 2026-03-28
Direct@claudiux: add show/hide session, make compatible with test-spice script

timfeierabend:
This PR adds a new applet setting, `showSession`, that controls whether the "Logout" and "Quit" actions are displayed.

In addition, the code was refactored to make the applet compatible with the **test-spice** dev script. Previously, you'd have to manually change the UUID in the code to add `devtest-`. Now, the main function sets the `MODULE_UUID` value and translation domain binding before the applet is created.

@claudiux

### DIFF
--- a/Direct@claudiux/files/Direct@claudiux/applet.js
+++ b/Direct@claudiux/files/Direct@claudiux/applet.js
@@ -281,7 +281,7 @@ class DirectApplet extends Applet.TextIconApplet {
             this.favAppsSection = new PopupMenu.PopupMenuSection();
 
             //initiate settings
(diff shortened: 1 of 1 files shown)


claudiux:
@timfeierabend 
Thanks for your PR! Good job!

Please replace all `this.metadata.uuid` with `MODULE_UUID` and replace the line
`Gettext.bindtextdomain(MODULE_UUID, HOME_DIR + "/.local/share/locale");`
with
`Gettext.bindtextdomain(MODULE_UUID, GLib.get_user_data_dir() + "/locale");`

timfeierabend:
@claudiux 
Thanks for the feedback! I've made the changes you've requested.

Issue #8473, opened 2026-03-28
Bug: yt-dlp download fails with metadata containing '&' and '~' characters (Radio++ applet)


DJRoby19:
### Applet version/Build date

Versione 2.2.4 (2026-03-27 09:08:33)

### Cinnamon version

Cinnamon 6.6.7

### Distribution

Mint 22.3

### Graphics hardware *and* driver used

Advanced Micro Devices, Inc. [AMD/ATI] Navi 32 [Radeon RX 7700 XT / 7800 XT]

### Applet name and maintainer

radio@driglu4it @jonath92

### What happened?

**Applet UUID:** radio++@claudiux

**Description of the issue:**
When using the yt-dlp plugin to download a song from certain radio stations (e.g., "Radio Monte Carlo Vip Lounge"), the download fails with the following error in the LookingGlass log:

Gjs-Message: JS LOG: [LookingGlass/error] Failed to copy download from tmp dir. The following error occurred: Error: filename couldn't be determined. This seems to be a problem with the used Youtube Download Cli tool.

The issue occurs only with stations that provide "rich" metadata, where the applet constructs a search query by concatenating multiple fields (title, artist, year, timestamp, station name, track ID, etc.) using the tilde (`~`) character as a separator.

**Steps to reproduce:**
1. Tune to "Radio Monte Carlo Vip Lounge" (or any station that provides metadata with multiple fields separated by `~`).
2. Try to download a song where the artist name contains an ampersand (`&`) (e.g., "Milk & Sugar").
3. The download fails with the error mentioned above.

**Example of the generated command:**

yt-dlp --output "/tmp/.8P3QM3/%(title)s.%(ext)s" --extract-audio --audio-format mp3 ytsearch1:"That Body [RMC Edit]~Milk & Sugar~~2019~DEM931900002~234~2026-03-28T09:23:29~2026-03-28T09:24:30~Radio Monte Carlo Vip Lounge~61.21~f36f7129-c8bf-4740-85aa-9a694ac9df5f" --add-metadata --embed-thumbnail

**Root cause analysis:**
1. The ampersand (`&`) in the artist name is interpreted by the shell as a command separator, breaking the `yt-dlp` command.
2. The tildes (`~`) are ambiguous: they may be expanded by the shell or cause an overly long and non-existent search string.
3. The search query includes too many metadata fields (timestamp, ID, etc.) instead of being limited to artist and title, making it impossible for yt-dlp to find a match on YouTube.

[...]

PR #8475, opened 2026-03-28
radio@driglu4it: improve error message when file name can't be extracted

jonath92:
This fixes https://github.com/linuxmint/cinnamon-spices-applets/issues/8473


jonath92:
I would be happy to fix the issues by the Best-practise-scanner but I honestly don't know how to do this in a non hacky way? There is a be a promisify helper in gjs: https://gjs.guide/guides/gjs/asynchronous-programming.html#promisify-helper that might help but it seems to be not yet in cjs? The issue was already in the code before so this pr doesn't make it worse.

PR #8482, opened 2026-03-31
bing-daily@keithdriscoll.nyc: Initial submission

keithdriscoll:
sets Bing image of the day as wallpaper, privacy-first via pea.pix, supports Cinnamon 5.x and 6.x

### DIFF
--- a/bing-daily@keithdriscoll.nyc/files/bing-daily@keithdriscoll.nyc/applet.js
+++ b/bing-daily@keithdriscoll.nyc/files/bing-daily@keithdriscoll.nyc/applet.js
@@ -419,6 +419,17 @@ class BingWallpaperApplet extends Applet.IconApplet {
         if (this.settings) {
             this.settings.finalize();
         }
+
+        // Clean up systemd unit files left on disk
+        let systemdDir = GLib.get_user_config_dir() + '/systemd/user/';
+        for (let name of ['bing-daily.service', 'bing-daily.timer']) {
+            try {
+                Gio.File.new_for_path(systemdDir + name).delete(null);
+            } catch (e) {
+                // File may not exist — not an error
+            }
+        }
+        Util.spawn(['systemctl', '--user', 'daemon-reload']);
     }
 }
 


rcalixte:
This already exists? https://cinnamon-spices.linuxmint.com/applets/view/320

keithdriscoll:
Thanks for the quick reply! They do solve the same problem, but the approach is different. Mine uses a Python subprocess instead of Soup (which avoids the Soup2/3 compatibility headache), routes through Peapix instead of bing.com (I build privacy first apps), and adds history navigation with selectable cache size, a systemd timer, region selection including global, and a settings panel. None of those exist in the other applet.

Happy to walk through the differences in more detail if that would help.

fredcw:
How is this a privacy issue? Microsoft can log your IP address but then so can every site you visit on the internet including peapix.com.

Edit: What's the advantage of a systemd timer, it seems like overkill to me. The `GLib.get_user_config_dir() + '/systemd/user/bing-daily.timer';` file is left behind after the applet is removed as well right?

keithdriscoll:
Good points. let me address each one...

On privacy:

You're right that IP logging happens everywhere. The distinction here is that the Bing endpoint (/HPImageArchive.aspx) is part of a Microsoft‑controlled API surface that includes telemetry, region personalization, and fingerprinting parameters. Microsoft’s own privacy policy states that API requests can be used for telemetry and service improvement.

My goal is simply data minimization: if there’s a stateless mirror (Peapix) with no cookies, no telemetry, and no personalization, I prefer routing through that instead of sending unnecessary requests to Microsoft.

On systemd:

The alternative inside an applet is a GLib polling loop firing every few minutes indefinitely, even when nothing needs to happen. Systemd gives deterministic scheduling, survives Cinnamon restarts, handles missed runs via Persistent=true, and is the standard Linux mechanism for background timers.

It’s not required, but it’s the more reliable tool for a once‑per‑day task.

On leftover files:

Good catch! you're right. I’ll add cleanup for the .service and .timer files in on_applet_removed_from_panel so nothing is left behind.

PR #8484, opened 2026-04-01
SpiceSpy v5.8.0: Improved tooltip displaying schedules

claudiux:


bri-tong:
## Code Review: SpiceSpy v5.8.0 — Improved tooltip displaying schedules

### Summary

This PR adds dynamic tooltip information to the SpiceSpy applet, showing either an in-progress timer, a countdown to the next check, or the absolute time of the next scheduled check. It also renames `UUID` to `MODULE_UUID` (set from metadata in `main()`), uses `GLib.get_user_data_dir()` for locale paths, and makes the shell script derive its UUID dynamically from the directory structure. The tooltip feature is a nice UX improvement, but there are a few correctness issues to address.

---

### Blocking Issues

#### 1. Tooltip never updates when `coloredIcon` is disabled

**File:** `applet.js`, `setIconColorAndTooltip()` (line ~382)

```js
setIconColorAndTooltip() {
    if (!this.coloredIcon) {
      this.actor.style = null;
      return  // <-- returns here, skipping ALL tooltip logic below
    }
```

The early `return` when `coloredIcon` is `false` exits the method before any tooltip update runs. Users who disable the colored icon preference will **never see the schedule tooltip** — the main feature of this PR. The tooltip logic should execute regardless of the icon color preference.

**Suggested fix:** Restructure so the icon color block and the tooltip block are independent:
```js
setIconColorAndTooltip() {
    if (!this.coloredIcon) {
      this.actor.style = null;
    } else if (this.commentsJobsList.length > 0 || this.issuesJobsList.length > 0) {
      this.actor.style = `color: ${this.colorWhileRefreshing};`;
    } else {
      this.actor.style = null;
    }
    // tooltip logic continues unconditionally...
    var nbr_seconds = 15 * this.commentsJobsList.length + 5 * this.issuesJobsList.length;
    // ...
}
```

#### 2. Negative countdown when `future_loop_datetime` is in the past

**File:** `applet.js`, line ~404

```js
let diff = Math.trunc(this.future_loop_datetime.difference(now) / 1000000);
```

If the loop is delayed (e.g., system suspended, heavy load), `future_loop_datetime` could be in the past, making `diff` negative. This would produce nonsensical tooltip text like `"-00:-01:-05"`. There should be a guard:
```js
let diff = Math.max(0, Math.trunc(this.future_loop_datetime.difference(now) / 1000000));
```

#### 3. `update_interval_value()` does not update `future_loop_datetime`

**File:** `applet.js`, lines ~438–447

When the user changes the update interval in settings, `update_interval_value()` resets `this.loopId` with the new interval but does **not** update `this.future_loop_datetime`. The tooltip countdown will show the time remaining from the *old* schedule until the next `loop()` call, misleading the user.

**Suggested fix:** Add `this.future_loop_datetime = GLib.DateTime.new_now_local().add_seconds(sec);` inside `update_interval_value()` after setting the new timer.

---

### Non-blocking Suggestions

#### 4. Extract a time-formatting helper (DRY)

The zero-padded `HH:MM:SS` formatting pattern is duplicated **four times** across the file (lines ~391–398, ~404–423, ~729–734). Consider extracting a helper:
```js
function _padTime(n) { return (n < 10 ? "0" : "") + n; }
function _formatHMS(h, m, s) { return _padTime(h) + ":" + _padTime(m) + ":" + _padTime(s); }
```
This would reduce ~40 lines of repetitive formatting to a few calls and eliminate the risk of inconsistencies between the copies.

#### 5. Shell script: unquoted variable expansions

**File:** `scripts/spices-cache-init.sh`

```bash
cd $SCRIPTPATH/..
UUID="$(basename $PWD)"
cd $SCRIPTPATH
```

`$SCRIPTPATH` and `$PWD` should be double-quoted to handle paths with spaces:
```bash
cd "$SCRIPTPATH/.."
UUID="$(basename "$PWD")"
cd "$SCRIPTPATH"
```

#### 6. `const` declarations reference `MODULE_UUID` before `main()` reassigns it

**File:** `applet.js`, lines 31–36

```js
var MODULE_UUID = "SpiceSpy@claudiux"; // Populated by main() function
// ...
const APPLET_DIR = HOME_DIR + "/.local/share/cinnamon/applets/" + MODULE_UUID;
```

These `const` values are evaluated at module load time using the **initial** value of `MODULE_UUID`. The comment says `MODULE_UUID` is "populated by `main()`", but the constants won't reflect any later reassignment. This works today because the hardcoded default matches the metadata UUID, but the code is misleading — if the UUID ever diverges, the constants will silently use the stale value while the rest of the code uses the new one. Worth a clarifying comment at minimum.

---

### Questions / Clarifications

- Is disabling `coloredIcon` a commonly used preference? If so, the tooltip regression in issue #1 above would affect a noticeable number of users.
- Is there an intentional reason `future_loop_datetime` is only set inside the `if (!this.loopId)` block? If the loop runs but `loopId` was already set by another path, the tooltip would show stale data.

PR #8537, opened 2026-04-13
[radio@driglu4it] survive network change

jonath92:
- Fixes: https://github.com/linuxmint/cinnamon-spices-applets/issues/8514
- Update readme to fix that yt-dlp is the default. This is already the case since years but was not updated in the Readme
- Fixes an edge case: In the past when a user had a lot of stations and the scrollbar was at the bottom, clicking on the applet icon wrongly also triggered the menu item which was behind the icon. This is now fixed with: `clip_to_allocation` (thanks to AI - I otherwise would have spent ages to find that out..)


jonath92:
@claudiux can you merge this? Would be much appreciated

claudiux (approved):

PR #8566, opened 2026-04-20
New applet: GraphicalSystemMonitor@batzkass, a graphical system monitor

batzkass:
Hi,

This is the first applet I'm writing, it is based on sysmonitor@orcus. I rewrote most of the code and added some cool features including:
* display resource data on the graph
* show a large resource graph on applet hover
Regarding the major changes (including code changes), I decided to rename the project to be more explicit and to provide a new set of readme/icon/snapshot.

Note that Orcus no longer seem to update it's applet (no answers in open issues, last cinnamon 3.2).

Thanks for reviewing.


fredcw:
This is a great improvement over sysmonitor@orcus. I did find one minor issue: If I add the applet to the bottom panel and the move the bottom panel to the top, the tooltips stop showing:

```
(cinnamon:1149): Cjs-CRITICAL **: 09:51:59.230: JS ERROR: TypeError: panel is null
show@/usr/share/cinnamon/js/ui/tooltips.js:359:75
show@/home/fred/.local/share/cinnamon/applets/GraphicalSystemMonitor@batzkass/6.6/graph.js:544:15
_onShowTimerComplete@/usr/share/cinnamon/js/ui/tooltips.js:133:18
```

One thing I'm wondering about though, is that since this does everything sysmonitor does but with extra stuff, maybe this should be an improvement to sysmonitor rather than a new applet (the "6.6" folder could simply be added to the existing applet) if @OrcusCZ is happy with that.

Also, I haven't tested it yet, but I suspect this may be compatible with earlier versions of cinnamon, not just 6.6.

Edit: Works on Mint 21.2 (cinnamon 5.8) but not earlier due to use of `Array.prototype.at()` so the directory can be `5.8` instead of `6.6`. I also tried mint 19.3 but it didn't like `cr.textExtents() (St.DrawingArea.get_context().textExtents())`

batzkass:
Thanks for your review, and especially for your tests on older versions !
I will fix the bug you found soon, I'm pretty new to cinnamon and even didn't realized that panels can be moved :-D

I noticed that @OrcusCZ was not answering on recent issues related to sysmonitor, and I didn't found any way to contact him directly. I didn't searched hard though, focusing on the development instead ;-)
Maybe will he have a look here as we mentioned him...

batzkass:
> I did find one minor issue: If I add the applet to the bottom panel and the move the bottom panel to the top, the tooltips stop showing

- [x] Fixed in f9806fb1d52d87418fa9ac15c410f3e606268763

Note about @OrcusCZ : I noticed he didn't pushed on github since 2023. He may not see this discussion.

batzkass:
> Edit: Works on Mint 21.2 (cinnamon 5.8) but not earlier

- [x] Fixed in 1a3f6d61a709b0cd0a483d8ee7540288e80d10e3.

PR #8608, opened 2026-04-29
app-launcher@mchilli: fix submenu clipped when list fills screen height

### LINKED ISSUE #8597: My Launcher list is as tall as my display. When I click on a right-arrow, only the top row of pixels of the top entry in the sublist is displayed.
### Applet version/Build date

Updated on 2026-01-23 07:43:24

### Cinnamon version

6.6.7

### Distribution

Linux Mint 22.3 - Cinnamon 64-bit

Intel GeminiLake [UHD Graphics 605] vendor: Acer Incorporated ALI driver: i915

app-launcher@mchilli @mchilli

### What happened?

My Launcher list is as tall as my display. When I click on a right-arrow, only the top row of pixels of the top entry in the sublist is displayed.
This is because there is no scrolling function. It could be solved by having the right-arrow open the sublist to the right of the main list, instead of inline with it.
See attached screenshot. I clicked on File Managers at the top.

### Other information

Image

manavdhamecha77:
Fixes #8597

## What happened
When the launcher list is as tall as the display, clicking a group's 
right-arrow only shows the top row of pixels of the submenu because 
it has no room to expand and there is no scroll container.

## What this fix does
Wraps the menu box in a `St.ScrollView` and caps the menu height at 
90% of the monitor height, so the list becomes scrollable and submenus 
are always reachable.

## Changes
- `app-launcher@mchilli/files/app-launcher@mchilli/applet.js`

## Testing
Unable to test locally. Requesting maintainer @mchilli to verify.

### DIFF
--- a/app-launcher@mchilli/files/app-launcher@mchilli/6.6/applet.js
+++ b/app-launcher@mchilli/files/app-launcher@mchilli/6.6/applet.js
@@ -652,6 +652,23 @@ class MyPopupMenu extends Applet.AppletPopupMenu {
             this.pointerY = 0;
             this.actorPlaced = false;
 
+            // Remove the menu box to wrap it in a ScrollView
+            this.actor.remove_actor(this.box);
+
+            // Create a scrollable container for the menu
+            this._scrollView = new St.ScrollView({ hscrollbar_policy: 2, vscrollbar_policy: 1 });
+
+            // Add the box to the scrollview and add the scrollview to the actor
+            this._scrollView.add_actor(this.box);
+            this.actor.add_actor(this._scrollView);
+
+            // Limit menu height to 90% of screen to prevent it from going off-screen
+            this._signals.connect(this.actor, 'notify::visible', () => {
+                let monitor = Main.layoutManager.primaryMonitor;
+                let maxHeight = Math.round(monitor.height * 0.9);
+                this._scrollView.set_style(`max-height: ${maxHeight}px;`);
+            });
+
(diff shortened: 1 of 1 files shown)


app-launcher@mchilli/files/app-launcher@mchilli/6.6/applet.js
```diff
@@ -652,6 +652,23 @@ class MyPopupMenu extends Applet.AppletPopupMenu {
             this.pointerY = 0;
             this.actorPlaced = false;
 
+            // Remove the menu box to wrap it in a ScrollView
+            this.actor.remove_actor(this.box);
+
+            // Create a scrollable container for the menu
+            this._scrollView = new St.ScrollView({ hscrollbar_policy: 2, vscrollbar_policy: 1 });
```
rcalixte: The test suggested to use the Cinnamon-based enums. Hard-coding the values is not suggested.

```suggestion
            this._scrollView = new St.ScrollView({ hscrollbar_policy: St.PolicyType.NEVER, vscrollbar_policy: St.PolicyType.AUTOMATIC });
```

rcalixte (changes requested):

mchilli:
After your changes, the menu wouldn't work anymore.
`this.box` is located in `this._boxWrapper`.
Also, it looks weird when the scrollbar is always visible. You could check the menu height every time a submenu is opened and then show the scrollbar if necessary. 
I also noticed that you can't scroll when the mouse is hovering over an open submenu.
And why do you have to check the monitor height every time you open the menu?

Can’t you test the code on your system first?

PR #8639, opened 2026-05-03
Update for applet BgRadio@spacy01

spacy01:
Applet description update.
Pot and bg.po files update.
One tv channel has been replaced.


rcalixte:
Please don't manually edit the translation files. There are scripts for a reason and there will eventually be automated execution of those scripts.

PR #8606, opened 2026-04-27
Update the progress bar url that tracks the evolution in the translation tables

h3nr1ke:
replace the current url the generate the progress bar that is offline

replacing by the service https://progressbar.info/

PR #8651, opened 2026-05-05
Use locally generated progress SVGs for translation tables

fredcw:
Replaces https://github.com/linuxmint/cinnamon-spices-applets/pull/8606

I can do the same for desklets and extensions if this is OK.


h3nr1ke:
I liked it :) thanks for the fix :)

h3nr1ke:
I have opened other PRs in the other projects, maybe we can close all since this options looks more aligned

Issue #8693, opened 2026-05-20
[kdecapplet@joejoetv] Broken SMS, photos & sharing features (setproctitle) and broken links (mount path and KDE Connect Config)


oleole39:
### Applet version/Build date

Version 2.1.1 (2025-10-26 18:03:32)

### Cinnamon version

6.4.8

### Distribution

Mint 22.2

### Applet name and maintainer

kdecapplet@joejoetv @JoeJoeTV

### What happened?

Hello,

I am encountering 3 bugs with the applet: 
1. `ModuleNotFoundError: No module named 'setproctitle'` is thrown when using "take photo", "write SMS", "share URL", or "share text" features. The error originates in `py/dialogs.py`, line 4.
   - `python3-setproctitle` apt package version 1.3.3-1build2 is installed on the system  
   - In addition, I tried `pip install setproctitle` in a virtualenv within the applet folder, but apparently it is not enough
2. When clicking on "Configure KDE Connect" in the context menu, nothing happens (and nothing related to be found in Looking Glass' logs).
3. Mounting phone partition does not mount the correct path (mounting `/run/user/1000/<device_id_string>`) 
   -  The fix suggested by @RensOliemans seems to work properly: https://cinnamon-spices.linuxmint.com/applets/view/334#:~:text=CommonUtils.openURL(%22file://%22+this.mountPoint+%22/storage/emulated/0%22)

All other features seem to work properly.

### Other information

KDE Connect version: 23.08.5 (installed via apt)

RensOliemans:
Hi oleole39,
I'm not a maintainer but since I looked at the code 2 days ago I thought I'd share my reproductions.
Per bug:
1. I cannot reproduce this, "write SMS", "share URL", "share text", work as expected. "take photo" does not exist in my menu, but perhaps that's a configuration issue on my part - I have never used nor needed this functionality.
This is probably because `$ python3 -c "import setproctitle"` works fine on my machine; the apt package `python3-setproctitle` is installed. I did not manually install this, but various preinstalled (for me) packages depend on it such as `mintinstall`, `mintbackup`, `mintupdate`. I'm running Mint 22.3, not sure if this difference matters.
2. Confirmed.
3. Good to hear that this works. I did not create a pull request for this change because I thought perhaps @JoeJoeTV intentionally mounted root, but perhaps a better solution then would be to create two mount options: mount and mount root, where the second fails with a clear error message if there is no permission (if the phone isn't rooted).

[...]

PR #8712, opened 2026-05-27
kdecapplet@joejoetv: Version 3.0

rcalixte:
* Update for KDE Tools 6.x

cc @JoeJoeTV


JoeJoeTV:
@rcalixte I have some questions about this PR:

Why did you change the major version of the applet? The diff is a bit annoying to read, since you both moved files and changed them, but it seems you only changed the 5 to a 6 for the kcmshell command invocation.

I think this is a bit of a bandaid fix and I'm not sure that tying this to the Cinnamon version is good, since Cinnamon does not directly depend on any Qt version.
This means we don't know what version of KDEConnect the person is running other than e.g. on Linux Mint.

With the other things in #8693 I wanted to rewrite this function anyways to either use the DBus method or to use the `kdeconnect-settings` command, which seems to be packaged with KDEConnect, at least on Debian.
One could add the kcmshell command as a further fallback but then I think there should be a check which command is available (either 5 or 6 or none) and if that still does not work, there should be an error message.

Don't want to sound too harsh and I think it's great that people think this extension I made is important enough to update.

rcalixte:
> Why did you change the major version of the applet?

It's technically a breaking change since it is a major upgrade even though the diff itself is small. I'm not attached to either version preference you might have.

> The diff is a bit annoying to read, since you both moved files and changed them, but it seems you only changed the 5 to a 6 for the kcmshell command invocation.

Agreed. Unfortunately, there wasn't a clean way around this and the default diff in git isn't much help in these scenarios either. This approach keeps the older version working for users who might still need it. Cinnamon 6.4 is the first where every major distro was certainly on Qt 6 and therefore KF6, making it a safe benchmark to implement this change.

> I think this is a bit of a bandaid fix and I'm not sure that tying this to the Cinnamon version is good, since Cinnamon does not directly depend on any Qt version.

Agreed but we both looked at the available API and it's the only one available to us since KF6. Similar to the previous logic, only when it's safe for all distributions should we update to KF7 and the matching Cinnamon version. Maybe this would make a good opportunity to engage with upstream about the APIs that were removed and see about having them made available at some point in the future?

> This means we don't know what version of KDEConnect the person is running other than e.g. on Linux Mint.

Users on newer or faster-moving distributions will be inconvenienced but that is the same inconvenience that exists with the applet in the current state. Without a clean API from upstream, our options for a simpler implementation are unfortunately limited.

> With the other things in #8693 I wanted to rewrite this function anyways to either use the DBus method or to use the `kdeconnect-settings` command, which seems to be packaged with KDEConnect, at least on Debian.
> One could add the kcmshell command as a further fallback but then I think there should be a check which command is available (either 5 or 6 or none) and if that still does not work, there should be an error message.

This was something I'd considered but that's a larger diff with lots of potential headaches for implementation with a result that could very likely still be brittle across all distributions. Maybe we can document the packages needed so that the environment could at least be consistent across distributions? I've waffled back and forth on the approaches we could take here.

> Don't want to sound too harsh and I think it's great that people think this extension I made is important enough to update.

I appreciate the feedback! I made sure to CC you for a reason! It's arguably one of the most useful applets on my setup. I've been running with this minor patch locally for a while but every update would reset it. I isolated a clear point in time when it would be safe to split the 5 vs 6 hard-coded changes. We should also have a few years before Qt 7 is out as well, making this a stopgap where we can have a few years to evaluate something sturdier. Even better if we can engage with upstream to have the previous D-Bus API restored in either the KF6 or KF7 versions of KDE Connect.

rcalixte:
@JoeJoeTV Have you had some time to reflect on this? I would like to rebase and merge this with your approval.

JoeJoeTV:
> @JoeJoeTV Have you had some time to reflect on this? I would like to rebase and merge this with your approval.

I did think about it and I still think that the method should be more general to increase compatibility and to not make the change breaking (e.g. I still have a PC that I use the applet on which does not have kcmshell6).
I have implemented this change how I envisioned it here: https://github.com/JoeJoeTV/cinnamon-spices-applets/commit/ae50ba61b235ea925186890ab07b340f90c77c63

If it would be OK for you I would like to use this approach instead (i.e. use DBus for the versions it is available and for all other cases or it DBus fails, try a few commands, with the first one being the most prevalent one).
This should ensure that most systems having KDEConnect installed should be supported by this.
From what I can see this would also not be a breaking change.

rcalixte:
> I did think about it and I still think that the method should be more general to increase compatibility and to not make the change breaking (e.g. I still have a PC that I use the applet on which does not have kcmshell6).

Can you share more about this setup? I made the versioned subdirectories specifically to address this. I did want to make sure users on older distros/versions of Cinnamon would not be impacted.

> I have implemented this change how I envisioned it here: [JoeJoeTV@ae50ba6](https://github.com/JoeJoeTV/cinnamon-spices-applets/commit/ae50ba61b235ea925186890ab07b340f90c77c63)

Do you think this is ready for a pull request? Have you tested it on the latest Mint? It's your applet so it's ultimately your call. If I felt strongly about a certain direction, forking the applet is always an option but I don't think that kind of fragmentation is something we desire or need for this.

JoeJoeTV:
> Can you share more about this setup? I made the versioned subdirectories specifically to address this. I did want to make sure users on older distros/versions of Cinnamon would not be impacted.

The changes here are basically only affected by the KDE Connect or KF version installed and not really by the Cinnamon version, so I'm not that fond of tying such changes to the cinnamon version, that's why I wanted to have a solution that is independent of the Cinnamon version.

The method I use in the commit I mentioned is to first try using the DBus method if the compatability level (i.e. KDE Connect version) is in a range that could support it.
If that does not work, or if the user is running a different KDE Connect version, then the code tries to run a command to open the Configuration window.
From what I can tell, the `kdeconnect-settings` command is the most likely to be available without any additional packages, so that is tried first.
And if that fails, both the KF5 and KF6 versions of the `kcmshell kcm_kdeconnect` are tried.
Finally, if that did not work, there are no other options as far as I'm aware of, so it will log an error.
I'm thinking if a message box makes sense here, but I'm not sure yet.

Since the method is basically a superset of the previous one (Only try DBus and `kcmshell5 kcm_kdeconnect` this would also not be a breaking change and would still work for people with KF5.

> Do you think this is ready for a pull request? Have you tested it on the latest Mint? It's your applet so it's ultimately your call. If I felt strongly about a certain direction, forking the applet is always an option but I don't think that kind of fragmentation is something we desire or need for this.

I did test it on Mint 22.3 where I'm already using the Applet. Also want to try it on the other PC running an older version, but it should work.

I want to maybe commit some more fixes that people mentioned in the other issue, so I want to review the code again, but else it would be ready for a pull request.

Do you think this is the wrong approach? My goal from the start of the 2.0 rewrite was to have broad compatibility with KDE Connect versions (as can be seen by the compatibility mode implementation), so I though keeping to that would be good. At some point, the really old code for KDE Connect 1.3 and 1.4 could be dropped, but it is not a big hassle at the moment I would say.

PR #8784, opened 2026-06-12
[GraphicalSystemMonitor@batzkass] Improve net up/down readability.

batzkass:
Minor change in network data display. Also change default font settings.


batzkass:
Result :
image
Less chars to unity, separate unities for up/down.

rcalixte (approved):

PR #8813, opened 2026-06-19
Add pacman-updates to Cinnamon applets

Twilight0:
Checks for updates in Pacman package manager database


claudiux:
Please inform users that this applet is intended for Arch-based distributions only, not for Linux Mint.

Twilight0:
Right, I 've just updated README, metadata and made pacman database check an asynchronous call.

claudiux (approved):

PR #8816, opened 2026-06-23
multicore-sys-monitor@ccadeptic23: fix negative Memory Free % and add ZRAM tooltip block

GayratTursunov:
## Summary
Two changes to **Multi-Core System Monitor** (`multicore-sys-monitor@ccadeptic23`), `5.6/` folder only.

@claudiux — would appreciate your review 🙏

### 1. Fix: negative Memory "Free" percentage (Gnome System Monitor method)
When *Memory Calculation Method* = **Gnome System Monitor**, the four memory tooltip
segments (Used / Cached / Buffers / Free) did not partition `MemTotal`: "Used" was
derived from `MemAvailable` (which already excludes the non-reclaimable page cache),
then Buffers and Cached were added again as separate segments. With a large page
cache the segments summed past 100% and **Free was shown as a negative %** (e.g. `-4.10 %`).

Now "Used" is based on `MemFree` (`Used = MemTotal − MemFree − Buffers − Cached`) and
`Free = MemFree / MemTotal`, so the four segments always sum to exactly 100% and Free
can't be negative — consistent with the existing htop method.

### 2. Feature: ZRAM info block in the tooltip
Adds a **ZRAM** section after Swap showing the same data as
`zramctl /dev/zram0 --output-all`, read directly from sysfs (no subprocess):
Algorithm, Data, Compressed, Total, compression Ratio, Mountpoint.

- Sources: `/sys/block/zram0/mm_stat`, `comp_algorithm`, and `/proc/swaps`.
- Auto-hides on machines without `/dev/zram0` (the sysfs read rejects → nothing rendered).
- New toggle **Enable ZRAM Tooltip** (`Swap_enabledZramTooltip`, depends on the Swap tooltip).

Example:
```
----------------- ZRAM -----------------
    Algorithm:        zstd
         Data:      1.6 GB
   Compressed:   389.90 MB
        Total:   403.00 MB
        Ratio:        4.00
   Mountpoint:      [SWAP]
```
(byte units follow the applet's existing binary/decimal size preference, like the Memory/Swap lines)

## Scope & testing
- Only the `5.6/` folder is changed (serves Cinnamon 5.6–6.6+); version bumped to 3.16.6 with a CHANGELOG entry.
- `validate-spice multicore-sys-monitor@ccadeptic23` → no errors; `node --check` passes.
- Tested on Cinnamon 6.6.7 / LMDE 7: Free now reads correctly; ZRAM values match `zramctl`; toggling the switch shows/hides the block.

Originally reported at ccadeptic23/Multi-Core-System-Monitor#24.

### DIFF
--- a/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/5.6/applet.js
+++ b/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/5.6/applet.js
@@ -329,6 +329,7 @@ var MCSM = class MCSM extends Applet.TextIconApplet {
         this.settings.bind("Mem_enabledBufferCacheSharedTooltip", "Mem_enabledBufferCacheSharedTooltip");
         this.settings.bind("Swap_enabled", "Swap_enabled");
         this.settings.bind("Swap_enabledTooltip", "Swap_enabledTooltip");
+        this.settings.bind("Swap_enabledZramTooltip", "Swap_enabledZramTooltip");
         this.settings.bind("Mem_chartType", "Mem_chartType");
         this.settings.bind("Mem_squared", "Mem_squared");
         this.settings.bind("Mem_method", "Mem_method");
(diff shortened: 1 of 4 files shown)


claudiux (approved):

claudiux:
@GayratTursunov 
Many thanks for the care!

PR #8826, opened 2026-06-25
sessionManager@scollins: confirm session-ending actions

dshep79:
## Summary

- route logout, restart, and shutdown through Cinnamon's session manager
- show the standard Cinnamon confirmation dialog before ending the session
- match the confirmation behavior used by Menu Eleven

## Validation

- `node --check sessionManager@scollins/files/sessionManager@scollins/applet.js`
- `./validate-spice sessionManager@scollins`

@collinss Please review this functional change when convenient.


collinss:
No, thank you. I don't want it to give me a confirmation dialog. The current way of doing it is intentional. If you don't like it, you are welcome to fork it and submit your own under a different name, but I'm not going to change this one.

PR #8835, opened 2026-06-27
multicore-sys-monitor@ccadeptic23: tooltip SI/IEC unit toggle, percentages-only option, ZRAM block rework

GayratTursunov:
## multicore-sys-monitor@ccadeptic23

Tooltip improvements for the Memory / Swap / ZRAM sections (version bumps v3.16.7 → v3.16.9). Only this applet's files are touched.

### v3.16.9
- Memory and Swap tooltip amounts are now always shown; the "Show amounts in Bytes" setting selects the **unit base** (ON = SI kB/MB/GB, OFF = IEC KiB/MiB/GiB) instead of hiding them. This fixes the Swap section disappearing when the setting was off, and adds the used percentage to the Memory header to match Swap.
- Buffers/Cache/Shared and ZRAM amounts follow the same unit toggle, so the whole tooltip uses one consistent unit base (OFF = IEC, matching `free -h`).
- The ZRAM title now shows the total RAM used with algorithm and ratio (e.g. `ZRAM 577 MiB (zstd, 3.87x)`); the block lists Disksize (capacity), Data and Compressed.
- New **"Show Memory/Swap byte amounts"** switch (default on) for a percentages-only tooltip; the SI/IEC unit setting is dependent on it.

### v3.16.8
- More compact ZRAM block (algorithm and ratio moved into the title).

### Testing
- Reloaded the applet on Cinnamon (X11): loads cleanly, settings upgrade succeeds, no JS errors.
- Cross-checked tooltip values against `free -h` / `free --si`, `swapon --show`, `cat /proc/meminfo`, and `zramctl`.

### DIFF
--- a/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/5.6/applet.js
+++ b/multicore-sys-monitor@ccadeptic23/files/multicore-sys-monitor@ccadeptic23/5.6/applet.js
@@ -2141,13 +2141,13 @@ var ZramDataProvider = class ZramDataProvider {
         if (!this.isRunning) return "";
         if (!this.available || this.data == null) return "";
 
-        let trans = _("ZRAM");
+        // Fold algorithm + ratio into the section title, e.g. "ZRAM zstd (2.66x)".
+        let trans = _("ZRAM") + " " + this.data.algorithm;
+        if (this.data.ratio > 0)
+            trans += " (" + formatNumber(parseFloat(this.data.ratio.toFixed(2)), 2) + "x)";
         let len = trans.length - 2;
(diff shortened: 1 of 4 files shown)


claudiux:
_Tooltip: Memory and Swap amounts are now always shown_

Users should be able to choose that.

GayratTursunov:
mem_without_amounts
The applet includes an option to display data as percentages only.

claudiux (approved):

claudiux:
@GayratTursunov 
Please can you also show ZSWAP data?
https://chrisdown.name/2026/03/24/zswap-vs-zram-when-to-use-what.html

GayratTursunov:
ZSWAP data:
image

claudiux (approved):

PR #8849, opened 2026-07-02
Add: Windows-style auto-hide tray applet

Puddings22:
Hello, 

This merge  adds a new applet that hides application tray icons behind an expand arrow by default (like the Windows taskbar overflow), showing them in a centered grid popup on click.

It: 
  - Hosts both XApp/StatusNotifier icons and legacy XEmbed tray icons under the same hide/show rules
  - Forwards popup clicks so Qt tray menus (e.g. FortiClient) still work correctly
  - Per-icon visibility toggles via right-click menu
  - Hover tooltips (falls back to app name when the app provides none)
  - Configurable: expand mode (popup/inline), max icon size, auto-collapse delay, collapse-on-click, auto-hide of new icons

  Validated with `./validate-spice collapsible-xapp-status` - no errors.

Tests ran: 
  - Ran validate-spice locally (passes)
  - Manually tested on Cinnamon 6.6 / Linux Mint (X11): expand/collapse, icon toggles, tooltips, legacy tray icon (evolution-alarm-notify), Qt app menu (FortiClient) from popup


Puddings22:
Regarding the global_begin_modal warning: this is copied verbatim from the stock systray@cinnamon.org applet's _onEvent(), which ships with every Cinnamon install it's the mechanism used to forward a button-press into an embedded legacy XEmbed tray icon so its own popup menu can take the input grab correctly. The grab is bracketed tightly around a single synchronous icon.handle_event() call and released immediately after, not held persistently. Since this applet also hosts legacy tray icons (to cover apps like evolution-alarm-notify that don't use the newer StatusNotifier protocol), the same pattern was needed here.

Puddings22:
@claudiux @rcalixte could one of you take a look at this one when you have a moment? It has been open since July with the checks green and no conflicts, and I suspect it slipped past because there are already two "collapsible systray" applets in the repo. Three things that set this one apart, in case that helps the decision:

1. It covers both icon systems under one set of rules: XApp/StatusNotifier icons and legacy XEmbed tray icons (evolution-alarm-notify, older Qt apps). Both existing collapsible applets wrap the legacy systray only, so XApp icons stay uncollapsed with them.

2. It replaces the stock System Tray and XApp Status applets together, with the Windows-style behaviour: every app icon hidden by default (including apps installed later), an arrow that opens a centred grid of the hidden ones, and a right-click list to promote favourites to the panel.

3. Clicks from the popup are forwarded only after the popup releases its grab, so Qt tray menus (FortiClient etc.) open correctly. That is also the reason for the scanner's global.begin_modal warning: it is the same bracketed grab the stock systray@cinnamon.org uses for XEmbed icons, held for one synchronous handle_event() and released immediately see my reply above.

If you would rather see any of it changed, or want it split differently, say the word and I will update the PR. Thanks for maintaining the spices.

PR #8865, opened 2026-07-10
Cinnamon sidebar@beatlink

BeatLink:
Dock a single window to the left or right edge of a monitor at a configurable
width. Other windows are kept out of the reserved area, so a docked window
behaves like a real sidebar: the rest of the screen stays usable and nothing
covers the dock.


BeatLink:
Bump

claudiux (approved):

PR #8866, opened 2026-07-11
ddcci-monitor-control@andr35: improve popup menu layout; fixed sliders

### LINKED ISSUE #7423: Applet Feature Request

ddcci-monitor-control@andr35 @andr35

`Current Contrast` on top of the list, please. Because the `Current Contrast` value takes longer to load it shifts the brightness options above so that one cannot click them instantly consistently. See the video below for clarity.

https://github.com/user-attachments/assets/d01d8de5-dba1-4748-b200-798f527cc508

inucat:
While being loaded, the applet now disables the sliders instead of hiding so that their positions are fixed in the popup menu.

- Removed `show()` and `hide()` as they make the slider positions unpredictable
- Instead, used `reactive` property to avoid the sliders reacting while the applet is being loaded

This Pull Request resolves linuxmint/cinnamon-spices-applets#7423.

@andr35 I'd appreciate your review. This is one of my favorite Cinnamon applets. Thanks in advance!

### DIFF
--- a/ddcci-monitor-control@andr35/files/ddcci-monitor-control@andr35/applet.js
+++ b/ddcci-monitor-control@andr35/files/ddcci-monitor-control@andr35/applet.js
@@ -136,17 +136,17 @@ MyApplet.prototype = {
         this.menuItemContrastValue.set_text(this._get_contrast_label(this.currContrastValue));
 
         if (this.currBrightnessValue === undefined) {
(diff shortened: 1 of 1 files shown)


andr35:
It seems a good solution for me 👍

Thank you

claudiux (approved):

PR #8867, opened 2026-07-12
orloj@faragofr: Add time zone, GeoIP, rise/set and moon phase display options; fix skipped same-day sunset

kosonya:
Hi @faragofr — this PR adds a set of opt-in display options to Orloj, plus one bugfix. Since this comes from a user rather than the author, it awaits your review; happy to adjust anything.

**Backward compatibility was the guiding principle: a user who upgrades and changes nothing sees identical behavior.** Every feature is off by default (or defaults to the original behavior) and is enabled through a new checkbox.

## Commits

1. **Fix same-day sun events being skipped west of Greenwich** (v1.1.2) — `sunriseSunset()` tables events by UTC date, but the scanner started at the current UTC date, so west of Greenwich the remainder of the local day (e.g. tonight's sunset after ~17:00 in California) was skipped and the panel jumped a day ahead. The scan now starts one UTC day earlier; the `> now` filter discards past events. *This is a pure bugfix — say the word if you'd prefer it split into its own PR.*
2. **Time zone selection** (v1.2.0) — a "Use Local Sidereal Time" checkbox (on by default: the dial center keeps its LST readout and everything works exactly as before). Unchecking it switches the center to civil time and reveals automatic (system) or manual time zone selection, applied to all displayed times. The zone list is populated at runtime from the system tz database via `settings.setOptions()` (same pattern as hwmonitor@sylfurd and qredshift@quintao), keeping ~600 zone names out of the schema and the translation template.
3. **Optional GeoIP location** (v1.3.0) — off by default. Queries ip-api.com (the service weather@mockturtl uses) at most hourly; manual coordinates remain the fallback. libsoup 3 is imported lazily so the applet still loads where the typelib is missing.
4. **Panel visibility toggles** (v1.4.0) — hide either body's entry; a ☀☾ placeholder keeps the applet clickable.
5. **Optional both-rise-and-set display** (v1.5.0) — per body, off by default: stacks the body's rise and set in two lines instead of showing only the soonest event. Each body renders in its own column so the ☀/☾ glyphs stay aligned.
6. **Optional moon phase indicator** (v1.6.0) — off by default: a schematic moon disk (same renderer as the dial) with illumination % and phase angle.
7. Screenshot and translation template updates (new strings merged into es/fr/hu as untranslated entries).

## Compatibility & testing

- JS kept compatible with the declared Cinnamon versions (5.4–6.6 / SpiderMonkey 78+): `imports.byteArray` instead of `TextDecoder`, lazy Soup 3 import with graceful fallback, `GLib.TimeZone.new_identifier` guarded.
- All I/O is async; timers, signal sources and the Soup session are cleaned up in `on_applet_removed_from_panel`; nothing writes to the install directory.
- Each commit was validated with `validate-spice` and live-tested individually on Mint 22.3 / Cinnamon 6.6 (x11), including the settings cascades and the GeoIP fetch path.

## Disclosure

Developed with AI assistance (Claude), per the repository README's AI-assisted coding section; the code was self-reviewed against `.github/copilot-instructions.md`, and every commit was reviewed and tested by me on a live panel.


faragofr:
Hi, thank you for your contribution!

 I am away for a few days but will try to check before the end of next week.

faragofr:
Thanks for this, @kosonya, especially for identifying and fixing the "west of Greenwich" bug. 
There are two things on the panel display I'd like to point out:

  1. The ☀☾ placeholder when panel entries are disabled

In `_refresh`:

```js
this.set_applet_label(sunShown || moonShown ? "" : "☀☾");
```

This only accounts for the rise/set columns and ignores `showMoonPhase`, so turning off both rise/set entries while keeping the moon-phase widget on produces a redundant ☀☾ label sitting next to the moon disk and the placeholder's stated purpose (keeping the applet clickable) is already satisfied by the visible widget.

I'd like the fallback to trigger only when nothing at all is shown (no rise/set columns and no moon-phase widget) and in that case to fall back to the applet icon (icon.png) rather than a glyph-pair text label. 

I'll readily grant the current icon is quite minimalistic and deserves a better design at some point, but as the empty-state placeholder it's the right direction.

  2. The moon-phase text is too small to read

```js
this._moonPhaseLabel = new St.Label({ style: "font-size: 8px;", ... });
```

At 8px the NN% MMM° line is very hard to read. Would you be OK to move the illumination/angle figures into the tooltip and keep just the disk in the panel?

Everything else looks good to me — happy to take another pass once these are in.

faragofr:
@kosonya Taking you up on the offer in your description to split the bugfix out: I've opened #8963 with just the west-of-Greenwich fix, which is your commit cherry-picked with your authorship intact, so that correctness fix can ship without waiting.
Thanks again for catching it!

PR #8892, opened 2026-07-20
Add Snippitor applet

fosscharlie:
Adds Snippitor, a system-wide text-expansion applet with a GTK settings editor.

- Panel applet (applet.js) starts/stops a background daemon (snippitor_daemon.py)
  that watches for configured trigger strings (e.g. "tel#") and expands them into
  full text (e.g. "01234567890") anywhere on the desktop.
- A companion GTK editor (snippitor_gui.py) lets the user add/edit/remove/export/
  import trigger -> expansion pairs, with autosave.
- pynput and its dependencies (python-xlib, six) are bundled in vendor/ so no
  separate pip install is required.
- X11 only (documented in the README's Known Limitations section) - this is a
  fundamental constraint of watching keystrokes this way, not a bug.
- Data is stored inside the applet's own folder rather than ~/.config, so removing
  the applet cleanly removes its data too.
- Passes ./validate-spice with no errors.

Licensed under GPL v3 (see LICENSE).


fosscharlie:
Is there something else I need to do?

claudiux (approved):

PR #8899, opened 2026-07-23
Tailscale Manager

inventor96:
This is a relatively simple Cinnamon applet that monitors and manages Tailscale via the CLI. It displays a tray icon indicating Tailscale's status, and exposes some quick options via a context menu.


inventor96:
Updated to:
- Chain async commands instead of using sync commands
- Use argv for commands in all instances; only the `tailscale set --exit-node` command used user input (by extension, from the user-defined node names), but applied it to all instances for consistency
- Changed UI-related logic to only update the UI when there are underlying state differences

PR #8911, opened 2026-07-30
tinycarebot@ap0r: Initial upload.

ap0r:
Uploaded tinycarebot (the applet that I made, see https://github.com/ap0r/Tinycarebot for detailed info) so that it can be downloaded directly from within Mint.


ap0r:
Let's hope this is good now, making it work with load_async() was surprisingly difficult. Probably skill issue on my end.

ap0r:
Please, whoever reads this, provide some clear feedback as to why this is not accepted, if any issues remain please explain them, I am new to contributing to open source and the experience has been superbly frustrating so far.

PR #8914, opened 2026-07-31
hwmonitor@sylfurd v1.4.0: Add optional GPU utilization and GPU memory graphs (NVIDIA + AMD)

wiebelhaus:
## What

Adds two new optional graphs to the Graphical hardware monitor: **GPU** (utilization %) and **GMEM** (GPU memory usage). Both are **disabled by default** and are enabled from a new "GPU" tab in the applet settings, with the same options as the existing graphs (size, custom label, detail label).

## How

A shared poller auto-detects the GPU vendor once, then feeds both graphs:

- **AMD** (amdgpu driver): read directly from sysfs (`gpu_busy_percent`, `mem_info_vram_used`, `mem_info_vram_total`). Instant reads, no subprocess.
- **NVIDIA**: `nvidia-smi --query-gpu=...` spawned **asynchronously** via Gio.Subprocess, so the Cinnamon UI thread never blocks. One subprocess per refresh tick regardless of how many GPU graphs are enabled; `getData()` returns the previous sample.
- **No supported GPU**: providers report "n/a" and return 0; since the graphs are off by default, machines without a GPU are unaffected.

Intel GPUs are not supported: the i915/xe drivers do not expose a utilization counter readable without elevated permissions (`intel_gpu_top` needs perf access), and iGPU memory is shared system RAM. This is documented in the README.

## Changes

- `3.8/providers.js`: shared `GpuPoller` + `GpuUtilDataProvider` / `GpuMemDataProvider`
- `3.8/applet.js`: settings binds, graph creation, theme object entries
- `3.8/graph.js`: custom label cases for the two new graph types
- `3.8/settings-schema.json`: new GPU page with two sections (all defaults off)
- `README.md`: GPU support and requirements section
- `metadata.json`: version 1.3.4 -> 1.4.0

## Testing

Tested on Linux Mint 22.3 (Cinnamon 6.6.5) with an NVIDIA GTX 1650 (nvidia-smi backend): graphs track utilization and VRAM correctly at 1s refresh, detail labels show % and formatted bytes, settings tab works, no errors in ~/.xsession-errors, and disabling the graphs removes them cleanly. The AMD path follows the documented amdgpu sysfs interface (same fields used by tools like radeontop); I do not have an AMD card in this machine, so review from an amdgpu user would be welcome.

https://claude.ai/code/session_01UUKno4V7XVo8nv95eJghdj

### DIFF
--- a/hwmonitor@sylfurd/README.md
+++ b/hwmonitor@sylfurd/README.md
@@ -94,3 +94,12 @@ Vertical pane examples:
 *  andreevlex
 *  clefebvre
 *  jacobwills
+## GPU graphs
+
+Since v1.4.0 the applet can optionally show GPU utilization and GPU memory graphs. They are disabled by default; enable them in the applet settings under "GPU".
+
+Supported hardware:
+- NVIDIA: requires the `nvidia-smi` tool (included with the NVIDIA driver)
+- AMD: read directly from sysfs (`amdgpu` driver), no extra tools needed
+
+Intel GPUs are currently not supported, because the i915/xe drivers do not expose a utilization counter that can be read without elevated permissions.
--- a/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.8/applet.js
+++ b/hwmonitor@sylfurd/files/hwmonitor@sylfurd/3.8/applet.js
@@ -162,6 +162,18 @@ GraphicalHWMonitorApplet.prototype = {
         this.settings.bind("diskwrite_use_custom_label", "diskwrite_use_custom_label", this.settingsChanged);
         this.settings.bind("diskwrite_custom_label", "diskwrite_custom_label", this.settingsChanged);
         this.settings.bind("diskwrite_show_detail_label", "diskwrite_show_detail_label", this.settingsChanged);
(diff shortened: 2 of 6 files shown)


wiebelhaus:
Regarding the two scanner warnings (`sync_file_get_contents` and `sync_file_test` in the new GPU code): both call sites operate exclusively on **sysfs** paths (`/sys/class/drm/cardN/device/...`), which are kernel-memory-backed virtual files - reads complete in microseconds and never touch disk or network, so they cannot meaningfully block the main loop. The `file_test` scan additionally runs only once per session (vendor detection is cached).

This is also consistent with the applet's existing code: `BatteryProvider` uses the same `GLib.file_get_contents`-on-sysfs pattern for `/sys/class/power_supply/BAT0/capacity`.

Where blocking is a real risk - the NVIDIA path, which spawns a subprocess - the implementation is fully async (`Gio.Subprocess.communicate_utf8_async`), with the provider returning the previous sample so the UI thread never waits.

Happy to convert the sysfs reads to `Gio.File.load_contents_async` if the maintainers prefer strict compliance with the scanner.

PR #8915, opened 2026-07-31
internet-speed-meter@mebs: Add new applet

MDHasan0078:
A lightweight Cinnamon applet that displays real-time internet upload and download speed in the panel.

## Features
- Real-time download/upload speed (auto-scaling units: B/s, KB/s, MB/s, GB/s or bit variants)
- Automatic interface detection via the default route (no external commands spawned)
- Compact single-line or stacked layout
- Configurable refresh interval, decimal places, and font size

## Implementation notes
- Reads `rx_bytes`/`tx_bytes` from `/sys/class/net/<iface>/statistics/`
- Uses only `Mainloop.timeout_add` + file reads per tick (no subprocesses, no watchdog warnings)
- GPL-3.0

Screenshot


MDHasan0078:
Fixed the two best-practices scanner findings. Replaced both synchronous `GLib.file_get_contents()` calls with async `Gio.File.load_contents_async()`:

- **Interface detection** now reads `/proc/net/route` asynchronously (with a `_detecting` guard so retries don't stack).
- **Speed sampling** now chains async reads per tick (`rx_bytes` \u2192 `tx_bytes` \u2192 compute) with a `_sampling` guard to prevent overlapping samples.

Verified live after reload: label updates correctly (`\u2193 49.0 KB/s \u2191 6.3 KB/s` \u2192 `\u2193 2.4 KB/s \u2191 2.4 KB/s`), no new warnings in `~/.xsession-errors`.

MDHasan0078:
Hi @claudiux @rcalixte sorry for the disturbance. 


It has been a long time since I submitted this PR. Could I have a small amount of your time to review it?

Happy to make any changes if needed. Thanks!

rcalixte:
> It has been a long time since I submitted this PR. Could I have a small amount of your time to review it?

There are several of these that already exist. Does this add anything that the others lack?

The README also states that Python is a dependency but this applet doesn't currently contain any Python?

MDHasan0078:
> > It has been a long time since I submitted this PR. Could I have a small amount of your time to review it?
> 
> There are several of these that already exist. Does this add anything that the others lack?
> 
> The README also states that Python is a dependency but this applet doesn't currently contain any Python?

Hi @rcalixte @claudiux ,

Thanks for looking, and good catch on the README — you're right, the applet is pure JS/GJS, no Python. Just fixed it (commit 1a02ec86d).

And fair question on "what does this add." Honestly, a few small things:

- No subprocesses — the existing speed applets spawn processes (cardsurf does a sync spawn, netspeed@iMayhem pipes through Gio.Subprocess). Mine just reads two sysfs files directly, nothing else running.
- Auto interface detection — picks the default-route interface itself, no hardcoding, no shelling out to ip.
- Zero deps — pure GJS, one small label on the panel.

i have a toaster laptop (acer aspire a315-33 4gb ddr3L ram) and every extra resources taken from me is like curse for my laptop. So felt some other people also fell the same as me. This applet does not solve a major problem but for the user small resources pile up to bloat. if every applet demand extra resources the os will eventually become bloated and heavy in future like windows.

PR #8941, opened 2026-08-07
tinycarebot@ap0r: Mental and physical health bot

ap0r:
Tinycarebot is an applet I made that sits on a panel of choice and displays a random message every now and then, or when clicked.

The messages can be user-edited, but the default messages revolve around the themes of mental health, physical health, ergonomics, cleanliness, and order. Some examples include "Drink water 💧", "Look outside 👀", "Get up and stretch 🧘", "Rest your eyes 👁️👁️", "Respond to a friend 🗣️", "Look at some nature 🪴", "Trouble? Ask for help 👥", "Take your medication 💊", and so on.

This applet was originally hosted at https://github.com/ap0r/Tinycarebot


ap0r:
Can a mantainer please take a look at this? @claudiux if something is wrong please tell me what is incorrect and I will fix it.

ap0r:
Can a mantainer please take a look at this? @claudiux or @rcalixte (I saw you have merged some PRs) if something is wrong please tell me what is incorrect and I will fix it. All automated checks pass or are auto-skipped.

I understand you guys may be swamped with AI slop but in that case just leave a note in contributing.md "don't contribute right now, we are too busy"

mtwebster:
Hi, a few issues here (in addition to the failed Spice validation, see above):
- During updates, the install folder is deleted. Any edits made there won't survive. Try to use applet settings instead.
- You may want to consider setting up translations for these messages.

ap0r:
Thank you for your considerated reply! I will get to the issues you mention! Thanks for explaining that the install folder is deleted on updates, I did not see that in the documentation so I will see how I can make the phrases reside in the applet settings instead. Regarding translations, I do not know how to do that, I will research it. Lastly is the issue of non-square icon the only issue with the validation, i.e. the script prints all errors or can there be other issues hidden?

mtwebster:
That script only checks for technical issues related to adding this spice to the server and publishing it. It's not actually checking your applet to see if it works.

Just about every other applet here implements localization - plenty of examples.

Writing directly to the settings folder isn't recommended either - I meant you should use the settings api. There's no guarantee we won't invalidate that folder somehow in the future - your only guarantee is using the api itself. You might consider making your list a dictionary with unique keys - you could then manage their visibility via the api while still keeping the list as read-only. A user could add/modify/remove, populating a settings list that works with these keys.

Functionally, I'm not sure about your 'ticker' when it encounters the emoji - it may end up printing an invalid code while scrolling due to encoding issues. You'd probably be better off having 2 separate labels, one for text, one for emoji.

PR #8947, opened 2026-08-09
kdecapplet@joejoetv: fix browsing single storage mounts

lucasleocs:
Addresses the mount-path problem described in #8693.

When KDE Connect reports exactly one remote storage directory through
`getDirectories()`, the applet now opens that directory instead of
always opening the SFTP mount root.

On the affected device, the mount root exists but returns "Permission denied"
when listed, while the directory reported by KDE Connect is accessible normally.

When multiple directories are reported, the existing mount-root
behavior is preserved. The same fallback is also used if
`getDirectories()` is unavailable or fails.

Tested with:
- Linux Mint
- Cinnamon 6.6.9
- KDE Connect 23.08.5
- CJS 115.1

Tested scenarios:
- mount and unmount
- repeated remounting
- Cinnamon reload
- device disconnect and reconnect
- remount after force-stopping KDE Connect on Android

@JoeJoeTV

### DIFF
--- a/kdecapplet@joejoetv/files/kdecapplet@joejoetv/js/modules.js
+++ b/kdecapplet@joejoetv/files/kdecapplet@joejoetv/js/modules.js
@@ -1343,12 +1343,29 @@ class SFTPModule extends KDECModule {
     }
 
     startBrowsing() {
-        // NOTE: There is also a DBus method to open the file manager at the mount point, but that is currently broken, at least for me
+    if (this.mounted == true) {
+        let path = this.mountPoint;
 
-        if (this.mounted == true) {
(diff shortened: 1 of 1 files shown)


JoeJoeTV:
Thanks, but i'm already woking on a more elaborate fix that also lets the user select which directory to open (or the first, if set in settings).
I think that would be more helpful.
Just have to get the last exam done tomorrow.

lucasleocs:
Thanks for the update! That sounds better, nice work.
Good luck with your exam!

JoeJoeTV:
@lucasleocs btw, this is in the newest update now.

PR #8950, opened 2026-08-10
workspace-switcher-plus@tiagomiguelcs: Add new Applet

tiagomiguelcs:
Workspace Switcher Plus is a Cinnamon panel applet that lets you switch between workspaces directly from the panel. It extends the original Workspace Switcher applet with additional appearance controls and interaction options.


rcalixte:
Hi, you only need to open and focus on a single pull request for a unit of work. Please stop closing and opening new pull requests for the same thing. If you have further edits, push them to the same open pull request.

PR #8952, opened 2026-08-12
focal: Initial release

zoharsnir:
A simple panel text label you either set yourself or point at a system (EDS) calendar to show whatever event you're currently in. Meant to be a lightweight, low-friction "what should I be focusing on right now" indicator.

- Two modes: type text directly, or drive it from a personal calendar (via Evolution Data Server)
- Configurable colors, optional icon, keyboard shortcut to edit
- No network calls, no telemetry - only local EDS access in Calendar mode

Passes validate-spice with no errors.

Upstream: https://github.com/zoharsnir/cinnamon-focal-applet


mtwebster:
Fwiw, you could also just use cinnamon's existing calendar dbus daemon.

https://github.com/linuxmint/cinnamon/tree/master/calendar-server
https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/calendar%40cinnamon.org/eventView.js#L324-L377
https://github.com/linuxmint/cinnamon/blob/master/src/org.cinnamon.CalendarServer.xml

It allows you to request a time-range of events then receive a list in a callback.

zoharsnir:
> Fwiw, you could also just use cinnamon's existing calendar dbus daemon.
> 
> https://github.com/linuxmint/cinnamon/tree/master/calendar-server https://github.com/linuxmint/cinnamon/blob/master/files/usr/share/cinnamon/applets/calendar%40cinnamon.org/eventView.js#L324-L377 https://github.com/linuxmint/cinnamon/blob/master/src/org.cinnamon.CalendarServer.xml
> 
> It allows you to request a time-range of events then receive a list in a callback.

Thanks @mtwebster , will look into it later

zoharsnir:
@mtwebster ,

I did some research about your suggested approach. Please see my summary below (concerns sorted most-to-least important to me).

In my view, while your approach is more elegant, the current approach is still much preferred.
The most important concerns I have with moving calendar fetching to js/push are process isolation (the last thing I want is risking a bug or transient issue impact cascade from my applet to the cinnamon shell), debuggability, and the added complexity. Also (but to a lesser degree, yet another layer of complexity) not happy to add to my applet the responsibilities of connection and time range handling, and state management (See Update mechanism concern).

Please let me know what you think.

# Summary table

Note: although it's not fundamental to the suggested change I bundled the calendar listing migration to JS as part of Suggested Approach as it makes sense if moving the event fetching/listening.

| Concern | Existing Approach (`calendar_helper.py`) | Suggested Approach (`CalendarServer` + JS listing) | Comments |
| --- | --- | --- | --- |
| Process isolation | Isolated async subprocess, 15s watchdog. Minimal and hard boundary from Cinnamon | In-process (same GJS process as the shell) | A calendar list/fetch hang/crash today is contained to a disposable subprocess - the worst case is a user facing "Calendar error" with zero impact on the shell. In-process has a structurally worse blast radius. Async practices can mitigate but complexity and the risk+cost of not getting it right is much higher. This applies both to this initial release and to any future release, or underlying breaking dependency change. |
| Standalone debuggability | Easy - Runnable directly from a terminal | No standalone path - indirect or custom debug code | `calendar_helper.py --current-or-next` can be easily run and inspected on its own with a very clear boundary and simple expected behavior. In-process signal logic debugging is only doable either indirectly (as part of applet.js general behavior) or with a separate custom/throwaway debugging js. |
| Complexity | Low - synchronous script (invoked async), one-shot JSON in/out | Higher - async throughout, signal accumulation, reconnection, rollover | Trading a simple, linear script for asynchronous, stateful, event-driven logic increases the complexity. More complexity means more effort (obviously) but it also inherently adds risk and potential bugs. |
| Connection lifecycle | N/A - stateless; fresh per poll | Must detect and handle `CalendarServer` restarts and other events. | With the new approach there's a need to thoroughly understand the different connection failure conditions and handle them. Some failure conditions (e.g. stale connection) might be silent to the user (no update to applet text). Testing all cases probably not going to be trivial. |
| Time Range Handling | N/A - "now" and next midnight recomputed fresh every poll | Requires explicit re-issue at midnight (and conditionally on sleep/wake) | `SetTimeRange` is bound to a fixed window. The client (applet.js) would have the responsibility of resetting the time range at midnight, sleep-wake, other cases? Same as with "Connection lifecycle" concern, not trivial to test all cases. |
| Timezone resolution | Manual workaround (`get_builtin_timezone` + offset math) | Native `TimezoneCache.get_timezone` + `as_timet_with_zone` | Existing approach needed a workaround for some cases (#4). The suggested approach matched the existing approach's time exactly. Workaround is stable and based on python's mature time zone libraries, but there's an advantage to not needing a workaround. |
| Load pattern | 1 Poll per minute (scoped for now..midnight, single selected calendar). Fresh connection per poll. | Long running connection watches all enabled calendar, continuously | `calendar_helper.py` connects to just the selected/default calendar each poll. `CalendarServer` opens an EDS view per enabled source regardless of selection - client-side calendar uid filtering to the selected calendar only. |
| Update mechanism | Poll (`poll-interval-seconds`, default 60s) | Push - signals on real data changes | Existing: fresh subprocess + query every interval. Minimum settable interval is 15s (for dev cycles). Suggested: `SetTimeRange` on load (and resume, midnight, etc.), then gets signals for changes within time range. Requires client to manage state (internal calendar data structure + update/query methods). |
| Per-calendar selection | Native - queries exactly the one selected calendar | Client-side uid-prefix filter on the merged stream | In the suggested approach - no server-side per-calendar filter exists. Event uids are prefixed with the source uid and filtering happens on the client. Also, not sure if a real concern but getting all calendars might produce unnecessary load for users with many calendars. |
| Python dependency | Required | None (assuming listing migrated too) | Not a big problem for a standard desktop install as I'm using only pre-installed packages. Performance hit is negligible - fetch takes ~150ms (real. ~0.035ms sys) on my pretty old laptop. Having said that - no python dependency is better. |

PR #8965, opened 2026-08-17
[weather@mockturtl] Center hourly toggle and stabilize refresh status

lucasleocs:
### Issue

This fixes a documented Known Issue in the Weather applet:

> Hourly forecast toggle button is not centered to the middle of the popup menu (See the [Weather applet page](https://cinnamon-spices.linuxmint.com/applets/view/17).)

The toggle uses `Align.MIDDLE`, but it is centered within its expanded `BoxLayout` allocation rather than relative to the whole bottom bar. Since the left and right sides have different natural widths, the toggle is shifted away from the actual center of the popup.

The refresh indicator also changes the natural width of the right side when it appears or disappears, which can shift the bottom-bar geometry during refreshes.

### Fix

- Make the three top-level bottom-bar regions homogeneous, keeping the hourly toggle geometrically centered.
- Group the provider credit and refresh status into the right-hand region.
- Keep the refresh-status width stable by overlaying the success, progress, and error states (`✓`, `⟳`, `✕`) and switching them with opacity instead of adding or removing actors.
- Preserve the refresh status across UI rebuilds.
- Move refresh-status ownership to `WeatherLoop`.
- Make refresh cleanup request-specific so a cancelled or superseded refresh cannot overwrite the status or cleanup of a newer request.

### Before / After

**Before**

Screenshot from 2026-08-17 17-50-11

**After**

Screenshot from 2026-08-17 17-55-30

### Testing

Tested on Cinnamon 6.6.9:

- The hourly toggle remains centered after the popup layout settles.
- Normal refresh transitions correctly from success → progress → success.
- UI rebuilds preserve the current refresh status.
- Two refreshes started 200 ms apart complete without a stale request incorrectly showing an error.
- Bottom-bar width remains stable across refresh states.
- Offline state shows the error indicator, and reconnecting starts a new progress state.
- `validate-spice weather@mockturtl` passes.
- `git diff --check` passes.
- The generated bundle passes `node --check`.
- `npx eslint ./src/3_8/` introduces no new lint errors; the remaining `uiHourlyForecasts.ts:74` nested-ternary error is already present upstream.

### Notes

Version bumped from 3.6.10 to 3.6.11.

The TypeScript source was built successfully with webpack during development.

A fresh webpack build also introduces several thousand lines of unrelated toolchain/transpilation changes in the tracked generated bundle, so the equivalent generated JavaScript changes were applied minimally to keep the PR diff reviewable.

### DIFF
--- a/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
+++ b/weather@mockturtl/files/weather@mockturtl/3.8/weather-applet.js
@@ -19067,7 +19067,6 @@ class WeatherLoop {
         this.appletRemoved = false;
         this.errorCount = 0;
         this.runningRefresh = null;
-        this.refreshingResolver = null;
         this.refreshing = null;
         this.NetworkMonitorUsed = null;
         this.OnNetworkConnectivityChanged = () => {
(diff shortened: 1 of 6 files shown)


lucasleocs:
Hi @mockturtl, just a small note since I opened a few Weather PRs close together: I really like this applet and use it regularly, so I've been looking into a few issues I noticed while using it and testing possible fixes locally. I ended up having several fixes ready around the same time, but decided to submit them separately so each PR stays small and easier to review.

I hope they help, and I'm happy to change anything if needed. Thanks for maintaining the applet!

PR #8975, opened 2026-08-20
[spicy-clipboard-applet] Add keyboard shortcut + fix history save crash

MSx11:
## Note:

> I am not a developer, and I have only been diving into Linux for less than a week now. I made these small changes to help speed up my workflow. If I have missed something or there are issues with my changes, I would really appreciate it if you could fix them on your end, since I am still learning. Hoping to see this update included in a future release. Thanks! 🙏

---

## Summary of changes

This PR adds two improvements to **spicy-clipboard-applet@Twilight0**:

### 1. Keyboard shortcut to open the clipboard menu

The applet previously had no way to open the clipboard history via keyboard. This adds a configurable keybinding setting:

- New `openShortcut` entry in `settings-schema.json` (type `keybinding`, unassigned by default)
- In `applet.js`: registers/unregisters the hotkey via `Main.keybindingManager` when the setting changes
- Hotkey is properly cleaned up in `on_applet_removed_from_panel`

### 2. Fix crash: `make_directory_with_parents_async` is not a function

The `_saveHistory` function used `Gio.File.make_directory_with_parents_async()`, which does not exist in the version of GJS shipped with Cinnamon. This caused a JS critical error every time an item was copied, preventing history from being saved.

Fix: replaced with the synchronous `make_directory_with_parents()`, which is appropriate here (directory creation is instant and idempotent). The actual file write remains async.

### Files changed
- `applet.js` — keyboard shortcut binding + directory creation fix
- `settings-schema.json` — new `openShortcut` keybinding setting

### DIFF
--- a/spicy-clipboard-applet@Twilight0/files/spicy-clipboard-applet@Twilight0/applet.js
+++ b/spicy-clipboard-applet@Twilight0/files/spicy-clipboard-applet@Twilight0/applet.js
@@ -293,7 +293,7 @@ MyApplet.prototype = {
             Main.keybindingManager.addHotKey(
                 this._keybindingId,
                 this.openShortcut,
-                Lang.bind(this, this._onOpenShortcut)
+                () => this._onOpenShortcut()
             );
         }
     },
--- a/spicy-clipboard-applet@Twilight0/files/spicy-clipboard-applet@Twilight0/applet.js
+++ b/spicy-clipboard-applet@Twilight0/files/spicy-clipboard-applet@Twilight0/applet.js
@@ -59,6 +59,7 @@ MyApplet.prototype = {
 
             // Bind settings
             this.settings = new Settings.AppletSettings(this, metadata.uuid, this.instance_id);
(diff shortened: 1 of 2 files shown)


Twilight0:
Hi @Mourad-S,

Thank you very much for taking the time to contribute and help improve the applet!

Regarding the crash fix: using synchronous `parent.make_directory_with_parents(null)` can block the Cinnamon UI thread during filesystem operations, which goes against Cinnamon Spices guidelines favoring non-blocking asynchronous I/O. In PR #9008, I have implemented a fully non-blocking, recursive asynchronous solution (`_ensureDirAsync` using GIO's `make_directory_async`) alongside a fix to prevent concurrent `xclip` process leaks.

Could you please close this PR? Once PR #9008 is merged, you are very welcome to submit a fresh PR focused solely on the keyboard shortcut feature (using an arrow function or `Function.prototype.bind()` to resolve the automated `Lang.bind` deprecation warning), and I'll be glad to support it!

Thanks again for your contribution and initiative!

PR #8976, opened 2026-08-20
window-closer@stefan-schmidbauer: adaptive scaling, scrolling and multi-monitor support (v1.1.1)

Stefan-Schmidbauer:
The overlay grid was laid out for a 1920x1080 screen, so cards looked tiny on 4K panels and ran off screen once many windows were open. This release makes the layout adapt to the actual display.

### Changes

- Card metrics derive from the monitor's logical size and the desktop text scaling factor, so the grid stays readable on HiDPI and 4K displays. (The first push read physical pixel density via `Gdk.Display`; that was replaced after the scanner finding — see the comment below.)
- Cards shrink only down to a recognisable minimum; beyond that the grid scrolls instead (mouse wheel, arrow keys, Page Up/Down, Home/End, Space) and the subtitle tells the user so
- The overlay opens on the monitor that holds the panel, while the backdrop dims every screen — previously the primary monitor was assumed
- Thumbnails are fitted to their card once the real allocation is known, so they keep their aspect ratio
- Screenshot resized from 2900px / 554 KB to 1600px / 130 KB
- Version bumped to 1.1.1

### Translations

Adds a German translation. No existing `msgstr` was modified anywhere — `es`, `fr` and `hu` only gained the two strings that are new to this release:

- the scroll hint, which arrived as a fuzzy match of the old subtitle and would otherwise have fallen back to English
- `metadata.json->description`, which the regenerated `.pot` exposes as translatable for the first time

I filled both in to keep the release complete — native speakers are very welcome to correct my wording.

### Testing

`validate-spice window-closer@stefan-schmidbauer` passes. Tested live on a dual-monitor setup (3840x2160 + 2560x1600, text scaling 1.3), with both few and many windows open.


Stefan-Schmidbauer:
The `gdk_display` finding above is from the first push and no longer applies — commit `39be8fb` removes `Gdk` entirely.

It was only there to read a monitor's physical size in millimetres for a density-based scale boost, which Muffin does not expose, so there was no direct replacement. The applet now reads `text-scaling-factor` from `org.cinnamon.desktop.interface` instead: people on dense panels turn that up, and St does not apply it to the inline pixel sizes used here, so the overlay was ignoring a preference it should have honoured. Behaviour is unchanged at a factor of 1.0.

Tested live on a dual-monitor setup (3840x2160 + 2560x1600, text scaling 1.3), with both few and many windows open.

The last commit adds a German translation. On the existing translations: no existing `msgstr` was modified. The only additions are the two strings that are new to this release — the scroll hint, and `metadata.json->description`, which the regenerated `.pot` exposes as translatable for the first time. I filled both in so the release is complete; native speakers are very welcome to correct my wording there.

Issue #5157, opened 2023-11-28
kdecapplet@joejoetv: Feature Request


rcalixte:
### Applet name and maintainer

kdecapplet@joejoetv @JoeJoeTV

### What would you like to see?

Similar to #5145, would it be possible to make changes to allow support for multiple instances of this great applet? On multi-display setups, having a maximum of just one instance is restrictive. If you want, I can take a shot at it or leave it up to you but want to make sure you were okay with the potential change first.

JoeJoeTV:
I had a quick look at the code and except for the metadata setting, I don't think there is much in the way of blocking this change, except that the configuration file will be renamed, because Cinnamon names the config files for single-instance applets after their UUID and for multi-instance applets it names them after their instance ID. This means people would have to export their current configuration and re-add it after the update or just rename the file after the update to reflect the new name.

klangman:
> This means people would have to export their current configuration and re-add it after the update or just rename the file after the update to reflect the new name.

@rcalixte  I had been thinking bit about this issue along with other issues with settings and multiple instances. My suggestion would be to have a special flag in settings-schema.json entries where the developer can specify whether a specific setting should be global or instance specific  (i.e.  "global": true).  

One issue I have in my applets is that some settings are clearly global in nature and should be the same for all instances of the applet, where other settings are instance specific. The settings feature I propose would allow the developer to decide what is most appropriate in each case. The end result would be applets that works better and are less confusing for the users.

It could be implemented with a general applet config file using the UUID name for global options and one instance specific config file for each instance. The cinnamon-setting code could use both files when creating a new AppletSettings object and read from/write to the appropriate config file based on the developers global flag for the specific setting.

This idea might help this applet too, if it did not need any instance specific settings then changing to allow multiple instances would not require the user to manually copy config files after the applet changes to allow multple instances since it could modify all the settings to include ("global": true") and therefore the UUID config file would still be used after the applet is updated.

fredcw:
@klangman I could be wrong but think this might be fairly easy to do in the applet. In `appletManager.js` is `function getRunningInstancesForUuid(uuid)` Maybe this could be used when a global setting is changed to then change it in all other instances with something like:
```
this.settings.someGlobalSetting = someValue;

appletManager.getRunningInstancesForUuid(this.uuid).forEach( instance => {
    instance.settings.someGlobalSetting = someValue;
});
```

I haven't used or tested any of this btw.

rcalixte:
[...]

PR #8977, opened 2026-08-20
[kdecapplet@joejoetv] Version 2.2.0

JoeJoeTV:
New version of the KDE Connect applet with following changes:

- FIX: Changed how the applet tries to open the KDE Connect settings window (Should fix that part of #8693)
- FIX: Make non-critical python modules optional (Also mentioned in #8693)
- CHANGE: Refactor custom menu item implementation
- CHANGE: Replace logger implementation with a simpler and more usable one
- FIX: Use the directories provided by KDE Connect and add a selection menu when clicking the button to start browsing.
  - This fixed the issue where the folder opened by the button would be empty, as only the specific subpaths of the directories were readable
- CHANGE: Use newer Javascript features and thus split by Cinnamon version based on support
- CHANGE: Added multi-instance support and added migration functionality which copies the configuration file to the new instance-specific name

This should resolve issues #5157 and #8693, aswell as superseede the PRs #8712 and #6478.

@rcalixte Sorry it took so long for some of this.


rcalixte:
> @rcalixte Sorry it took so long for some of this.

No need to apologize.

rcalixte:
Feel free to shout when you're ready for this to be reviewed.

JoeJoeTV:
> Feel free to shout when you're ready for this to be reviewed.

I think it's ready. I implemented some changed based on the automated scan, but I believe for the rest of the items it detected, the usage is justified.

The code using synchronous query_exists is based on code from the extension system itself and needs to know the status of files existing to make a decision. it also only runs once at applet startup.

And for the shell_string_spawn, I am actually using argv, but this is done by using the options (opts) argument of the function, which overrides the command string, so it is empty.

rcalixte (approved):

PR #8983, opened 2026-08-23
lumendusk@kasun: add new applet

Kasun24:
New applet. One schedule that moves the desktop theme, night light and monitor brightness together.

Each of those usually needs a separate tool, and they drift apart. Here a transition is one event: appearance is driven by Mint's own style catalog, so the shell, panel, window borders, GTK/GTK4, Flatpak (XApp portal), icons and accent all change at once, alongside the night light and the brightness preset.

Day and night come from sunrise/sunset computed offline, or from fixed times. No network calls at any point.

## Menu

- Automatic / Manual. Manual hands the desktop back to the user; nothing moves until they switch back.
- A brightness slider (all monitors, or one at a time) across sysfs/brightnessctl, ddcutil (DDC/CI) and xrandr, normalised to 0-100%.
- Apply day/night now, settings, and a shortcut to the config file.
- Each phase picks its own appearance, so "dark all day" is a setting rather than turning the applet off.

## Validation

`./validate-spice lumendusk@kasun` passes with no errors. All 38 files are inside `lumendusk@kasun/`; nothing else in the repo is touched.

## Notes for review

**The engine is Python, bundled under `files/`.** A Spices install extracts a zip and runs no installer, so the applet carries the engine plus two pure-Python dependencies (astral for sunrise/sunset, tomli for TOML on Python < 3.11). Both are unmodified, both carry their licences in `engine/vendor/`, and nothing compiled ships — no `.pyc`, `.so`, or `.mo`. The applet talks to it through `Gio.Subprocess` with an argv array.

**Runtime data goes to XDG directories**, never the installation directory: config in `GLib.get_user_config_dir()`, log in the state dir, monitor cache in the cache dir.

**Self-reviewed against `.github/copilot-instructions.md`** before submitting, as the README asks. It found three things, now fixed: the build was shipping `__pycache__` bytecode (its own smoke test ran the engine after the cleanup step, writing it back); config was read with the synchronous `GLib.file_get_contents`, now `load_contents_async`; and three timer callbacks returned `false` instead of `GLib.SOURCE_REMOVE`. Timers and the settings file monitor are released in `on_applet_removed_from_panel()`.

**`cinnamon-version` is a floor of 5.4.** Developed and tested on Linux Mint 22.3 (Zena), Cinnamon 6.6.9.

Source, changelog and issue tracker: https://github.com/Kasun24/lumendusk (MIT). Submitting v0.3.0.


Kasun24:
Thanks — all five scanner findings are addressed in 7c6f4ba.

The three `GLib.file_test` calls are now `Gio` `query_info_async`. Two of them pick the engine out of four candidate paths and one checks whether `config.toml` exists yet; two of those paths are under the user's home directory, so the point about network filesystems is well taken — a stalled stat there hangs the shell, not just this applet.

That made the engine lookup callback-based, which reached its three callers. None of them lost anything: every one was about to spawn a subprocess, so none could have used a synchronous answer. The menu build split into an async entry point and the part that draws.

The two `spawnCommandLine` calls are now `Util.spawn` with the argv array. They were already building an array and shell-quoting each part, but argv is simply safer — and the arguments being literals today is a poor reason to keep a shell in the path.

One more that the scanner didn't reach, found in my own pre-submission pass: `config.toml` was read with the synchronous `GLib.file_get_contents`, now `load_contents_async`. There are no synchronous file or spawn calls left in the applet.

Verified on Linux Mint 22.3 / Cinnamon 6.6.9 after the change: the menu builds with the engine found, an Automatic to Manual to Automatic round trip works, and `config.toml` comes back byte-identical.

PR #8987, opened 2026-08-23
notifications-enhanced@hilyxx: cache notification actors offscreen (~160x faster menu)

rumours86:
**The bug.** Opening the applet menu tanks the whole desktop: St repaints every stolen notification actor (including its themed shadow) on every stage frame while the menu is open.

**Measurements** (Cinnamon 6.6.9, 3000x2000 panel, 8 notifications in the list, constant animation on stage as load):
- menu closed: **1.8 ms/frame**
- menu open: **363 ms/frame** (2.5 fps — videos visibly stutter, desktop lags)
- menu open, notification bin hidden: 2.15 ms/frame → the notification actors themselves are the cost
- menu open with this fix: **2.26 ms/frame** (max 5 ms)

**The fix.** Set `offscreen-redirect ALWAYS` on each notification actor when it is added to the list: the painted subtree is cached in a texture and reused between frames. Clutter invalidates the cache automatically when content changes (e.g. the timestamp refresh on menu open), so nothing else needs adjusting.

### DIFF
--- a/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/applet.js
+++ b/notifications-enhanced@hilyxx/files/notifications-enhanced@hilyxx/applet.js
@@ -301,6 +301,16 @@ class CinnamonNotificationsApplet extends Applet.TextIconApplet {
         // subtree is reused between frames (~2ms/frame measured); the cache is
         // invalidated automatically when the content changes (timestamps etc).
         notification.actor.set_offscreen_redirect(imports.gi.Clutter.OffscreenRedirect.ALWAYS);
+        // Give each notification its own theme node: St caches the
+        // CPU-prerendered (cairo) background per theme node PER SIZE, and all
+        // stolen notification actors share one node.  Notifications of
+        // slightly different heights therefore evict each other's cached
+        // background, and every menu open re-rasterizes every background via
+        // cairo (~40ms each, ~400ms per open measured with 8 notifications).
+        // A unique (unstyled) class per actor makes the node - and its paint
+        // cache - private: repeated menu opens drop to ~6ms.
+        this._cacheSlotSeq = (this._cacheSlotSeq || 0) + 1;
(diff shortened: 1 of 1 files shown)


rumours86:
Pushed a second commit that removes the remaining per-open cost.

Profiling showed two independent problems:

1. **While the menu is open** — every stage frame repainted all notification actors: 363 ms/frame (2.5 fps). Fixed by the offscreen redirect in the first commit → 2.3 ms/frame.
2. **On every menu open** — St re-rasterized every notification background through cairo: the prerendered background is cached on the theme node *per size*, and all stolen notification actors share one interned node, so notifications of slightly different heights evict each other's cache on every open (~40 ms each, ~400 ms per open with 8 notifications). Fixed by giving each notification actor a unique, unstyled style class → a private theme node whose paint cache survives: repeated opens measured at **~6 ms** (was ~400 ms).

The first open after new notifications arrive still pays the one-time cairo render; every open after that is effectively free.

rumours86:
Third commit: even with private theme nodes, the *first* open after the list changes still paid the one-time cairo rasterization (new notification's background + the resized menu background) — up to ~1s with a fresh notification.

Fixed by warming the caches off the click path: 1.5s after a list change the menu actor is shown for a single frame at opacity 1/255 (invisible; opacity 0 would be culled) with `reactive` disabled, then hidden. St fills its prerender caches during that frame.

End-to-end: new notification arrives → warmup runs in background → first click opens in **7 ms** (was ~1000 ms). With all three commits the applet menu is now effectively free to open and keep open.

Hilyxx:
Looks good to me. I'll just add one more specific cleanup in on_applet_removed_from_panel() to properly destroy the timer.
I was already planning to fix a few things in the code, so this will be included in my next commit. (without the huge comments :p)
Thanks!

Hilyxx:
This has been merged in PR #8989 . Therefore, this can be closed.

Issue #8998, opened 2026-08-30
Applet Bug Sensors@claudiux 


aklakan17:
### Applet version/Build date

5.11.5

### Cinnamon version

Cinnamon 6.6.9

### Distribution

Mint Cinnamon 22.3

### Graphics hardware *and* driver used

RTX 3060, 595.84

### Applet name and maintainer

Sensors@claudiux @claudiux

### What happened?

After updating from 5.11.4 to 5.11.5, the GPU temperature stopped being displayed.

claudiux:
Ping @joppino

joppino:
@aklakan17 please attach ~/.xession-errors

aklakan17:
@joppino 

[.xsession-errors.txt](https://github.com/user-attachments/files/31626830/default.xsession-errors.txt)

> [@aklakan17](https://github.com/aklakan17) please attach ~/.xession-errors

joppino:
@aklakan17 this is the issue:

Gjs-Message: 14:33:11.638: JS LOG: [LookingGlass/info] Nvidia SMI vnull command: /usr/bin/nvidia-smi --format=csv,noheader --query-gpu=name,pci.bus_id,temperature.gpu,index
Gjs-Message: 14:33:11.638: JS LOG: [LookingGlass/error] Could not find power_state handle or information for Nvidia card at /sys/bus/pci/devices/0000:2B:00.0/power_state, is sysfs mounted on /sys?

So I will provide a PR promptly to fix the problem.

Though, in order to provide a more comprehensive fix, @aklakan17 , can you confirm /sys is mounted? And can you provide the content of /sys/bus/pci/devices?

Thanks

aklakan17:
@joppino 

aklakan@aklakan-MS-7C91:~$ mountpoint /sys
/sys является точкой монтирования

aklakan@aklakan-MS-7C91:~$ ls -la /sys/bus/pci/devices
итого 0
drwxr-xr-x 2 root root 0 авг 31 15:59 .
drwxr-xr-x 5 root root 0 авг 31  2026 ..
lrwxrwxrwx 1 root root 0 авг 31  2026 0000:00:00.0 -> ../../../devices/pci0000:00/0000:00:00.0
lrwxrwxrwx 1 root root 0 авг 31  2026 0000:00:00.2 -> ../../../devices/pci0000:00/0000:00:00.2
[...]

PR #9009, opened 2026-08-31
Sensors@claudiux: Set sysfs path optional

joppino:
Fix GPU disappearing from sensors if for any reason the sysfs power_state interface is not accessible.


joppino:
Fixes linuxmint/cinnamon-spices-applets#8998

claudiux (approved):

rcalixte (approved):

PR #9023, opened 2026-09-07
sound150@claudiux: pause seeker polling while hidden

woodpile72:
This is a follow-up to #9019.

@claudiux

I have been working through a few Sound150 player-control problems in separate branches because I have had a lot of trouble with UI stalls here. I am trying to keep the fixes limited and easy to review instead of mixing all of it into one large change.

This patch keeps the seeker rule simple: if the seeker is not visible, do not poll MPRIS Position.

The previous change cleaned up player transport handling and stale player handling. This one stops the seeker from doing background Position work when the Sound150 popup is closed or when the seeker is hidden. The polling starts again when the popup is opened and stops when the popup closes.

Changes:
- remove the initial Position request from Seeker construction
- start seeker polling on the applet popup open-state-changed signal
- stop seeker polling when the popup closes
- stop seeker timers when the seeker is hidden or destroyed
- keep Position requests behind the same visibility/polling guard
- leave album art untouched for a separate patch

Tested locally with Sound150 reloaded in Cinnamon:
- validate-spice sound150@claudiux
- node --check on both 6.6 and 6.7 lib/s150PopupMenu.js
- git diff --check

### DIFF
--- a/sound150@claudiux/files/sound150@claudiux/6.6/lib/s150PopupMenu.js
+++ b/sound150@claudiux/files/sound150@claudiux/6.6/lib/s150PopupMenu.js
@@ -172,6 +172,7 @@ class Player extends PopupMenu.PopupMenuSection {
         this._owner = owner;
         this._busName = busname;
         this._applet = applet;
+        this._menuOpenStateChangedId = 0;
         players_without_seek_support = this._applet.players_without_seek_support;
 
         // We'll update this later with a proper name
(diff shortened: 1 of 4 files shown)


claudiux:
@woodpile72 
Thank you for all your work! Please update `CHANGELOG.md` and `metadata.json` files using `16.5.0` as version number.

claudiux (approved):

rcalixte (approved):

PR #9026, opened 2026-09-08
capsule-monitor@gaborkis11: add new applet

gaborkis11:
A compact system monitor that keeps every value in a fixed-width slot, so the panel never reflows as the numbers change. Memory, CPU, NVIDIA video memory, disk and network in one capsule; eight metrics in total, all optional and reorderable from the settings.

The network cell is a two-way graph with a centre line and the current speeds printed at the end. Each direction is scaled against its own recent peak so an upload stays visible during a large download, and only physical interfaces are counted so container bridges and VPN tunnels do not make traffic count more than once.

Memory, swap, video memory and disk can show either the percentage used or the amount used; the bar always shows the percentage, and alert thresholds are evaluated on the percentage in both modes.

Requires gir1.2-gtop-2.0. nvidia-smi is optional; without it the two GPU elements show a dash and never alert.


gaborkis11:
Thanks for the automated scan. I went through all nine findings; here is what
I found, in case it saves a reviewer some time.

**panel_height_property (2 findings, applet.js:326 and :500)** — false
positives. Both matched comment text, not code. The applet never reads or
writes `this.panel.height`; those comments explain why the code deliberately
avoids assigning to `_panelHeight` (the base class defines it as a getter-only
accessor, and `this.panel` is still null while the constructor runs). Reading
the whole file for `this.panel.height` outside comments returns nothing.

**sync_file_get_contents, sync_file_enumerate_children, sync_file_test** —
these are reads of `/proc` and `/sys`: memory statistics, the CPU temperature
from hwmon, and the check that decides which network interfaces are physical.
They are kernel virtual files and the reads take microseconds; the interface
enumeration additionally runs only once every ten seconds, not on every
sample. The one genuinely expensive call in the applet, `nvidia-smi`, is
already asynchronous via Gio.Subprocess and runs on its own slower timer.

**shell_string_spawn (applet.js:638)** — this is `Util.spawnCommandLine` for
the "program to launch on click" setting, which is the documented API for
exactly that, and the value comes from the user's own configuration.

Happy to change any of these if a maintainer would prefer a different
approach.

PR #9027, opened 2026-09-08
sound150@claudiux: avoid synchronous cover image loading

woodpile72:
This is a follow-up to #9023.

@claudiux

This patch keeps the scope limited to cover image loading in the Sound150 player menu.

Sound150 was already using St.TextureCache.load_image_from_file_async() for cover art, but _showCover() also loaded the same file synchronously afterward with GdkPixbuf.Pixbuf.new_from_file_at_size() and Clutter.Texture. That extra synchronous decode/load runs in Cinnamon's UI process and is not needed.

Changes:
- remove the unused getImageAtScale helper
- remove the direct GdkPixbuf import from s150PopupMenu.js
- remove the synchronous GdkPixbuf/Clutter.Texture cover load path
- keep the existing async St.TextureCache cover load path
- centralize cover actor replacement so stale cover art is removed cleanly
- keep the generic media-optical cover behavior when no art is available
- bump Sound150 to 16.5.1

Tested locally:
- validate-spice sound150@claudiux
- node --check on both 6.6 and 6.7 lib/s150PopupMenu.js
- git diff --check
- installed and reloaded in Cinnamon
- checked the user journal for Sound150/Cinnamon JS errors after reload

### DIFF
--- a/sound150@claudiux/files/sound150@claudiux/6.6/lib/s150PopupMenu.js
+++ b/sound150@claudiux/files/sound150@claudiux/6.6/lib/s150PopupMenu.js
@@ -965,7 +965,7 @@ class Player extends PopupMenu.PopupMenuSection {
             this._cover_path = cover_path;
             this._applet._icon_path = cover_path; // Added
             this._applet.setAppletIcon(this._applet.player, cover_path); // Added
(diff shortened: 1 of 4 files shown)


claudiux (approved):

claudiux:
@woodpile72 Many thanks for the care!

rcalixte (approved):

PR #9034, opened 2026-09-10
sound150@claudiux: Keep sliders within popup menu

woodpile72:
Fixes a layout problem I noticed on a small/scaled tablet display where the Sound150 popup contents could draw past the right side of the menu.

The slider rows were using Cinnamon's generic popup slider minimum width, which was too wide for this menu layout. This sets the Sound150 volume and balance sliders to the same 6em minimum used by Cinnamon's stock sound applet. It also clips the menu content box and the nested Applications section to their allocation, and applies the saved menu width through one scale-aware helper so the popup width is consistent when the applet is added or opened.

Tested with node --check for the touched JS files and ./validate-spice sound150@claudiux.


woodpile72:
This is what I was seeing on a small/scaled tablet display before the fix. The popup contents could draw past the right side of the menu.
image

claudiux (approved):

claudiux:
@woodpile72 
Thank you Erick!

rcalixte (approved):

PR #9038, opened 2026-09-12
Live Tennis Scores: new applet showing live scores on the panel

bensynapse:
**Vendor disclosure first:** I operate the Live Tennis API this applet reads from. It is a third-party commercial service with a free plan and no card required. I am submitting this as the API's operator, not as an unaffiliated user, and I would rather you weigh it knowing that. If a vendor-authored applet is not something you want in the collection, say so and I will withdraw it without argument. AI tooling was used while writing it; the code and the checks below are mine.

This would be **the first sports or scores applet** in the collection. I checked all 284 and there is nothing in that category today.

## What it does

Shows a live tennis score on the panel, cycling through the matches in play. The drop-down lists the whole board. Each line carries the set scores, the game in progress, a dot in front of the player serving, `TB` when the two numbers are a tiebreak count rather than 0/15/30/40, and `BP` on a break point, meaning the receiver at advantage or the receiver on 40 while the server is on 0, 15 or 30, and never during a tiebreak.

## One endpoint, one request per update

`GET /matches?status=live`, which is on the free plan. The URL is a fixed string with no user input interpolated. The key goes in a request header rather than the query string so it cannot land in a logged URL, and it is never written to the Cinnamon log. Tour and draw are filtered locally rather than as query parameters, so changing them costs no requests.

## Quota discipline, since this is someone else's allowance

A free key allows 100 requests a day, so 15 minutes, or 96 a day, is the floor. The settings spinner will not go below it and the code clamps it as well, so a hand-edited settings file cannot make it poll faster. A rate-limit response backs off a full hour rather than retrying into a budget that is already spent. `max-instances` is 1 for the same reason, so two copies cannot double the rate against one key. The panel rotation is local and free.

## With no key

The applet shows a short "set API key" label and makes no network request at all. Every failure is one short label and one log line naming only the failure kind. No stack traces, no URLs, no key.

## Housekeeping

Both libsoup 2 and 3 are handled, asynchronously only. Timer ids are tracked and zeroed when the applet is removed from the panel, with the documented source constants as return values rather than booleans. Nothing is written to the install directory. Strings use the standard gettext boilerplate, plurals go through the plural call, and format tokens are used rather than concatenation. The code stays within ES2020 for the Firefox 102 floor your compatibility table names.

`python3 validate-spice livetennis@livetennisapi` and `python3 validate-spice --all` both pass. I read the validator itself rather than guessing, so the square icon, the single `.pot` under `files/`, the absence of `.mo` files and the ASCII-only metadata values are all deliberate.

## Two things I could not do myself, flagged rather than papered over

**The screenshot is an illustration, not a capture, and it says so on its face.** I had no Cinnamon desktop on the machine I built this on, so there was nothing to photograph. I deliberately left that line in the image rather than presenting a mock-up as a real screenshot on a site users browse. The icon is hand-drawn too. Replace either, or ask me to, and I will.

**The applet has never run on a panel.** It parses in gjs, and its pure logic is covered by a local harness of 117 checks: break-point derivation across the point pairs and tiebreaks, player-major set-score formatting, the null-points and empty-games shapes a completed match returns, the local tour and draw filter including the rule that a null value matches neither option rather than acting as a wildcard, the response and error mapping, and the cadence clamp. But the panel label, the drop-down layout, the stylesheet, the timers and the libsoup 2 path are all unexercised. Point me at anything that looks wrong under a real Cinnamon and I will fix it.

I also made no authenticated API call, so field shapes come from our published OpenAPI document rather than an observed response. No translations ship yet, only the template.


bensynapse:
Thanks, the scanner is right. Fixed and force-pushed.

`Gettext.bindtextdomain` now uses `GLib.get_user_data_dir() + "/locale"` instead of hardcoding `.local/share`, so it respects `XDG_DATA_HOME`. Five applets in the tree already use that form, so it matches an existing convention rather than inventing one.

`python3 validate-spice livetennis@livetennisapi` and `--all` still pass.

PR #9041, opened 2026-09-15
ScreenShot@tech71: Fix screenshot commands and area selection

woodpile72:
Selected Area had been taking two attempts to start for me. This closes the applet menu without animation and starts gnome-screenshot from an idle callback, after the menu click has finished. Area selection now starts on the first attempt.

The menu actions also called Main.Util.spawnCommandLine, which throws "Main.Util is undefined" on Cinnamon 6.6.9. They now use Util.spawn with argument arrays through the Util module that the applet already imports.

Tested on Linux Mint 22.3 with Cinnamon 6.6.9 on X11. I tested Selected Area and the other screenshot menu operations; all worked. JavaScript syntax checking and validate-spice also pass. I also retested all menu operations after switching the six commands to argument arrays; all worked.


woodpile72:
Just a quick note: I'll be offline for a bit, so I wanted to check whether anything else is needed from me right now.

  I've been running this locally on Mint and it fixes the screenshot/area selection issue for me.

PR #9042, opened 2026-09-15
sound150@claudiux: Avoid adding Applications submenu twice

woodpile72:
The Applications submenu was added to _appVolumeSection during initialization and then added again in _showFixedElements(). Each applet load produced two Clutter warnings about actors that already had a parent.

Remove the second addMenuItem call and keep the existing initialization. The submenu is already in the section when the main menu is built.

Tested on Linux Mint 22.3 with Cinnamon 6.6.9 on X11. After reloading the applet, both parenting warnings were gone. I checked the Applications submenu and its volume sliders, and they work. JavaScript syntax checking and validate-spice pass.

@claudiux


claudiux (approved):

claudiux:
@woodpile72 Thank you for debugging this. Please update CHANGELOG and metadata.
In README, add you as Contributor (end of file).

claudiux (commented):
Same for 6.7, please.

rcalixte (approved):

PR #9044, opened 2026-09-16
Sensors v5.11.6: Compact tooltip layout (one line per sensor, no high/crit/min/max noise)

pavelhot-oss:
## Summary

Makes the applet tooltip readable: each sensor now occupies a **single, aligned line** showing only the name and the current value, instead of the current multi-line `name` + `value` + `high:` + `crit:` (or `min:`/`max:`) dump.

No new translatable strings are introduced (the `high:`, `crit:`, `min:`, `max:` labels are removed from the tooltip), so no `.pot` changes are required.

## Before

```
 CPU
  52°C  high: 95°C  crit: 100°C
 Composite
  45°C  high: 74°C  crit: 74°C
 fan1
  1234  min: 0
```

## After

```
 CPU      52°C
 NVMe     45°C
 fan1    1234
```

## Implementation

Each of the four tooltip sections (temperatures, disk temps, fans, voltages) in `updateTooltip()` was rewritten to append one formatted line:

```
" <b>" + name.padEnd(10, " ") + "</b> " + str_value.padStart(N, " ") + "\n"
```

- Names are left-aligned (10 columns), values right-aligned (5/5/8/7 for temp/disk/fan/voltage).
- `high:`/`crit:` for temperatures and disk temps, `min:` for fans, and `min:`/`max:` for voltages are removed — the readouts are already colored/altered by the applet's own threshold settings, and the tooltip no longer needs the cluttered threshold text.
- The section dividers (temp/fan/voltage header lines) are preserved.
- Note: the `bold_values`, `bold_italics_main_sensors`, and `show_temp_name` tooltip-related styling options no longer affect the tooltip layout (they still apply to the panel label).

## Compatibility

Only uses `String.prototype.padStart`/`padEnd` (available since Firefox 48), so it works on all Cinnamon versions the applet already supports (3.8+).

## Testing

Tested on Cinnamon 6.4 (Mint 22.x) with `k10temp`, `spd5118` (DDR5), `drivetemp`, `amdgpu`, and an NVIDIA GPU read through `nvidia-smi`. Syntax-checked with `cjs`.

Version bumped to `5.11.6` so users receive the update notification; happy to adjust or drop the bump if preferred.


claudiux:
@pavelhot-oss 
Thank you but this should be an option.

pavelhot-oss:
> @pavelhot-oss Thank you but this should be an option.

Yeah, makes sense. 
You mean like flip a switch for friendly names in the settings?

claudiux:
> You mean like flip a switch for friendly names in the settings?

Yes, please.

PR #9050, opened 2026-09-17
Antigravity Usage Monitor: add new applet

rawendil:
Adds `agy-usage@rawendil`, a panel applet showing Google Antigravity quota usage — the 5-hour and weekly windows, with the time left until the 5-hour window resets.

**How it gets the data.** It runs Antigravity's own command-line tool in the background (`agy -p "/usage" --output-format json`) through `Gio.Subprocess`, asynchronously, and parses the JSON. The applet makes no network calls of its own and talks to no API directly. It requires Antigravity to be installed and signed in; `agy` is looked up in the settings, then on `PATH`, then in `~/.local/bin`.

**Credit where it is due.** This is a derivative of `claude-usage@mtwebster` (GPL-3.0, @mtwebster). The panel layout, the settings wiring and the colour classes follow that applet, and `stylesheet.css` is essentially its stylesheet with a renamed prefix so the two can share a panel. What is new here is the data source (a subprocess instead of an HTTP endpoint), the response parsing, the cache that keeps the last known numbers on screen when a refresh fails, and the failure classification. The attribution is in the README's Credits section and in the file headers, and the applet is GPL-3.0-or-later.

**Notes for review.**
- `./validate-spice agy-usage@rawendil` passes.
- The computing logic lives in `lib.js` as pure functions with no Cinnamon imports, so it can be unit-tested outside the panel; that suite lives in the development repository.
- Translations follow the usual boilerplate. Because `lib.js` has to stay loadable outside Cinnamon, `applet.js` injects the translator into it instead of `lib.js` importing `imports.gettext` itself. `po/agy-usage@rawendil.pot` was generated with `cinnamon-spices-makepot`; a Polish catalogue is included.
- Timers and the in-flight subprocess are cancelled in `on_applet_removed_from_panel()`; persistent state goes through the settings API, nothing is written to the installation directory.


rawendil:
Thanks — the scanner's first finding was right and is fixed in the latest commit: `GLib.file_test()` is gone from the `agy` path resolution. Nothing needed the pre-check, because `Gio.Subprocess.new()` already throws when the binary is missing and the existing catch renders the same message.

On the second one (`hardcoded_data_dir`) I would like to keep `GLib.get_home_dir() + '/.local/share/locale'` as it is, and I think following the advice would actually break translations. Cinnamon installs xlet catalogues to that path with `.local/share` hardcoded, ignoring `XDG_DATA_HOME`:

- `cinnamon-settings/bin/Spices.py:27` — `locale_inst = f'{home}/.local/share/locale'`, used when installing a spice's `.mo` files (line 741)
- `cinnamon-xlet-makepot:11` — `LOCALE_DIR = os.path.join(os.path.expanduser('~'), '.local/share/locale')`, used by `cinnamon-spices-makepot --install`

So with `XDG_DATA_HOME` set to anything else, `GLib.get_user_data_dir() + '/locale'` would point at a directory the installer never writes to, and the applet would silently fall back to English. The stock applets shipped with Cinnamon bind the same way. Happy to change it if the installer side moves to the XDG variable.

PR #9055, opened 2026-09-19
yasm@sniemetz: sync features accumulated since initial submission

sniemetz:
Fast-forwards the applet with changes made upstream in the YASM repo since #8769. Prepares the tree for a follow-up bugfix PR (CPU temperature stuck at 0 on AMD hosts) that will be submitted after this lands.

## What's changed

**`applet.js`**
- CPU/GPU temperature alert notifications, each with a cooldown so a hot chip doesn't spam the notification tray on every poll.
- NVIDIA suspend fix — a GPU in PRIME/on-demand mode now shows as "suspended" instead of dropping off the panel with a read error.
- Icon-picker refactor: the per-tile icon picker was replaced by a per-tile display switch (icon vs. label). Simpler UI, fewer settings keys.
- Load-threshold reading: applet now reads `load-warn` / `load-alert` as percentages, matching what the schema already stores. (Previously it treated the stored 70/90 as fractions, so alerts never fired.)
- Small refactors and defensive-read cleanups.

**`settings-schema.json`**
- `cpu-temp-alert-notify` (checkbox) — whether to notify on CPU temp alert.
- `gpu-temp-warn` / `gpu-temp-alert` (spinbutton, °C) — GPU temperature thresholds, previously CPU-only.
- `gpu-temp-alert-notify` (checkbox) — GPU counterpart of the CPU flag.
- Minor renumbering in GPU section key ordering.

**`README.md`**
- Rewritten with a technical overview: data sources, polling cadence, why no daemon, GPU metrics via `nvtop`.
- Minimum refresh interval raised from 1 s to 3 s — sub-3 s polling made the applet's own subprocess spawns (nvtop, df) visible in the metrics it was trying to measure.

## Compatibility

No user-visible behaviour regressions expected. Existing stored settings remain valid under the updated schema.

## Testing

- Verified live on Cinnamon 6.6.9 (Mint 22.x).
- Applet loads, all tiles render, no errors in `~/.xsession-errors`.


sniemetz:
Hi @claudiux  - wasn't sure whether to submit single PR or split - opted to split. 
Let me know your preference. This PR doesn't fix much, and it's prereq to an actual bug fix. 

Cheers!

claudiux (approved):

rcalixte (approved):

PR #9058, opened 2026-09-20
yasm@sniemetz: v1.1.0 — AMD CPU temp, battery Full-state, vertical panels, compact mode

sniemetz:
Follow-up to #9055 (merged). Ships the v1.1.0 release of the YASM applet, bundling two user-visible bug fixes and a batch of layout / display features.

### 1. Fix: CPU temperature stuck at 0 on AMD hosts
The kernel driver reports CPU temp under `k10temp` / `zenpower` on AMD platforms; auto-discovery only matched Intel's `coretemp` and returned 0 elsewhere. New `findCpuTempChips()` matches all three known CPU-temp chip names, plus any chip carrying a `Package` / `Tctl` / `Tccd` label. A `cpu-hwmon-list` settings key mirrors the existing `fan-hwmon-list` for manual overrides on unusual hardware. Existing Intel systems unaffected.

### 2. Fix: battery tile treated 'Full' as fake discharge
On AC with the battery topped up, the kernel status is `Full` (or `Not charging`) and `current_now` reports an unsigned trickle from leakage or top-up. The applet's binary `isCharging`/not branch misread that as discharge (tooltip: `Draw: -0.01 W`; the RAPL usage estimate disappeared because it was gated on `isCharging`). Tri-state now: Charging / **Full (idle on AC)** / Discharging. AC icon whenever plugged in, RAPL estimate visible in both AC states, tooltip labels adjusted (`Idle: 0.00 W`).

### 3. v1.1.0 — vertical panel support, compact mode, tile polish
- **Vertical panels:** `setAllowedLayout(BOTH)` — Cinnamon reads `getAllowedLayout()`, not the metadata's `panel-orientation-support` field. Vertical panels now accept the applet.
- **Compact mode setting** (`Auto` / `Icons only` / `Text always`, default `Auto`). Auto = icons on vertical, text on horizontal. Compact mode hides tile text and forces icons on, drops the uptime tile's clock icon (load icon remains), and shrinks the icon multiplier from 60% → 45% of panel thickness.
- **Tooltip positioning** on vertical panels — anchors to the opposite side of the tile so it doesn't obscure the panel; bounds clamped to the primary monitor.
- **GPU tile** trimmed to `{busy%} | {temp°C}` (VRAM MB / clock MHz stay in the tooltip).
- **Disk tile** shows cumulative `{%free} | {I/O rate}` across all enabled mounts, matching the tooltip aggregate. Device row in the tooltip drops its redundant per-device I/O.
- **Fan tile** shows `off` when every fan reports 0 rpm (instead of `0rpm`, which read as a broken sensor).
- Version bumped to 1.1.0.

## Testing
Live-tested on Intel Cinnamon 6.6.9 (Mint 22.x): horizontal + vertical panels, all tiles render correctly, tooltip anchoring on left / right panels verified, disk aggregate matches tooltip, fan `off` triggers when curve is inactive.


claudiux (approved):

rcalixte (approved):
It would be good to look into adding translation support as well.

sniemetz:
> It would be good to look into adding translation support as well.

Appreciate that @rcalixte  - held off until someone else but me was using this ;) 
Will look into

